PR #51, opened 2020-09-23
Re-export image w/ paths

eohammond:


wesruv:
Not sure why tugboat didn't give a link to the environment it spun up for this PR, but it did make one: https://pr51-h7scz79zm9nd9hhwdqx8r2torkt1ohzn.tugboat.qa/

eohammond:
@wesruv what is Tugboat and does it change my workflow?

wesruv:
It will build a full version of the web site with just your changes applied. So we can check to make sure that everything looks right when it's not on our 'local'.

Just a good sanity check, doesn't have to change anything you do.

eohammond:
Got it. Thanks @wesruv

PR #196, opened 2021-11-15
Adding updates to limit width of text across site, and adding notes about it in documentation

dcaryll:
…bout it in documentation


coreyvickery:
@dcaryll Change `752px` to `750px`, otherwise good to go.

PR #231, opened 2022-04-14
Removed openshift.com from home page content

ryanaltvater:

PR #232, opened 2022-04-18
Removed openshift.com from home page content

ryanaltvater:
supercedes #231

### DIFF
--- a/pages/_templates/layout-base.njk
+++ b/pages/_templates/layout-base.njk
@@ -22,7 +22,7 @@
   <meta property="og:title" content="Red Hat digital design system">
   <meta property="og:site_name" content="Red Hat digital design system">
   <meta property="og:url" content="ux.redhat.com">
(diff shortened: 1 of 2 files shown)


bennypowers:
Thanks! looks like there's a conflict, specifically, that the changes in `pages/_templates/layout-base.njk` should be in `docs/_includes/layout-base.njk`, and likewise for `pages/index.njk` should be in `docs/index.njk`

ryanaltvater:
@bennypowers Ah, okay. `main` is the default branch, and so I created a branch off of that. So I didn't have a `docs` folder when I made the changes. I'll shift things over to use the `design-system` branch, and then make the changes in the correct spot.

Maybe switch `design-system` in the settings as the "default" branch for people to start from? 🤷‍♂️

bennypowers:
we're planning on merging that branch shortly, which is why I asked you to rebase off that one.

apologies for the confusion

ryanaltvater:
Okay cool, noted! 😄

PR #240, opened 2022-04-28
feat(alert): :tada: adding rh-alert and pfe-tools patch

brianferry:
Adding rh-alert element into the project along with an update to patch pfe-tools to allow css


bennypowers:
This is currently blocked by https://github.com/patternfly/patternfly-elements/pull/2016

which itself is sorta blocked by either me fixing that branch or by https://github.com/patternfly/patternfly-elements/pull/2017

bennypowers:
@brianferry I made some changes, PTAL and if you like it, merge it :D

bennypowers:
I think this is G2G now, any lingering bits we can adjust later

bennypowers (approved):

PR #254, opened 2022-05-04
feat(footer): add rh-footer

bennypowers:
### repo
- add eslintrc
- refactor tests in rh-alert
- add `/lib` with matchmedia controller and tokens (matchmedia controller to be upstreamed to PFE in a later PR)

### `<rh-footer>`
- copied over rh-footer from https://github.com/heyMP/rh-footer
- picked some code style nits therein
- made some perf improvements there wrt `lit/static-html.js`
- replaced CDN imports with bare specifiers (BREAKING!)
- moved styles to `.css` files, except responsive styles, which are in `.css.ts` so as to import tokens
- refactored matchmedia controller
  - `value` => `matches`
  - moved element-specific code to rh-footer
  - added `onChange` callback option

### `<rh-demo>`
- added rh-demo
- for the sake of consistency, we should (in a later PR) either adopt this across the board or remove it in favour of api-viewer


bennypowers:
@heyMP so there are some style changes I observed when using pfe-accordion 2:

your repo:
image

this pr:
image

Not sure how to proceed, what's the intended style?
In this PR, the accordion headers get their style from the lightdom stylesheet:
```css
rh-footer :is([slot^=links], [slot=footer-links-primary], [slot=footer-links-secondary]):is(h1, h2, h3, h4, h5) {
  font-weight: 500;
  font-size: 14px;
  margin-top: 0;
  margin-bottom: 0;
}
```
Whereas in your repo, they get them by virtue of being appended into the pfe-accordion v1 shadow root

Based on that, I _think_ that this PR represents the intended style, but I'm not certain

elements/rh-footer/rh-footer-lightdom.css
```diff
@@ -0,0 +1,35 @@
+rh-footer :is([slot^=links]) a {
+  gap: var(--rh-footer-links-gap, 10px);
+}
+
+rh-footer :is([slot^=links], [slot=footer-links-primary], [slot=footer-links-secondary]):is(h1, h2, h3, h4, h5) {
```
bennypowers, line 5: @heyMP couldn't we do this in shadow DOM through the magic of `:is` e.g.

```css
::slotted(:is([slot^=links], [slot=footer-links-primary], [slot=footer-links-secondary]):is(h1, h2, h3, h4, h5)) {
  font-weight: 500;
  font-size: 14px;
  margin-top: 0;
  margin-bottom: 0;
}
```

or is this in lightdom for noscript?

bennypowers: I went ahead and implemented this to show that it works, we'll revert if that's contrary to your original intention

heyMP: No, you're correct.  This should go in the shadow DOM. 👍

lib/MatchMediaController.ts
```diff
@@ -0,0 +1,49 @@
+import type { ReactiveControllerHost, ReactiveController } from 'lit';
+
+export class MatchMediaController implements ReactiveController {
+  /**
+   * The matchMedia result
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia
+   */
+  public matches = false;
```
heyMP, line 8: yes! Way more sense to match the API.  love it

elements/rh-demo/rh-demo.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import style from './rh-demo.css';
+
+@customElement('rh-demo')
+export class RhDemo extends LitElement {
+  static readonly styles = style;
+
+  constructor() {
+    super();
+    import('@patternfly/pfe-accordion');
```
heyMP, line 12: @bennypowers to your point about two different pfe-accordions.  This line is mostly leftover code from the original POC.  My thought was to get rid of it and have pfe-accordion a "requirement" of using this component.  That's my get out of jail free card option.

lib/MatchMediaController.ts
bennypowers: Should be initialized to null, I think

elements/rh-demo/rh-demo.ts
bennypowers: That makes sense but we'd probably still want to import accordion here.


I presented 3 options in the systems feds chat, will see what they say

lib/MatchMediaController.ts
bennypowers: I reworked this a bit, removing the resize observer from MatchMediaController, for these reasons:
1) it's not called [ResizeController](https://github.com/lit/lit/blob/main/packages/labs/observers/src/resize_controller.ts)
2) ResizeObserver throws errors even when it works, and [that's not going to change](https://stackoverflow.com/questions/39354331/how-can-i-prevent-mocha-from-trapping-unhandled-exceptions). This interferes with testing. There are some [p. bad workarounds](https://stackoverflow.com/questions/39354331/how-can-i-prevent-mocha-from-trapping-unhandled-exceptions), but I figure, hey - our RO is really only just calling `matchMedia`, why not just listen for changes on the media query list instead?

heyMP: Yeah, makes sense. 👍

heyMP:
@bennypowers I think this PR is looking pretty solid. I think it's probably ready for design review once we get preview urls set up. 🎉

bennypowers:
heads: rebased

bennypowers:
I'm going to try to solve that build error on main, will rebase when it's ready

bennypowers:
@brianferry  This depends on (in order)
1. https://github.com/patternfly/patternfly-elements/pull/2030
2. https://github.com/RedHat-UX/red-hat-design-system/pull/273

coreyvickery:
@heyMP @bennypowers 

Last-minute design tweaks, please.

---

Screen Shot 2022-05-12 at 10 59 20 AM

- [ ] `--pfe-icon--size: var(--rh-footer--social-icon--size, 32px);`: Change to `28px`

- [ ] `.social-links rh-footer-links, .social-links slot::slotted(rh-footer-links)`: Change `row-gap` to `24px`

---

Screen Shot 2022-05-12 at 11 11 09 AM

- [ ] Are these links not picking up dark theme?

---

Screen Shot 2022-05-12 at 11 11 51 AM

- [ ] Can we vertically align these links to the fedora?

---

Screen Shot 2022-05-12 at 11 11 38 AM

- [ ] Is this link not picking up dark theme either?

---

Screen Shot 2022-05-12 at 11 16 55 AM

- [ ] I think I remember in a previous footer, the Summit graphic was aligned to the top of `© 2022 Red Hat, Inc.` and `Privacy statement`. If we can do this again, great! If not, don't worry about it.

---

Please ping me with any further questions about these notes.

elements/rh-footer/RhFooter.ts
```diff
@@ -0,0 +1,236 @@
...
+import { pfelement } from '@patternfly/pfe-core/decorators.js';
+import { getRandomId } from '@patternfly/pfe-core/functions/random.js';
+import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
+
+import style from './rh-footer.css';
+import { responsiveStyles } from './rh-footer-responsive.css.js';
+
+import { tabletLandscapeBreakpoint } from '../../lib/tokens.js';
+// TODO: upstream to PFE
+import { MatchMediaController } from '../../lib/MatchMediaController.js';
```
zeroedin, line 13: Definitely would like to use this in `rh-secondary-nav` as well.  🔥

zeroedin: Issue ref: https://github.com/patternfly/patternfly-elements/issues/2032

zeroedin (commented):
Lemons Grapes Tangerines Mangos

elements/rh-footer/rh-footer.css
```diff
@@ -0,0 +1,327 @@
...
+  gap: 16px;
+}
+
+:host(:not([is-mobile])) .links {
+  display: grid;
+  grid-template-columns: repeat(1fr, 25%);
+  grid-template-rows: repeat(2, min-content auto);
+  grid-template-columns: 25%;
+  gap: var(--rh-footer-links-column-gap, 32px);
+}
```
zeroedin, line 203: `grid-template-columns` is defined twice, the second definition keeps the .links columns from taking up more then 25% when there might only be 2 sets of links instead of 4.  @coreyvickery brought this up in chat, this is my suggestion for a fix.

```Suggestion
:host(:not([is-mobile])) .links {
  display: grid;
  grid-template-columns: repeat(1fr, 25%);
  grid-template-rows: repeat(2, min-content auto);
  gap: var(--rh-footer-links-column-gap, 32px);
}
```

coreyvickery: Example screenshot.

Got approval that version B is how we want it to look.

Footer content flow examples

bennypowers: Screen Shot 2022-05-15 at 11 13 44
Added 2- and 3-column layouts to the demo to confirm this fix

elements/rh-footer/RhFooter.ts
```diff
@@ -0,0 +1,290 @@
...
+import '@patternfly/pfe-icon';
+import '@patternfly/pfe-accordion';
+
+function isHeader(tagName: string) {
+  return !!tagName.match(/^H[1-6]$/i);
+}
+
+/**
+ * @element 'rh-footer'
+ * @slot    base
```
bennypowers, line 33: @heyMP going through the motions of documenting these got me to wondering how many of these we want to expose. Do we have a concrete use case for the `base` slot, which overrides all the shadow content? If not, we should consider removing it (and other such slots), considering that it's easy to add to API surface, but difficult to subtract from it. WDYT?

bennypowers:
NB: I rebased this branch today

bennypowers:
If https://github.com/patternfly/patternfly-elements/pull/2035 releases first, we should update this PR to use the controller from core

elements/rh-footer/RhFooter.ts
coreyvickery: @bennypowers My instinct is to get rid of extra slots, creates the potential for customizations that we may not be aware of if or when they happen.

heyMP: Let's move this into a discussion thread for since this will have an effect on how we build all of our components.

heyMP (approved):
LGTM!

elements/rh-footer/RhFooter.ts
bennypowers: agreed. we're within our rights to make these kinds of breaking changes during pre-release season. however, we should be up front with our users about it: "just so you know, these escape hatches are here in this version, but we've discussed removing them before 1.0, so if you rely on them, it's on you to let us know"

Issue #236, opened 2022-04-20
[feat] Secondary Nav


markcaron:
Create Secondary Nav component.

### Links

- [**Design Spec**](https://xd.adobe.com/view/2da5ba7d-263c-4821-b482-29269d47a4e5-78f6/)
- [Early version of component on "What is OpenShift?"](https://cloud.redhat.com/learn/what-is-openshift)
- [Internal documentation](https://docs.google.com/document/d/1YID0q9datTquTuAGZI1FT1jG90QhK66lvlBvR3Uxlv0/edit#heading=h.ex2nfayqn2gx)
- [Internal checklist](https://docs.google.com/document/d/1yzgeVPiyKr_Wf4nePYmOa-c8Xb4SHLnwm0ukfHr7g0Q/edit#heading=h.rsbtjo2b9idy)

_Requestor: JVPs team_

methomps:
@coreyvickery @zeroedin cleaning up Trello and saw the Trello card had this XD linked: https://xd.adobe.com/view/72d9148d-c898-4d7d-beaa-b9f746acdbbf-278a/ 19 pages as opposed to 18. Should we update the link here? aka is what we have here the latest?

PR #259, opened 2022-05-05
feat(secondary-nav): add `<rh-secondary-nav>`

zeroedin:
## What I did

1. Implement `<rh-secondary-nav>` and child components
2. Add `variant=dark` attribute

Fixes: #236

## Testing instructions 

1. `npm run test` will perform available unit tests for this component

## Notes for Reviewers

1. Running `npm run start` will display the secondary nav in an iframe like view.  In order to see the correct breakpoints, you will need to use the demo's fullscreen view.  **Caveat:** Full-screen mode breaks the `position: sticky` so menu will no longer stick to the top on scroll.  
2. Running `npm run proxy rh-secondary-nav` in a second terminal window will allow you to view the secondary nav in context with redhat.com, after you run the command it will tell you what port it is running on ie: http://localhost:8001 this will allow you to view with `position: sticky` and a generalized idea of implementation of how it could overlay primary nav + banners/universal-nav above.  Doing so will also provide a better showing of how secondary nav overlays over real content when a dropdown is clicked and scroll behavior on mobile and short screen sizes.

**TODO:**
- [x] add overlay on dropdown menu open, missed that in design spec
- [x] clean up dropdown state management between mobile and desktop views 
- [x] ensure when focus is outside element its state returns to closed.
- [x] mobile testing
- [x] write spec tests
- [x] write inline documentation / JSDoc
- [x] accessibility review
- [x] design review
- [x] code review


bennypowers:
hey @zeroedin you know `<dialog>` doesn't trap focus if it's not opened with `showModal()`?

try it on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

zeroedin:
@coreyvickery Committed fixes for your requested changes/fixes.  The alignment issue is still unresolved and open for further discussion.

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/**
+ * @summary 'A pop up menu for secondary nav, available in full-width and fixed-with sizes'
+ *
+ * @slot base             - Slot to override all shadow dom contents for full-width or fixed-width menus
+ * @slot section          - Hidden if empty, section slot applies an auto column grid system for `<ul>` link lists
+ * @slot links            - Only for type="fixed-width", expects a `<ul>`
+ * @slot cta              - Menu level CTA, expects a `<pfe-cta>`
+ *
+ * @csspart full-width    - {HTMLElement} container - The wrapper for full-width menus
```
bennypowers, line 23:
```suggestion
 * @csspart full-width    - container - The wrapper for full-width menus
```
and others. `@csspart` doesn't take a type, the analyzer will just spit out a description string with `{HTMLElement}` in it, which isn't great.

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+ * @csspart fixed-width   - {HTMLElement} container - The wrapper for fixed-width menus
+ * @csspart sections      - {HTMLElement} container - The wrapper for menu sections
+ * @csspart cta           - {HTMLElement} container - The wrapper for cta
+ */
+@customElement('rh-secondary-nav-menu') @pfelement()
+export class RhSecondaryNavMenu extends LitElement {
+  static readonly styles = [styles];
+
+  // TODO: change to variant instead of type which is used by html element properties such as input.
+  @property({ reflect: true }) type: 'fixed-width' | 'full-width' = 'full-width';
```
bennypowers, line 33: in this case what about `layout` or `size`?

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  @queryAssignedNodes('cta', true)
+  private _ctaNodes: NodeListOf<HTMLElement> | undefined;
```
bennypowers, line 36:
```suggestion
  private _ctaNodes?: NodeListOf<HTMLElement>;
```
equivalent, a drop terser, and to my eye nicer. season to taste

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+    this.id ||= getRandomId('rh-secondary-nav-menu');
+  }
+
+  /**
+   * `isMobile` property is true when viewport `(min-width: ${tabletLandscapeBreakpoint})`.
+   * Property is observed for changes, and its value is updated using matchMediaControler
+   * when viewport changes at breakpoint or first load of the component.
+   */
+  @observed
+  @property({ reflect: false }) _isMobile = false;
```
bennypowers, line 59:
```suggestion
  @state() private _isMobile = false;
```

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+   * @returns {boolean}
+   */
+  static isDropdown(element: Element|null): element is RhSecondaryNavDropdown {
+    return element instanceof RhSecondaryNavDropdown;
+  }
+
+  /**
+   * Updates isMobile property with matchMedia when viewport changes
+   */
+  protected matchMedia = new MatchMediaController(this, `(min-width: ${tabletLandscapeBreakpoint})`, {
```
bennypowers, line 73: can we use RHDSScreenSizeController here instead, so that we don't have to import tokens here? you also wouldn't have to set a state property either, just 
```ts
const { mobile } = this.#screenSize;
return html`<div class=${classMap({ mobile })}></div>`;
```

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+  });
+
+  firstUpdated() {
+    this._isMobile = (window.outerWidth < parseInt(tabletLandscapeBreakpoint.toString().split('px')[0]));
+  }
+
+  render() {
+    const classes = { 'is-mobile': this._isMobile };
+
+    let menuContents;
```
bennypowers, line 86: can we inline this?
```js
return html`
  <div id="container" class="${classMap(classes)}">${this.type === 'full-width' ? html`
    <div id="full-width" part="full-width">
      <slot name="base">
        <div id="sections" part="sections" hidden>
          <slot name="section" @slotchange=${this.#onSectionsSlotChange}></slot>
        </div>
        <div id="cta" part="cta" hidden>
          <slot name="cta" @slotchange=${this.#onCtaSlotChange}></slot>
        </div>
      </slot>
    </div>` : html`
    <div id="fixed-width" part="fixed-width">
      <slot name="base">
        <slot name="links"></slot>
      </slot>
    </div>`}
  </div>`;
  ```

File: elements/rh-secondary-nav/rh-secondary-nav-menu.ts [hunk 7]
```diff
@@ -0,0 +1,163 @@
...
+    return html`<div id="container" class="${classMap(classes)}">${menuContents}</div>`;
+  }
+
+  /**
+   * Toggles visibility of menu by added or removing a `visible` attribute to the host
+   * @param visible {boolean}
+   */
+  toggleVisibility(visible: boolean) {
+    if (visible) {
+      this._container?.classList.add('visible');
```
bennypowers, line 101: see previous, wrt imperative updates of shadow dom

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+    } else {
+      this._container?.classList.remove('visible');
+    }
+  }
+
+  #fullWidthMenu() {
+    return html`
+      <div id="full-width" part="full-width">
+        <slot name="base">
```
bennypowers, line 110: see prev re `base`

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * When CTA slot is changed, check to see if it has children
+   * if it does then remove hidden attributes
+   * @returns {void}
+   */
+  #onCtaSlotChange(): void {
+    if (!this._ctaNodes) {
+      return;
+    }
+    const hasCta = this._ctaNodes.length > 0;
```
bennypowers, line 141:
```suggestion
    this._cta?.toggleAttribute('hidden', this._ctaNodes.length > 0);
```
https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute#parameters

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
```diff
@@ -0,0 +1,163 @@
...
+   * When section slot is changed, check to see if it has children
+   * if it does then remove hidden attributes
+   * @returns {void}
+   */
+  #onSectionsSlotChange(): void {
+    if (!this._sectionsNodes) {
+      return;
+    }
+    const hasSections = this._sectionsNodes.length > 0;
+    hasSections ? this._sections?.removeAttribute('hidden') : this._sections?.setAttribute('hidden', '');
```
bennypowers, line 155: see prev

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
bennypowers: you could also avoid the `if(p) return` by nullish coalescing to 0 with optional chaining on the `toggleAttribute` call

elements/rh-secondary-nav/rh-secondary-nav-overlay.ts
```diff
@@ -0,0 +1,62 @@
...
+ * @summary A overlay to cover content when navigation is expanded.
+ */
+@customElement('rh-secondary-nav-overlay') @pfelement()
+export class RhSecondaryNavOverlay extends LitElement {
+  static readonly styles = [styles];
+
+  @state()
+  public open = false;
+
+  connectedCallback() {
```
bennypowers, line 30: remove

elements/rh-secondary-nav/rh-secondary-nav-overlay.ts
```diff
@@ -0,0 +1,62 @@
...
+    super.connectedCallback();
+  }
+
+  render() {
```
bennypowers, line 34: remove

elements/rh-secondary-nav/rh-secondary-nav-overlay.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Sets or removes the hidden attribute to toggle the overlay
+   * @param toggle {HTMLElement} overlay to toggle
+   * @param state {boolean} open = true, closed = false
+   * @param parent {RhSecondaryNav} parent nav of overlay
+   */
+  @bound
+  public toggleNavOverlay(toggle: HTMLElement, state: boolean, parent: RhSecondaryNav) {
```
bennypowers, line 45: rather than making this an imperative method call, why not just set `.open = true`, then handle the side effects here (or even in css)
```ts
@property({ type: Boolean, reflect: true }) open = false;
```
```css
:host(:not([open])) {
  display: none;
}
```

elements/rh-secondary-nav/rh-secondary-nav.css
```diff
@@ -0,0 +1,147 @@
...
+  z-index: var(--rh-secondary-nav-z-index);
+}
+
+#container.expanded ::slotted([slot="nav"]) {
+  display: flex !important;
+  list-style: none;
+  flex-direction: column;
+  padding: 2em 1em 1em;
+  margin: 0 !important;
+  max-height: calc(100vh - var(--rh-secondary-nav-min-height ,4.625em));
```
bennypowers, line 38:
```suggestion
  max-height: calc(100vh - var(--rh-secondary-nav-min-height, 4.625em));
```

elements/rh-secondary-nav/rh-secondary-nav.css
```diff
@@ -0,0 +1,147 @@
...
+  height: fit-content;
+  background-color: var(--rh-secondary-nav-background-color);
+  z-index: var(--rh-secondary-nav-z-index);
+}
+
+#container.expanded ::slotted([slot="nav"]) {
+  display: flex !important;
+  list-style: none;
+  flex-direction: column;
+  padding: 2em 1em 1em;
```
bennypowers, line 36: caveat, tokens

elements/rh-secondary-nav/rh-secondary-nav.css
```diff
@@ -0,0 +1,147 @@
...
+  background-color: var(--rh-secondary-nav-expanded-background-color);
+}
+
+button {
+  grid-area: menu;
+  border: none;
+  display: flex;
+  height: 100%;
+  align-items: center;
+  font-family: var(--pfe-theme--font-family);
```
bennypowers, line 61: warning, tokens

elements/rh-secondary-nav/rh-secondary-nav.css
```diff
@@ -0,0 +1,147 @@
...
+  background-color: var(--rh-secondary-nav-background-color);
+}
+
+button:hover {
+  color: var(--rh-secondary-nav-button-hover-color);
+  border-top-color: var(--rh-secondary-nav-brand-color);
+}
+
+
+button:after {
```
bennypowers, line 76:
```suggestion
button::after {
```

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -0,0 +1,389 @@
...
+  @query('button') _mobileMenuButton: HTMLButtonElement | undefined;
+
+
+  /**
+   * `isMobile` property is true when viewport `(min-width: ${tabletLandscapeBreakpoint})`.
+   * Property is observed for changes, and its value is updated using matchMediaControler
+   * when viewport changes at breakpoint or first load of the component.
+   */
+  @observed
+  @property({ reflect: false }) _isMobile = false;
```
bennypowers, line 74: see previous re RHDSScreenSizeController

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -0,0 +1,389 @@
...
+   */
+  protected matchMedia = new MatchMediaController(this, `(min-width: ${tabletLandscapeBreakpoint})`, {
+    onChange: ({ matches }) => {
+      this._isMobile = !matches;
+    }
+  });
+
+  connectedCallback() {
+    super.connectedCallback();
+    this.addEventListener('change', this._dropdownChangeHandler as EventListener);
```
bennypowers, line 96: in this case i recommend sticking with lib's typing and do an `instanceof` check inside the handler. if a user slots a `select` or `input` element and it changes...

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -0,0 +1,389 @@
...
+    this.addEventListener('overlay-change', this._toggleNavOverlay as EventListener);
+    this.addEventListener('focusout', this._focusOutHandler);
+    this.removeAttribute('role');
+  }
+
+  firstUpdated() {
+    // after update the overlay should be available to attach an event listener to
+    this._overlay?.addEventListener('click', this._overlayClickHandler as EventListener);
+    this._isMobile = (window.outerWidth < parseInt(tabletLandscapeBreakpoint.toString().split('px')[0]));
```
bennypowers, line 105: if this is needed, i'd rather see it calculated in `updated`, or in a (new) callback in RHDSScreenSizeController

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -0,0 +1,389 @@
...
+   * When dropdown event is triggered gets dropdown index that triggered
+   * event then closes all dropdowns.
+   * If the event is to open a dropdown, run #expand(index)
+   * If isMobile is set dispatch an SecondaryNavOverlayEvent event
+   * to open the overlay
+   * @param event - {SecondaryNavDropdownChangeEvent}
+   * @return {void}
+   */
+  @bound
+  private _dropdownChangeHandler(event: SecondaryNavDropdownChangeEvent): void {
```
bennypowers, line 167: 🤔 thinking this over

if you want to keep the parent in charge of setting state, would it make sense to call the event `expand-request`?

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -0,0 +1,389 @@
...
+  /**
+   * When isMobile value is changed
+   * Get all open navMenus
+   * If isMobile is true, open mobile menu
+   * If isMobile is false, close mobile menu and close overlay
+   * @param oldVal {boolean | undefined}
+   * @param newVal {boolean | undefined}
+   * @returns {void}
+   */
+  private __isMobileChanged(oldVal?: boolean | undefined, newVal?: boolean | undefined): void {
```
bennypowers, line 223: perhaps best handled with a (new) effects callback on screensize controller

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -0,0 +1,389 @@
...
+
+  /**
+   * Sets attributes on _mobileMenuButton and _container to open mobile menu
+   * @returns {void}
+   */
+  #openMobileMenu(): void {
+    if (this._mobileMenuButton?.getAttribute('aria-expanded') === 'true') {
+      return;
+    }
+    this._mobileMenuButton?.setAttribute('aria-expanded', 'true');
```
bennypowers, line 367: prefer to template this, using private state if necessary

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
bennypowers: but in any case, we should prefer to template this based on component state, rather than doing imperative updates

elements/rh-secondary-nav/test/rh-secondary-nav.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+        await aTimeout(100);
+        expect(isDesktop()).to.be.true;
+      });
+
+      it('should not have class is-mobile on shadow root nav', async () => {
+        await expect(element.shadowRoot?.querySelector('nav')?.classList.contains('is-mobile')).to.be.false;
+      });
+
+      it('lightdom passes the a11y audit', async function() {
+        await expect(element).to.be.accessible();
```
bennypowers, line 86: i don't believe you need to run this on both the element and it's `shadowDom`. Just `expect(element).to.be.accessible()` is sufficient

elements/rh-secondary-nav/test/rh-secondary-nav.spec.ts
```diff
@@ -0,0 +1,185 @@
...
+      .to.be.an.instanceOf(klass)
+      .and
+      .to.be.an.instanceOf(RhSecondaryNav);
+  });
+
+  it('should remove role="navigation" after upgrade', async function() {
+    expect(element.hasAttribute('role')).to.be.false;
+  });
+
+  it('should have an overlay set to hidden after upgrade', async () => {
```
bennypowers, line 41:
```suggestion
  it('should have an overlay set to hidden after upgrade', async function() {
```

prefer [function keyword in mocha tests](https://mochajs.org/#arrow-functions)

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -0,0 +1,414 @@
...
+  <p>We dont have container queries yet, so this compnent's media query break points are set in understanding that the component will exist at full screen width.  The breakpoints will appear off if the component isn't 100% page width. You can use the fullscreen button above to test media query breakpoint.  However doing so will break position: sticky of the menu.</p>
+</pfe-band>
+
+<rh-secondary-nav role="navigation">
+  <a href="#" slot="logo" id="logo-id"> <!-- id is upgraded if not given -->
+    Red Hat Ansible Automation Platform <!-- <svg> or text can be added here -->
+  </a>
+  <ul slot="nav" aria-labelledby="logo-id"> <!-- is this ul labelledby the text or above what if its a svg? -->
+    <li>
+      <rh-secondary-nav-dropdown>
```
bennypowers, line 17: I've got this itch in the back of my neck that says "pssst... abstract this into `/elements`..."

but also YAGNI, we'll worry about that when the time comes

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -0,0 +1,414 @@
...
+        <rh-secondary-nav-menu slot="menu">
+          <rh-secondary-nav-menu-section slot="section">
+            <h3 slot="header" id="ansible-platform"> <!-- id will upgrade if not given -->
+              <a href="#">Why Red Hat Ansible Automation Platforms</a>
+            </h3>
+            <ul slot="links" aria-labelledby="ansible-platform""> <!-- aria-labelledby will upgrade if not given -->
+              <li>
+                <a href="#">
+                  Integrations
+                  <pfe-icon icon="web-icon-new-window" color="info" size="sm"></pfe-icon>
```
bennypowers, line 28: beware eventual deprecation of pfe-icon or replacement with rh-icon

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -0,0 +1,414 @@
...
+          </pfe-card>
+
+        </rh-secondary-nav-menu>
+      </rh-secondary-nav-dropdown>
+      </li>
+    <li>
+      <rh-secondary-nav-dropdown>
+        <a href="#" slot="link">Grids</a>
+        <rh-secondary-nav-menu slot="menu">
+          <div slot="base"><h3>Grid</h3></div>
```
bennypowers, line 358: I dunnoo..... did people ask for this? I'd rather encourage them to use `::part()` and grid tokens, tbh

bennypowers (changes requested):
Code looks neat and organized, it's clear the care and thought that went into this and the results speak for themselves

My comments here are divided into three categories:
0. caveats about future (internal) api changes
1. nitpicks
2. proper change requests
you should be able to suss out which are which. ping here or DM with qs.

Nice work!

elements/rh-secondary-nav/demo/rh-secondary-nav.html
zeroedin: `<pfe-icon>` in that case is light dom and could be replaced by anything.

elements/rh-secondary-nav/demo/rh-secondary-nav.html
zeroedin: Can you explain this a bit more?  Are you suggesting getting rid of the `<rh-secondary-nav-dropdown>` element and abstracting it into the shadow dom so to speak?  If so let's have a conversation on that.  That should be something i should be able to tackle.

elements/rh-secondary-nav/test/rh-secondary-nav.spec.ts
zeroedin: Fixed in: de6c9fedc09782910db0fe477bb6879c55876c26

elements/rh-secondary-nav/test/rh-secondary-nav.spec.ts
zeroedin: Fixed in bd4ee5b939bc72ff81264f416e1a76b5ac421e2d

elements/rh-secondary-nav/demo/demo.css
```diff
@@ -0,0 +1,23 @@
+:root {
+  --pfe-theme--font-family: "Red Hat Text", "Overpass", Helvetica, Arial, sans-serif;
```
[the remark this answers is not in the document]
zeroedin: I'm actively looking forward to cleaning up some of the CSS when they do!

elements/rh-secondary-nav/rh-secondary-nav-dropdown.css
```diff
@@ -0,0 +1,56 @@
...
+  box-shadow: var(--rh-secondary-nav-box-shadow);
+}
+
+::slotted([slot="link"]) {
+  justify-content: space-between;
+  position: relative;
+  gap: 0.85em;
+}
+
+::slotted([slot="link"]):after {
```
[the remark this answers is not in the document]
zeroedin: old dog 🐶 bad habits 😆.  Good catch.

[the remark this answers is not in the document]
zeroedin: Fixed in 190153fba40663184149e3b5a41720e7b49f1dd6

bennypowers:
@zeroedin please also add yourself to the contributors list in package.json

elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts
```diff
@@ -0,0 +1,136 @@
...
+
+/**
+ * @summary A wrapper component to upgrade a top level nav link to include dropdown functionality
+ *
+ * @slot link   - Link for dropdown, expects `<a>`
+ * @slot menu   - Menu for dropdown, expects `<rh-secondary-nav-menu>`
+ *
+ * @fires { SecondaryNavDropdownChangeEvent } - Fires when a dropdown is clicked
+**/
+@customElement('rh-secondary-nav-dropdown') @pfelement()
```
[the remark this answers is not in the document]
zeroedin: Fixed in 0507600cb628ab536778cd1ab2bda775da1ff888

elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts
```diff
@@ -0,0 +1,136 @@
...
+
+import styles from './rh-secondary-nav-dropdown.css';
+
+/**
+ * @summary A wrapper component to upgrade a top level nav link to include dropdown functionality
+ *
+ * @slot link   - Link for dropdown, expects `<a>`
+ * @slot menu   - Menu for dropdown, expects `<rh-secondary-nav-menu>`
+ *
+ * @fires { SecondaryNavDropdownChangeEvent } - Fires when a dropdown is clicked
```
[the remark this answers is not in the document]
zeroedin: Fixed in dd20d1bff186b7270ea2ebd6995f3376ef3d8caf

elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts
```diff
@@ -0,0 +1,136 @@
...
+  @observed
+  @state() expanded = false;
+
+  connectedCallback(): void {
+    super.connectedCallback();
+
+    this.id ||= getRandomId('rh-secondary-nav-dropdown');
+
+    const [link] = this.#slots.getSlotted<HTMLElement>('link');
+    link.setAttribute('role', 'button');
```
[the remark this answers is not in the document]
zeroedin: You are not mistaken.  Would a guard clause with a logger return suffice here?

```js
const [link] = this.#slots.getSlotted<HTMLElement>('link');
const [menu] = this.#slots.getSlotted<HTMLElement>('menu');
if (link === undefined) {
  this.#logger.warn('[rh-secondary-nav-dropdown][slot="link"] expects a slotted <a> tag');
  return;
}
if (menu === undefined) {
  this.#logger.warn('[rh-secondary-nav-dropdown][slot="menu"] expects a slotted <ul> or <ol> tag');
  return;
}
 ```

File: elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts [hunk 34]
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * When a dropdown is clicked set expanded to the opposite of the expanded property
+   * and then dispatch that value in a SecondaryNavDropdownChangeEvent
+   * @param event {MouseEvent}
+   */
+  @bound
+  private _clickHandler(event: MouseEvent) {
+    event.preventDefault();
+    const expanded = !this.expanded;
+    // trigger change event which evokes the mutation on this.expanded
```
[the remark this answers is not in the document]
zeroedin: You aren't kidding 🙃   Redundant logic, was definitely written while tired or distracted.

Fixed in b0fc9e3ed05996d19da209b0479a06a6b233161d

elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts
```diff
@@ -0,0 +1,136 @@
...
+    menu?.toggleVisibility(false);
+  }
+
+  /**
+   * Toggles the dropdown as expanded by adding or removing the class `expanded` on the container
+   * @param expand {boolean}
+   */
+  #toggleExpand(expand: boolean) {
+    if (expand) {
+      this._container?.classList.add('expanded');
```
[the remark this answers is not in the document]
zeroedin: Fixed in b0fc9e3ed05996d19da209b0479a06a6b233161d

elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts
```diff
@@ -0,0 +1,136 @@
...
+    // trigger change event which evokes the mutation on this.expanded
+    this.dispatchEvent(new SecondaryNavDropdownChangeEvent(expanded, this));
+  }
+
+  /**
+   * Sets or removes attributes needed to open a dropdown menu
+   * @returns {void}
+   */
+  #open(): void {
+    if (this._container?.classList.contains('expanded')) {
```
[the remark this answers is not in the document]
zeroedin: Fixed in b0fc9e3ed05996d19da209b0479a06a6b233161d

elements/rh-secondary-nav/rh-secondary-nav-menu-section.ts
```diff
@@ -0,0 +1,86 @@
...
+ * @summary 'A menu secion which auto upgrades header and sibling link list accessibility attributes'
+ *
+ * @slot base       - Slot to override shadow dom contents
+ * @slot header     - Adds a header tag to section, expects `<h1>, <h2>, <h3>, <h4>, <h5>, <h6>`
+ * @slot links      - Adds a ul tag to section, expects `<ul>, <ol>`
+ * @slot cta        - Adds a section level CTA, expects `<pfe-cta>`
+ *
+ * @csspart base    - {HTMLSectionElement} container, <section> element
+**/
+@customElement('rh-secondary-nav-menu-section') @pfelement()
```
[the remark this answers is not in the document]
zeroedin: Fixed in 0507600cb628ab536778cd1ab2bda775da1ff888

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: That's muy caliente 🌶️ !  Yes definitely!

elements/rh-secondary-nav/rh-secondary-nav-overlay.ts
zeroedin: Fixed in 61bea434ccfe42654370827ed28825074a8c6f9d

elements/rh-secondary-nav/rh-secondary-nav.css
zeroedin: Fixed in 190153fba40663184149e3b5a41720e7b49f1dd6

elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
```diff
@@ -0,0 +1,262 @@
...
+  display: flex;
+  align-items: center;
+  height: 100%;
+  max-width: var(--rh-secondary-nav-logo-max-width, 10em);
+  text-decoration: none;
+  color: var(--rh-secondary-nav-logo-link-color);
+  font-size: var(--rh-secondary-nav-logo-font-size);
+  letter-spacing: 0.18px;
+  line-height: var(--rh-secondary-nav-logo-line-height, 1.25em);
+  margin-left: 1em;
```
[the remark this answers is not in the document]
zeroedin: I don't have much background in supporting multi-language environments, so this is a good practice to start using for sure!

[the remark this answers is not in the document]
zeroedin: Fixed in e95ac81f0a678bee7e2e3c088f056742f907078b

elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
```diff
@@ -0,0 +1,262 @@
...
+    flex-direction: row;
+    height: 100%;
+    align-items: center;
+    list-style: none;
+    padding: 0;
+    margin: 0;
+  }
+
+  rh-secondary-nav > [slot="nav"] > li {
+    border-top-width: 0px;
```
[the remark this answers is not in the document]
zeroedin: Fixec in e95ac81f0a678bee7e2e3c088f056742f907078b

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: `layout` is most likely the most appropriate IMHO.  

The `TODO:` comment was a holdover from prior to posting the discussion https://github.com/RedHat-UX/red-hat-design-system/discussions/293 on this very question.  I am open and would like to hear more opinions on this specific instance from the designers and will reference this comment there before making a change.

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in 2f88b975241399448e917d6be39fb408e4026d29

elements/rh-secondary-nav/rh-secondary-nav-menu-section.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * @summary 'A menu secion which auto upgrades header and sibling link list accessibility attributes'
+ *
+ * @slot base       - Slot to override shadow dom contents
+ * @slot header     - Adds a header tag to section, expects `<h1>, <h2>, <h3>, <h4>, <h5>, <h6>`
+ * @slot links      - Adds a ul tag to section, expects `<ul>, <ol>`
+ * @slot cta        - Adds a section level CTA, expects `<pfe-cta>`
+ *
+ * @csspart base    - {HTMLSectionElement} container, <section> element
```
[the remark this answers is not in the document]
zeroedin: Fixed in 2f88b975241399448e917d6be39fb408e4026d29

elements/rh-secondary-nav/rh-secondary-nav-dropdown.ts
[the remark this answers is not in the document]
zeroedin: Fixed in d8a379871a80d1538340ff2c0d890069a481e92a

elements/rh-secondary-nav/rh-secondary-nav-menu.css
```diff
@@ -0,0 +1,102 @@
+:host {
+  display: contents;
```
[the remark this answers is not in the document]
zeroedin: Fixed in d3516a1bf3183a7c3ad3590e3311b1e356701e6d

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in a90891b0d89c88bb1e42d7c922099a606c44cf7b

coreyvickery:
## 6/14 feedback

### Styling
* The Dark theme CTA arrow is not changing color on hover ([watch video](https://drive.google.com/file/d/1JwOXNdJ4JReMrdoKW4o3h158CRRFWwUA/view?usp=sharing))
* It might be my weird computer (I swear it was working correctly earlier), but some styles in both themes are not being picked up; worth looking into just to be safe ([watch video](https://drive.google.com/file/d/1adVTE7Mptb56cd0lnC-hKbJ2k6zfvXCi/view?usp=sharing))
Screen Shot 2022-06-14 at 2 31 32 AM
Screen Shot 2022-06-14 at 2 38 57 AM

### Accessibility
* After I close the menu, the red hover line is still visible. Is this because I'm still technically hovering over the menu I just closed? ([watch video](https://drive.google.com/file/d/19dfhnhUJZG6wuV_zRU_JE20A0b1ygek3/view?usp=sharing))
* Possible to click or tap on the component background to close the menu? If not, no worries ([watch video](https://drive.google.com/file/d/1JLbWJczPTy6cElYlTB69xL9163RT5RxV/view?usp=sharing))
* There is still no hover state for the product name text in both themes (XD has the correct styles)
* Whether using a cursor or keyboard (and in addition to the above request to be able to click or tap on the component background), can we have the `esc` key close the menu as well? (Pressing `enter` opens the menu and pressing `enter` again closes the menu if focus stays where it is, which is good)
* Still seeing the below when hovering on tablet and mobile sizes
Screen Shot 2022-06-14 at 2 10 20 AM

### Spacing
* Possible to reduce the amount of spacing between the title (`Use cases`) and bottom edge to 24px?
Screen Shot 2022-06-14 at 2 12 38 AM

### Other
* How are we feeling about so many scrollbars? I would like to hear about how others feel about scrolling within an accordion panel as well as a menu. This could also get screwy on mobile when content definitely scrolls a lot, let's keep discussing this
Screen Shot 2022-06-14 at 2 16 09 AM

Thanks so much, looking really good.

zeroedin:
> The Dark theme CTA arrow is not changing color on hover
Fixed in e0088a4d0828ed87d9ae632e93330c9a28d0a4aa

> It might be my weird computer (I swear it was working correctly earlier), but some styles in both themes are not being picked up; worth looking into just to be safe
We discovered this was an issue with not having the latest branch updates.  

> After I close the menu, the red hover line is still visible. Is this because I'm still technically hovering over the menu I just closed?
The menu has focus, this is a focus style being applied, this looks to be correct per the accessibility page in the design spec.  Let me know if that needs to be changed. 

> Possible to click or tap on the component background to close the menu? If not, no worries
Possible, yes, but I wonder if it is desirable as far as accessibility concerns go with reduced dexterity users and the mouse if creating too large of an area that would accidentally close the menu.  I wouldn't love to have @markcaron and @kelsS weigh in on this issue and find out if it adds a benefit or detriment to users.

> Whether using a cursor or keyboard (and in addition to the above request to be able to click or tap on the component background), can we have the esc key close the menu as well? (Pressing enter opens the menu and pressing enter again closes the menu if the focus stays where it is, which is good)
I don't think this is unreasonable but would like @markcaron and @kelsS to weigh in here.  We left out adding additional keyboard navigation as the menu itself should be navigated as a semantic UL of links.  Should we add in `esc` keypress event listener and manage focus by placing it on the "last open" menu item button after the menu is closed as an expected default?

> Still seeing the below when hovering on tablet and mobile sizes
Fixed in a005f23687bb80eb133afa65f346c6cae25484ce

> Possible to reduce the amount of spacing between the title (Use cases) and bottom edge to 24px?
Fixed in 25f2260596bb395c3c043418fd44de52ab949ec8

> How are we feeling about so many scrollbars? I would like to hear about how others feel about scrolling within an accordion panel as well as a menu. This could also get screwy on mobile when content definitely scrolls a lot, let's keep discussing this
It's definitely not my favorite.  Going to take a look closer at how current `rh-navigation` handles the overflow/scrolling.

elements/rh-secondary-nav/rh-secondary-nav-menu-section.ts
```diff
@@ -0,0 +1,86 @@
...
+  connectedCallback(): void {
+    super.connectedCallback();
+
+    this.#updateAccessibility();
+  }
+
+  render() {
+    return html`
+      <section part="base">
+        <slot name="base">
```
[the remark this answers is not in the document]
zeroedin: Fixed in dae47d21f1dbb493c0ea02097e84efb399fe5e64

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in dae47d21f1dbb493c0ea02097e84efb399fe5e64

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in dae47d21f1dbb493c0ea02097e84efb399fe5e64

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Cleaned up a bit in f5c0101068cec480c8513addb9eb25fb85abe708.  However, I would like to talk about the "preferred" method more.  I might be missing some points on how we can use the component state as it relates to checking if the slot has slotted children.  If there is another method I would love to learn it.

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in 1c4690143e173a313836f8b8944be2cb8fd11efc

elements/rh-secondary-nav/rh-secondary-nav-menu-section.ts
```diff
@@ -0,0 +1,86 @@
...
+import styles from './rh-secondary-nav-menu-section.css';
+
+
+/**
+ * Checks if the tagName string value matches
+ * h1, h2, h3, h4, h5, h6 and returns boolean
+ * @param tagName {String}
+ * @returns {boolean}
+ */
+// TODO: Rh-footer uses this function as well, should be abstracted into functions. (pfe-core/functions?)
```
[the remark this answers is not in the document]
zeroedin: I moved the function to `/lib` in 400704e4a3c413a1a0751e330368d1eabf7ee918
I implemented the type predicate in 4fef40481478723f2aaf90fa32b85b5946fe246a

Let me know if the type predicate solution is what you were looking.

elements/rh-secondary-nav/rh-secondary-nav-overlay.ts
zeroedin: Fixed in a4fa730f5058f56e4e829a9938b5264bf25a2b8d

elements/rh-secondary-nav/demo/rh-secondary-nav.html
zeroedin: Fixed in dae47d21f1dbb493c0ea02097e84efb399fe5e64

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in 4e6a18a554cbf20c1268d594fdadca59cf0d9b94

elements/rh-secondary-nav/rh-secondary-nav.ts
zeroedin: Assuming this is also just a naming issue, I have no problem switching that to expand-request.  It does give the event a more specific meaning.  

As far as wanting the parent in charge,  I assumed the parent would be the best place to handle the state of its children's dropdowns.  More specifically that any new dropdown open event closes all other open dropdowns so I just followed the parent -> child model thought process. 

You had mentioned at some point the idea of possibly abstracting the dropdowns (flyout menus)  into a completely separate component.  I haven’t thought about how that change would impact how I'm currently handling the open-close here.  

I can potentially see dropdowns listening for other dropdowns open request events, and if one is open all others close in response.  For reusability, If they aren’t used in a “grouping” like in the navigation menu itself, maybe you might not want that functionality. You might want to allow more than 1 to be open at a time, such as in an accordion, maybe?  Thought for later.  

That said, would something like that approach be preferable?  The dropdowns listen for `expand-request` events, and if it wasn’t the dropdown that called the event it would change its own state to closed?

elements/rh-secondary-nav/rh-secondary-nav.ts
bennypowers: My issue here isn't that the initial implementation of `RHDSScreenSizeController` is sacrosanct or anything - by all means improve it! My only issue here is doing this logic inside the component. If you need to make changes to `RHDSScreenSizeController` to get this feature in secondary-nav, go for it, I just don't want to see breakpoints imported or hard coded into components.

The reason why this is important is so that we maintain some level of uniformity and control over the design system. This is also the reason we created the tokens repo. We could very easily loosen this discipline, but that would lead to a situation where each individual component[ author] makes their own design assumptions, and keeps them private and local to line N in whichever component file. Even in the short term, this degrades maintainability.

So to answer your question in the google doc, yes, we should eventually put those components in the tokens repo and import them into `RHDSScreenSizeController.ts`. If, for the time being, you need to hard code values there or import them from `lib/tokens.ts`, no problemo, it's still less to worry about than having those token references scattered about in component lifecycle methods.

> But secondary-nav uses values not represented in tokens repo or tokens.ts

Add those values there. Components should not own design tokens.

> But I need to get this out for secondary nav now

Add the values in `RHDSScreenSizeController.ts`, in this PR. Make sure it doesn't break table. This shouldn't be a blocker, from what you've described, the problem is the internal implementation of `RHDSScreenSizeController`, which is very much open to change and improvement.

My _only_ requirements here are that breakpoints be handled exclusively in the controller, and that the component APIs be declarative (i.e., that components aren't doing any comparisons against token values). This is so that tokens are
1. available to any component while being owned by none
2. defined (if even initially aspirationally/locally) in the tokens
3. themeable

lib/functions.ts
```diff
@@ -1,9 +1,7 @@
 /**
- * Checks if the tagName string value matches
- * h1, h2, h3, h4, h5, h6 and returns boolean
- * @param tagName {String}
+ * Checks if the given `node` is an instanceof HTMLHeadingElemnt
  * @returns {boolean}
  */
-export function isHeader(tagName: string): boolean {
-  return !!tagName.match(/^H[1-6]$/i);
+export function isHeadingElement(node: Node | null): node is HTMLHeadingElement {
```
bennypowers, line 5: [TIL](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHeadingElement) thanks!

I didn't know this class exists. If you want, you can remove this and just inline the instanceof check, as far as TypeScript is concerned, it's the same thing. You can keep this predicate here if you prefer, too.

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
bennypowers: if you ask me, go with `layout` for now, and if we need to change it later per design's input, we'll do that in a subsequent PR

elements/rh-secondary-nav/rh-secondary-nav.ts
zeroedin: Fixed in 7996deb4adcfb81f26ebf40d05427e5bb55abb5c and c03d4820dbd8a6b95ee2520b3a7f5c4968b6e766

elements/rh-secondary-nav/rh-secondary-nav.ts
zeroedin: Fixed in c03d4820dbd8a6b95ee2520b3a7f5c4968b6e766

elements/rh-secondary-nav/rh-secondary-nav-menu.ts
zeroedin: Fixed in e441b00b80889be1fa6635a336fbf011e64c5334

elements/rh-secondary-nav/rh-secondary-nav.ts
zeroedin: Fixed in a14bf72a07b3519ddbe0c9e65ea0840f047dc53f

elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
```diff
@@ -0,0 +1,262 @@
+rh-secondary-nav {
+  --rh-secondary-nav-min-height: 4.625em;
```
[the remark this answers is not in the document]
zeroedin: Modified this approach by representing some of these CSS props as "private"  although they are still exposed in lightdom.css for critical CSS needs.  Changes in 66c0fb8444aa8747f4421325959617c674a37201

I've added token values for colors, but I have not approached `<length>` yet due to open discussions on `px` vs `r/em` on the tokens themselves. 

https://github.com/RedHat-UX/red-hat-design-tokens/discussions/12

kelsS:
@zeroedin below are the accessibility remediation items for the secondary nav.

Remediation items:

  The `esc` key needs to close the secondary nav menu completely
    -   focus should go back the "menu" button
    -   When secondary nav loses focus the menu should close automatically if it is open
         - The requirements from https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html(https://www.w3.org/WAI/ARIA/apg/example-index/disclosure/disclosure-navigation.html) for the navigation menu need to be implemented
  axe-devtool errors need to be fixed (to see run the axe devtools extension on the component demo.)
  
I am happy to answer any questions and/or walk through the remediation items with you in the Web Accessibility office hours. 🙂

elements/rh-secondary-nav/demo/proxy.html
```diff
@@ -13,11 +13,7 @@
               <h3 slot="header" id="ansible-platform">
                 <a href="#">Why Red Hat Ansible Automation Platforms</a>
               </h3>
-<<<<<<< HEAD
```
bennypowers: resolve please

bennypowers (approved):

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -436,6 +416,14 @@ export class RhSecondaryNav extends LitElement {
       this.style.setProperty('--_chevron-transform-expanded', 'rotate(var(--_chevron-down-angle)) translate(var(--_chevron-thickness), calc(-1 * var(--_chevron-thickness)))');
```
bennypowers, line 416: Think we could set this on a shadow container?

zeroedin: @bennypowers https://github.com/RedHat-UX/red-hat-design-system/pull/259/commits/b78be73d8eb8f5536bf8b98c1f2d8be054897f7c this change work for you?

bennypowers: yup

bennypowers:
Great work, @zeroedin and @kelsS. Very impressive!

PR #267, opened 2022-05-09
fix(alert)!: change HTML APIs

bennypowers:
Closes #265 

## What I did

1. `action-groups` slot -> `actions`
2. `description` slot -> anonymous
3. `title` slot -> header

These are in keeping with the conventions we've developed in PFE

4. use plain element selectors (e.g. `header`) when possible
5. use ID selectors when selecting single elements
6. remove the requirement to use specific classes in light DOM

This is in keeping with the proto-style-guide we developed in Google
Docs for PatternFly 2022

7. Prefer h2-6 tags for header, in keeping with semantic HTML

8. replace `dismiss` and `cancel` custom attrs with `data-action="dismiss|cancel"`

## Testing Instructions

1. `npm run start`, and compare with previous


brianferry (approved):
This all looks great to me!  Thank you for taking the time and looking at this component, I'll definitely keep these conventions in mind in the future.  Thanks @bennypowers

PR #279, opened 2022-05-16
feat(pagination): add `<rh-pagination>`

### LINKED ISSUE #234: [feat] Pagination Component
Create a Pagination component for both desktop and mobile.

### Latest Design

Image

### Design Doc

[Latest XD mockup](https://xd.adobe.com/view/75d471e8-9396-4721-8e3c-cf9aa0688d57-2087/)

### Other Resources
- [Old spec](https://xd.adobe.com/view/14878292-ce78-470c-b214-2eb41e11cada-41d4/)
- [Previous issue on PFE](https://github.com/patternfly/patternfly-elements/issues/200)

bennypowers:
## What I did

1. add pagination component. closes #234 

## Reviewer Notes
Be sure to use the browser's responsive dev tools to evaluate the demos
- [basic demo](https://deploy-preview-279--red-hat-design-system.netlify.app/components/pagination/demo/)
- [overflow demo](https://deploy-preview-279--red-hat-design-system.netlify.app/components/pagination/demo/many-pages/)
## TODO:
- [x] rebase to main after footer lands


bennypowers:
@ronniemcbride i'm going to try and improvise some dark colour pallette stuff. will ping you when it's ready for review.

elements/rh-pagination/rh-pagination.ts
```diff
@@ -0,0 +1,179 @@
...
+    const prevHref = this.#prevLink?.href;
+    const nextHref = this.#nextLink?.href;
+    const lastHref = this.#currentLink === this.#lastLink ? undefined : this.#lastLink?.href;
+    return html`
+      <div id="container" class=${classMap({ mobile, [size as string]: true })}>
+        <a id="first" class="stepper" href=${ifDefined(firstHref)} aria-label=${labelFirst}>${L2}</a>
+        <a id="prev" class="stepper" href=${ifDefined(prevHref)} aria-label=${labelPrevious}>${L1}</a>
+
+        <div id="nav-container" ?hidden=${mobile}>
+          <slot @slotchange=${this.#update}></slot>
```
heyMP, line 80: We've struggled with this is the past with PFE.  Should we use slotchange or mutation observer with a child tree.  We opted for mutation observers to make it easiest on the our users.

Example, this wont fire the "#update()" function

```
for (let item of [6,7,8,9]) {
    document.querySelector('rh-pagination ol li:last-of-type').insertAdjacentHTML('afterend', `<li><a>${item}</a></li>`)
}
```

elements/rh-pagination/rh-pagination.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  #update() {
+    this.#updateLightDOMRefs();
+    this.overflow = this.#getOverflow();
+    this.#validateA11y();
+  }
+
+  #getOverflow(): 'start' | 'end' | 'both' | null {
+    const overflowAt = 9;
```
heyMP, line 105: Should we make this an API option?

elements/rh-pagination/rh-pagination.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  #getCurrentLink(): HTMLAnchorElement | null {
+    const ariaCurrent = this.querySelector<HTMLAnchorElement>('li a[aria-current="page"]');
+    if (ariaCurrent) {
+      return ariaCurrent;
+    }
+    for (const link of this.#links ?? []) {
+      const url = new URL(link.href);
+      if (url.pathname === location.pathname && url.search === location.search && url.hash === location.hash) {
```
heyMP, line 129: These might not be valid urls if they are in an SPA.  Like if they do the "#"'s.

elements/rh-pagination/rh-pagination.ts
bennypowers: any less than 9 doesn't really work for us. we could maybe make larger overflows optional, but design didn't ask for it

elements/rh-pagination/rh-pagination.ts
bennypowers: agreed. fixed

elements/rh-pagination/rh-pagination.ts
bennypowers: I added some more facility for SPAs. check the readme and LMK what you think. The demo is basically an SPA. I also added some buttons to the demo which add/remove pages (like your MO case from above)

elements/rh-pagination/rh-pagination.ts
heyMP: Oh gotcha.  That makes sense.

elements/rh-pagination/rh-pagination.ts
heyMP: ahh, I'm sorry.  I missunderstood what this was doing.  I thought it was also validating the URL.  The docs explain perfectly. 👍

elements/rh-pagination/README.md
```diff
@@ -0,0 +1,63 @@
...
+    <li><a href="#5">5</a></li>
+  </ol>
+</rh-pagination>
+```
+
+For best results, navigation link text should be numbers only, as shown above.
+
+## Accessibility
+
+By default, `rh-pagination` comes with accessible English labels for it's internal components. When
```
heyMP, line 24:
```suggestion
By default, `rh-pagination` comes with accessible English labels for its internal components. When 
```

elements/rh-pagination/README.md
```diff
@@ -0,0 +1,63 @@
...
+```js
+import { router } from '#app/router';
+const paginator = document.querySelector('rh-pagination');
+router.addEventListener('navigation', function() {
+  paginator.requestUpdate();
+});
+```
+
+Calling `requestUpdate()` causes the paginator to check `window.location` (i.e. the url bar) and
+update it's own state in accordance. Therefore, for best results, ensure that your SPA uses URLs,
```
heyMP, line 60:
```suggestion
update its own state in accordance. Therefore, for best results, ensure that your SPA uses URLs,
```

elements/rh-pagination/demo/rh-pagination.html
```diff
@@ -0,0 +1,37 @@
+<link rel="stylesheet" href="demo.css">
+<link rel="stylesheet" href="../rh-pagination-lightdom.css">
+<script type="module" src="rh-pagination.js"></script>
+
+<header><h2 slot="header">Valid HTML</h2></header>
+
+<section>
+  <p>Paginators with 5 or fewer pages should not overflow, meaning all links should be visible.
+     Once a paginator has more than 5 pages, then must overflow, meaning some links will be hidden.
```
heyMP, line 9:
```suggestion
     Once a paginator has more than 5 pages, then it must overflow, meaning some links will be hidden.
```

elements/rh-pagination/rh-pagination.ts
```diff
@@ -0,0 +1,215 @@
...
+    const { mobile, size } = this.#screen;
+    const { label, labelFirst, labelPrevious, labelNext, labelLast } = this;
+    const firstHref = this.#currentLink === this.#firstLink ? undefined : this.#firstLink?.href;
+    const prevHref = this.#prevLink?.href;
+    const nextHref = this.#nextLink?.href;
+    const lastHref = this.#currentLink === this.#lastLink ? undefined : this.#lastLink?.href;
+    return html`
+      <div id="container" class=${classMap({ mobile, [size as string]: true })}>
+        <a id="first" class="stepper" href=${ifDefined(firstHref)} ?inert=${!firstHref} aria-label=${labelFirst}>${L2}</a>
+        <a id="prev" class="stepper" href=${ifDefined(prevHref)} ?inert=${!prevHref} aria-label=${labelPrevious}>${L1}</a>
```
heyMP, line 94: I think we should consider putting part attributes on all of the elements in the shadowroot, even if they are undocumented.  Keep them super flexible if/when the edge-case arises.

bennypowers: :grimacing: i don't want to get into a sticky situation here. the design guidelines were pretty clear, and a lot of the functionality is tied to the css. opening up parts could easily break stuff in this case.

heyMP: Gotcha. Sounds like we'll have a fundamental disagreement here. 🙃

I'll mark this as resolved and maybe we can come back to it.

brianferry:
On the pagination on mobile when I add or remove a new page it looks like it doesn't update the count until I either click on the arrow or resize the browser and bring it back down to mobile, so maybe a render update function needs to be called on change?

elements/rh-pagination/rh-pagination.ts
```diff
@@ -0,0 +1,215 @@
...
+
+        <a id="next" class="stepper" href=${ifDefined(nextHref)} ?inert=${!nextHref} aria-label=${labelNext}>${L1}</a>
+        <a id="last" class="stepper" href=${ifDefined(lastHref)} ?inert=${!lastHref} aria-label=${labelLast}>${L2}</a>
+
+        ${!mobile ? '' : html`
+        <div id="numeric">
+          <span id="go-to-page">
+            <slot name="go-to-page">Go to page</slot>
+          </span>
+          <input inputmode="numeric" aria-labelledby="go-to-page" value=${(this.#currentIndex ?? 0) + 1} />
```
zeroedin, line 108: Does it make sense to also have this input as type="number" and do a min and max range?

```Suggestion
          <input type="number" inputmode="numeric" aria-labelledby="go-to-page" min="1" max="${this.#links?.length}" value=${(this.#currentIndex ?? 0) + 1}  />
```

I would presume this should also trigger  `onchange` when the value is changed?

lib/tokens.ts
```diff
@@ -1,13 +1,13 @@
 import { unsafeCSS } from 'lit';
 
 export const mobileBreakpoint = unsafeCSS`700px`;
+export const mobilePortraitBreakpoint = unsafeCSS`320px`;
```
zeroedin, line 4: Not for this PR, but because I'm looking at this change and thinking...  we should pull these in with the token project as a dependency to get these values.

elements/rh-pagination/rh-pagination.ts
bennypowers: we don't want `type=number` since we provide our own spinners. `inputmode=numeric` suffices for the design spec. adding validations is a good idea though. I added the min max and required attrs, although they don't actually do anything, and we still need to manually validate. Still, i think there's value in exposing intent and for a few bytes of HTML it's worth the cost.

elements/rh-pagination/rh-pagination.css
```diff
@@ -0,0 +1,89 @@
...
+}
+
+input {
+  height: var(--rh-length-lg, 16px);
+  width: var(--rh-length-xl, 24px);
+  background: var(--rh-color-surface-lightest, #ffffff);
+  border: 1px solid var(--rh-color-black-300, #d2d2d2);
+  border-block-end: 1px solid var(--rh-color-black-600, #6a6e73);
+  box-sizing: border-box;
+  padding: 2px;
```
marionnegp, line 74: Width should be 64px, and height should be 32px.
Font size should be 16px.

The linked number after "of" should have #06c for the text color and no underline.

marionnegp (changes requested):
This demo made pagination easy to test!

A few things that I couldn't comment on in the code:
- [ ] On mobile, pagination is centered.
- [ ] Since the "go to page" section is going to be visible at all screen sizes, it looks like the number of pagination buttons, including arrows and truncation ellipses, will have to be limited to 9 on desktop and tablet.

bennypowers:
> * On mobile, pagination is centered.

I didn't catch this, do you mean that the paginator *shouldn't* be centered on mobile screens, or that it *should* be centered? I imagine this should be a pattern-level concern, too, but i could apply some sensible fallbacks like text-align place-self, etc.

> * [ ]  Since the "go to page" section is going to be visible at all screen sizes, it looks like the number of pagination buttons, including arrows and truncation ellipses, will have to be limited to 9 on desktop and tablet.

I didn't know the go to page spinner was supposed to be always visible, it's not like that in the XD. Please clarify

elements/rh-pagination/rh-pagination-lightdom.css
```diff
@@ -0,0 +1,131 @@
...
+  display: contents;
+}
+
+rh-pagination li {
+  display: contents;
+}
+
+rh-pagination li > a {
+  text-decoration: none;
+  color: black;
```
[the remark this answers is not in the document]
bennypowers: fixed. if it's not too difficult, please refer to token names instead of values, I had to look these up. not a huge deal, but it helps, and if the values change later, the comment will still make sense.

elements/rh-pagination/rh-pagination-lightdom.css
```diff
@@ -0,0 +1,131 @@
...
+  outline: none;
+  width: var(--_link-size);
+  height: var(--_link-size);
+  display: grid;
+  place-content: center;
+  background: var(--rh-pagination-background, #f5f5f5);
```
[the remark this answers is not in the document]
bennypowers: fixed in 37ef4f45

elements/rh-pagination/rh-pagination-lightdom.css
```diff
@@ -0,0 +1,131 @@
...
+rh-pagination:is([overflow="start"], [overflow="both"]) li:first-child:after {
+  content: "…";
+  width: var(--_link-size);
+  height: var(--_link-size);
+  padding-inline-end: 4px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background: white;
+  align-self: start;
```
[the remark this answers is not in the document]
bennypowers: --rh-font-size-body-text-lg corresponds to 20px at 16px font, using that in 55d83fb4

marionnegp:
@bennypowers, it should be centered on mobile based on [this mockup](https://xd.adobe.com/view/75d471e8-9396-4721-8e3c-cf9aa0688d57-2087/screen/62c75d2d-b0d4-4acb-af1f-a743c1f0ff73/specs/), but I see what you mean about letting whoever's using the component decide that.

Looking back at the XD, I see that the [SERP-style pagination](https://xd.adobe.com/view/75d471e8-9396-4721-8e3c-cf9aa0688d57-2087/screen/dc09b39c-d0bc-4eb8-a5b7-d605dd9bbc37/specs/) doesn't have the go to page piece, but the [data-style pagination](https://xd.adobe.com/view/75d471e8-9396-4721-8e3c-cf9aa0688d57-2087/screen/ef86d413-ef56-40b2-a350-4e96df179221/specs/) and [this search page mockup](https://xd.adobe.com/view/75d471e8-9396-4721-8e3c-cf9aa0688d57-2087/screen/92df9fcb-e049-4306-a491-a75d9a23564d/specs/) does have it. @coreyvickery or @ronniemcbride, can you clarify whether the go to page input should show up at all screen sizes?

marionnegp:
Haha, I think we felt similarly, so I just assumed that the mockup was the example to follow for desktop screens.

elements/rh-pagination/rh-pagination.ts
```diff
@@ -0,0 +1,268 @@
...
+
+        <nav ?hidden=${mobile} ?inert=${mobile} aria-label=${label}>
+          <slot></slot>
+        </nav>
+
+        <a id="next" class="stepper" href=${ifDefined(nextHref)} ?inert=${!nextHref} aria-label=${labelNext}>${L1}</a>
+        <a id="last" class="stepper" href=${ifDefined(lastHref)} ?inert=${!lastHref} aria-label=${labelLast}>${L2}</a>
+
+        ${!mobile ? '' : html`
+        <div id="numeric">
```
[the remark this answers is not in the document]
bennypowers: ok, as discussed this now always shows.

elements/rh-pagination/rh-pagination.css
bennypowers: ok done, but for the colour, I improvised the hover and visited states, PTAL

marionnegp (commented):
Question about validation: When someone tries to input a number that's higher than the number of pages, it shows a "cannot navigate to..." tooltip now. What do you think about the [way PatternFly handles it](https://www.patternfly.org/v4/components/pagination/)? Would this be difficult to implement? I know this wasn't in the specs.

elements/rh-pagination/rh-pagination.css
```diff
@@ -0,0 +1,119 @@
...
+  background: var(--rh-color-black-300, #d2d2d2);
+}
+
+#numeric {
+  margin-block-start: var(--rh-space-md, 8px);
+  width: 100%;
+  display: flex;
+  align-items: center;
+  flex: 1 1 100%;
+  gap: 0.5em;
```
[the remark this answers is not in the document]
bennypowers: np

marionnegp (commented):
Spacing changes look good!

Instant validation is kind of working. If there are 5 pages, it instantly validates when you input a single digit number that's greater than 5. 

However, it doesn't pick up when you try to input 10 or 400. It will instantly validate an input of 60 or 70.
Screen Shot 2022-10-11 at 12 30 13 PM

A similar thing happens when you have 12 pages. Inputting 130 will trigger instant validation, but 110 won't.
Screen Shot 2022-10-11 at 12 31 02 PM

bennypowers:
lol oof yeah that's because of javascript lexical string comparison. 
```js
console.assert(101 > 11)
console.assert('101' > '11') // fails
```

bennypowers:
excellent qa work, @marionnegp. please try again

marionnegp (commented):
Nice, works really well, Benny!

marionnegp (approved):

bennypowers:
@heyMP looks like tests are failing in footer can you PTAL?

PR #288, opened 2022-05-18
feat(dialog): implement `<rh-dialog>`

### LINKED ISSUE #283: [feat]: <pfe-modal> to <rh-dialog> ready to migrate
### Tag Name

rh-dialog

### RHDS Only

- [ ] The component will be **RHDS only**, delete the PFE component once published

### Name changes
When migrating `<pfe-modal>`, we should consider changing the name to `<rh-dialog>` to be more inline with [W3C](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) and community standards.

In RHDS, perhaps we should combine the terms like, "Modal Dialog" similar to [WAI Best Practices example](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html). Implementation should be `<rh-dialog>` to be closer to the HTML element.

Within PFE, we'd keep `<pfe-modal>` to match [PF4's docs](https://www.patternfly.org/v4/components/modal).

### Additional Info

We need to also bring in design updates from the old [PFE issue for modal](https://github.com/patternfly/patternfly-elements/issues/811).

bennypowers:
## What I did

1. import and extend `PfeModal`
2. Add `variant=video` attribute
3. Add `open` boolean attr
4. Position and style close button per https://ux.redhat.com/components/modal/

## Testing Instructions

1. deploy preview is available at https://deploy-preview-288--red-hat-design-system.netlify.app/components/dialog/
2. there are no regression tests or unit tests - yet

## Notes to Reviewers
1. max-width of the dialog is set to `min(90%, 1140px)`. the `1140px` part is taken from the [XD](https://xd.adobe.com/view/e925215c-e21c-4f4d-9af1-44d2fbcb7194-bc76/screen/f2408289-f60a-4370-a1a1-a2ff63042540/variables/), the `90%` part, I made up 🙈 
3. specs on [ux-dot](https://ux.redhat.com/components/modal) say the dialog should have an overlay that is
   > black solid with opacity

   Checking the provided graphic for video, the final rendered colour for the overlay was #898989 for a video dialog on a white background. Setting the video dialog overlay background colour to this got me to #8f8f8f, which was as close as I came
   ```css
   :host([variant=video][open]) [part=overlay] {
     background-color: rgba(0, 0, 0, .45);
   }
   ```
4. is `variant` the right attr?
5. What's the correct hover-state colour for the video dialog close button?

Closes #283


bennypowers:
@RedHat-UX/designers searching the rh-repo design tokens spreadsheet, I couldn't find `#6a6e73` as specified on ux.redhat.com/components/modal, but that colour was specified for the close button for a standard dialog. Should I use `--rh-color-text-secondary-on-light, #6c6c6c` instead?

bennypowers:
Ok I bumped this back down to draft for 2 reasons:
1. We need @coreyvickery 's review once he gets back from PTO
2. I did some work in pfe-modal to make it more pf1:1, so I'd like to merge that first. 
I'll link to the PR in PFE shortly

marionnegp:
> @RedHat-UX/designers searching the rh-repo design tokens spreadsheet, I couldn't find `#6a6e73` as specified on ux.redhat.com/components/modal, but that colour was specified for the close button for a standard dialog. Should I use `--rh-color-text-secondary-on-light, #6c6c6c` instead?

Sorry I missed this yesterday! @coreyvickery had adjusted the grays while making the design tokens spreadsheet, so that's likely why it's used in the component but is missing from the tokens. I think that's the closest gray, but it's probably best to check with Corey.

marionnegp:
> @coreyvickery @marionnegp @ronniemcbride do we have tokens for grid column / gutter widths? The spec defines widths in terms of number of columns, not e.g. container %.

I added a comment in the tokens spreadsheet about adding grid column and gutter widths. I'm not sure if the size tokens will work here.

bennypowers:
Rebased just now

bennypowers:
Some outstanding API questions:

- [ ] Should we have `type="video"` or would we rather `<rh-video-dialog>`.
- [ ] Should we remove the `trigger` slot (and if so, in PFE or here?) see https://github.com/patternfly/patternfly-elements/pull/2036#issuecomment-1134460631

heyMP:
> * [ ]  Should we have `type="video"` or would we rather `<rh-video-dialog>`.

I think the video type makes sense to me.  It's close enough to dialog.

> * [ ]  Should we remove the `trigger` slot (and if so, in PFE or here?) see [feat(modal): PF1:1 patternfly/patternfly-elements#2036 (comment)](https://github.com/patternfly/patternfly-elements/pull/2036#issuecomment-1134460631)

This is removed in the next release of PFE.

heyMP:
@bennypowers There is a z-index issue on the first dialog demo on the demo site I believe.

Screen Shot 2022-06-05 at 9 13 56 PM

bennypowers:
Rebased just now. @heyMP I couldn't reproduce your issue, probably it was fixed with the new dependencies? 🤷

heyMP:
Hey @bennypowers, here are some layout issues I found.

- [ ] In Chrome, scrollbar is always present in the dialog
Screen Shot 2022-06-07 at 9 54 00 AM

- [ ] Video Iframe doesn't expand the full width and height of the dialog
Screen Shot 2022-06-07 at 10 01 46 AM

elements/rh-dialog/rh-dialog.ts
```diff
@@ -0,0 +1,67 @@
...
+  render() {
+    const { mobile } = this.#screenSize;
+    return html`
+      <div id="rhds-wrapper" class=${classMap({ mobile })}>
+        ${super.render()}
+      </div>
+    `;
+  }
+
+  protected override renderHeaderSlot() {
```
heyMP, line 38: This typescript error is blocking new Netlify builds.

elements/rh-dialog/rh-dialog.ts
```diff
@@ -0,0 +1,67 @@
...
+    return html`
+      <rh-context-provider color-palette="lightest">
+        <slot name="header"></slot>
+      </rh-context-provider>
```
heyMP, line 42: Just double checking that we still need this now that we've made the dialog trigger a sibling of the dialog component.  Could this work as an alternative?

```html
<rh-band color-palette="dark">
  <rh-dialog color-palette="lightest" trigger="trigger">
    <div slot="header">...</div>
    <div slot="content">...</div>
    <div slot="footer">...</div>
  </rh-dialog>
  <rh-button id="trigger">Open Dialog></div>
</rh-band>
```

elements/rh-dialog/rh-dialog.ts
bennypowers: good catch. fixed

elements/rh-dialog/rh-dialog.ts
bennypowers: don't need it anymore. will loop back to put `color-palette` on dialog if and when it becomes a req

heyMP:
> blocked by [patternfly/patternfly-elements#2039](https://github.com/patternfly/patternfly-elements/pull/2039)

PFE issue merged. Waiting on release.

bennypowers:
@heyMP video dialog should be fixed in docs pages now. the scroll thing will require some more work

heyMP:
@bennypowers everything looks good.  The only think I noticed was that the video doesn't stop playing on dialog close.  What I've typically done before is "the video dialog option supports `<video>` tags and youtube iframes. If you need to support additional video players, use the dialog close event hook."  I kinda like that option as it limits the amount of custom JS sites would have to implement and maintain.

bennypowers:
@heyMP seems a reasonable approach to me

docs/components/dialog.md
```diff
@@ -0,0 +1,290 @@
...
+  </pfe-cta>
+</rh-dialog>
+<pfe-cta id="standard-trigger">
+  <button id="first-dialog">Open modal dialog</button>
+</pfe-cta>
+{% enddemo %}
+
+{% demo headline="Video modal dialog" %}
+<rh-dialog id="video-modal" type="video" trigger="video-trigger">
+  <iframe src="https://www.youtube.com/embed/aqz-KE-bpKQ" title="YouTube video player" frameborder="0"
```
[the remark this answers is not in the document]
bennypowers: Thanks

heyMP (approved):
Awesome job @bennypowers.  Lots of hard work went into this.  LGTM!

Issue #281, opened 2022-05-17
[feat]: <rh-stat> ready for development


markcaron:
### Tag Name

rh-stat

### Purpose

The Stat component is used for displaying statistics in an eye-catching block form.

<img width="746" alt="image" src="https://user-images.githubusercontent.com/1890950/168862456-eb0275f3-ebe9-4929-aa69-8d3a891a0b1a.png">

Stat can also be used inside other components, like `<rh-card>`.

<img width="325" alt="image" src="https://user-images.githubusercontent.com/1890950/168862556-d6f7dbe2-d74b-42d4-9760-ba007e180620.png">

[More specifications](https://xd.adobe.com/view/067ac1e2-5343-4dd7-adf4-ef3e18a91387-5329/)

### Where will it be used?

RH Stat will be used on marketing pages, customer portal pages, dashboards and other areas where displaying eye-catching statistics is important.

### End user

A customer or potential customer who would be interested in seeing statistics around a certain feature or product.

### Intended end use

A customer or potential customer would be presented with a nice visual representation of a statistic.

### Properties

[See specifications](https://xd.adobe.com/view/067ac1e2-5343-4dd7-adf4-ef3e18a91387-5329/)

RH Stat needs a few slots:

1. Statistic (text, number)
2. Title (optional)
3. Caption / short description / body text
4. Icon (optional, above statistic always)

Options:

- Position of title can be above or below the statistic.
- Sizes: default and large
- Theme: on light and on dark
- Text is centered (always)

May need to discuss inclusion of CTA due to spacing requirement below the stat.

**Examples:**

<img width="554" alt="image" src="https://user-images.githubusercontent.com/1890950/168869531-76b23a58-7d8b-4bff-8598-32c91b9ed2da.png">

<img width="430" alt="image" src="https://user-images.githubusercontent.com/1890950/168867302-fbd1aa0b-6a26-4665-8d4c-ab3737527030.png">

### Page Builder

- [ ] My component **will** be part of a page-builder experience
- [X] I'm not sure

### Accessibility Features

- [X] **Color**: Color should not be the primary visual means of communicating information ([1.4.1](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html))
- [X] **Content or microcopy** Text in any language should be clear, meaningful, structured, etc.
[...]

PR #330, opened 2022-05-25
feat(stat): implement `rh-stat`

brianferry:
## What I did

1. Created a component for `rh-stat` based off of the design specs located at https://xd.adobe.com/view/067ac1e2-5343-4dd7-adf4-ef3e18a91387-5329/. 

## Issue Link
(#281)

## Testing Instructions

1. Deploy preview available at https://deploy-preview-330--red-hat-design-system.netlify.app/components/statistic/
2. There are currently spec.ts tests for the component to make sure the slots are loaded

## Notes to Reviewers
1. There are two sets of demos, one without a theme and the other with the dark theme added through a `pfe-band` around the component.


elements/rh-stat/rh-stat.css
```diff
@@ -4,9 +4,9 @@
 
 :host {
     display: flex;
-    --accent-color: var(--rh-footer-accent-color, #e00);
-    --red-hat-text:  var(--rh-theme--font-family--heading, 'Red Hat Text');
-    --red-hat-text-display: var(--rh-theme--font-family--heading, 'Red Hat Display');
+    --accent-color: var(--rh-theme--color--ui-accent, #e00);
```
bennypowers, line 7: nit: `style` commit type is actually for _code style_ i.e. refactorings, lintings, things like that.

CSS changes should have `fix:` commit type. Confusing? yeah, just one of those things

brianferry: Got it, that makes sense. I was curious why style was mentioning white-space but didn't seem like the other categories really fit either.  Will keep in mind for next time.

brianferry:
> * Icons should have max-width and max-height of 40px.
> * Stat in dark mode should use brand red (#E00). It'll have enough color contrast because the text is large.
> * Placeholder stat titles aren't showing up for me.
> * What's the spacing between the icon and the stat? I can see the 8px of padding below the icon, but it looks like there's a bit more. Total spacing between the two should be 16px.
> 
> Feel free to message me if you have questions!

- Sent you a message about the icons, md is 32px and lg size is 64px.  We can probably fix this with CSS but wanted to check with you beforehand to double check that we wanted 40px or if we meant 32px
- Switching that back now
- Sent a message about this too, basically just trying to check how we want the user to use the API with the component.  Since the component can (at this time) include both an icon / title, just an icon OR just a title it makes sense to me to leave them blank and allow the user to put in whichever they want to show up.  
- Spacing between the sections is 8px and the icon has an additional padding on the bottom giving it 16 px of spacing between it and the section below it.

brianferry:
@marionnegp - Please check / confirm that the icons for md sized icons are now showing up as 40px for you.

marionnegp:
- Yep, the icons are showing up as 40px now. Thanks!
- I realized today that the red for the dark theme stat was supposed to be #FF3333 for good color contrast. *facepalm* Sorry to have you change that back! It's not currently one of our color tokens, but I talked to @coreyvickery about adding it as an option for our dark theme.
- I think the title slot is adding extra space between the icon and the stat when it's not there. Is there a way to prevent that?

marionnegp:
Are the titles using a medium font weight? It looks thicker in XD, but I think there are sometimes differences between how the fonts look in XD vs on our site.

brianferry:
@marionnegp - Hey Marionne, quick update.  So I've switched back the dark theme color so it should be #FF3333 again.  I explicitly added the font-weight to the title but it looks like that was the value beforehand as well.  I've also made changes to improve the spacing so the spacing should look correct in the demo now.

brianferry:
Also to update, I've added a few more demo examples so we can really see how many different configurations of the `rh-stat` component exist.  I put them into four main groups:
- Default
- Large
- Dark Theme
- Statistic Above Title

Inside of these groups are the different configurations of icon, title, description, and statistic.  Since icon can come in either as an attribute or a slot there are different sections to make sure all of the styling lines up either way. 

Let me know any questions you have!

marionnegp:
@brianferry, thanks for the additional demos! Those were helpful to see.

A couple things:
- I think stat ended up getting a medium weight after the last demo, but it should have a regular weight. Might have been because the font-weight was explicitly added. Since the font weight for title was already medium before, I don't mind if you drop that. Sometimes our site fonts don't match up exactly to how they look in XD. :-/
- Row-gap seems to be adding 8px to every row, but there isn’t a top or bottom margin between the stat and title in the XD. Does this spacing have to be here? The spacing from the XD doc should make each line look more uniformly spaced.
Screen Shot 2022-06-07 at 2 38 48 PM

brianferry:
@marionnegp @coreyvickery - I see what my problem was now, I was seeing the spacing between the description and the other pieces and not realizing that the spacing was different between title and the statistic.  There should now be a margin-bottom on the icon and a margin-top on the description and no spacing between the title & the statistic.  

I've also made the whole component have the regular font-weight of '400' so everything should look correct now I think?  It's a little difficult to tell on my screen the difference between 400 / 500 but 300 looks very different but the docs show 400 as being the regular font-weight. 

Thank you both again for the notes & feedback.  This is super helpful!

marionnegp:
Yeah, I'd say the stat should be 300, and the title should be 400. The title's using the same style as "Layout Title" on https://ux.redhat.com/foundations/typography/.

brianferry:
@marionnegp @coreyvickery - Very cool, updated the title to have 400 font weight and stat to have 300! Thank you so much again Marionne and hope you have a great time on PTO

marionnegp (approved):
Updates to stat from June 8 look good! Thanks for your hard work and your patience, @brianferry!

brianferry:
@marionnegp - Do we want to update or remove the docs page before we merge in these changes or is the components page good for now?

marionnegp:
The CTA should have a space of 24px between it and the body copy. Not sure why the XD link is showing a 26px margin. :-/ 

@coreyvickery, did you want Brian to add the card version now? We can hold off if you want to revisit Krystal's stat card design. (Brian, for context, the Customer Portal team has a stat component in a card for their dashboard. They use it a little differently than we do, so we initially didn't align our designs very closely.)

marionnegp:
We're not moving forward with a stat in a card right now. The rest of it looks good to go!

PR #343, opened 2022-05-31
docs: copy lightdom styles to web root

bennypowers:
## What I did

1. copy component lightdom styles to web root when building eleventy

## Testing Instructions

1. link a component's lightdom styles in it's docs page, eg
   ```html
   <link rel="stylesheet" href="{{ '/assets/rh-dialog-lightdom.css' | url }}">
   ```

Taken from #327

### DIFF
--- a/.eleventy.cjs
+++ b/.eleventy.cjs
@@ -17,7 +17,9 @@ const markdownItAnchor = require('markdown-it-anchor');
 const pluginToc = require('@patternfly/pfe-tools/11ty/plugins/table-of-contents.cjs');
 const sassPlugin = require('eleventy-plugin-dart-sass');
 
+const glob = require('glob');
 const path = require('path');
+const fs = require('fs');
 
 const markdownLib = markdownIt({
   html: true,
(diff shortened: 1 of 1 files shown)


.eleventy.cjs
```diff
@@ -139,6 +134,21 @@ module.exports = function(eleventyConfig) {
             .filter(x => x && x !== '@patternfly/pfe-styles'),
         ],
       })));
+  /** Copy any component lightdom styles to the `assets` dir in the web root */
+  eleventyConfig.on('eleventy.before', async () =>
+    Promise.all(glob.sync('element/*/*-lightdom.css').map(x =>
+      fs.copyFile(x, path.join('_site', 'assets', path.basename(x))))));
```
zeroedin, line 140: A couple of issues are resolved with this suggestion.  

- Corrects typo for base folder `elements/*` instead of `element/*`.
- The glob pattern needed `**` in the pattern `elements/**/*-lightdom.css` for subdirectories search
- `fs.copyFile` third parameter is a callback  unless you use `fs/promises` if you'd rather use the callback less version include `const fs = require('fs/promises');` and replace the callback version with `fs.copyFile(filePath, destPath);`

After the change is added `rh-footer-lightdom.css` is correctly moved to `_site/assets/lrh-footer-lightdom.css`


```Suggestion
  eleventyConfig.on('eleventy.before', async () =>
    Promise.all(glob.sync('elements/**/*-lightdom.css').map(filePath => {
      const destPath = path.join('_site', 'assets', path.basename(filePath));
      fs.copyFile(filePath, destPath, err => {
        if (err) {
          // eslint-disable-next-line no-console
          console.log('Error Copying lightdom css', err);
        }
      });
    })));
```

zeroedin (commented):
Is the directory these lightdom files are placed in the `_site` structure up for debate.  There is an already existing `css` folder in the site structure, should we use that?.

Otherwise, with the suggested fix this PR looks good to me.  @coreyvickery since Benny added you as requested review ill leave you the honors of approval.

.eleventy.cjs
bennypowers: you got it!

bennypowers:
Superceded by #435

bennypowers:
light dom css (and all element assets) is available at, for example https://ux.redhat.com/assets/elements/rh-footer/rh-footer-lightdom.css

PR #353, opened 2022-06-02
Fixes missing column in Component status table

coreyvickery:
## What I did

The WebRH column was missing in the `Component status` table, this is a fix for that


heyMP (approved):
LGTM 👍

PR #388, opened 2022-06-12
fix: Feat/dialog mp video fixes

heyMP:
## What I did

1. In Chrome and Safari there the video iframe wasn't expanding the full width of the dialog.
Screen Shot 2022-06-12 at 12 39 48 PM

2. Added a calculation that prevents the video player from overflowing off the bottom of the screen.  Accounts for max-hight of the viewport and aspect ratio.
Screen Shot 2022-06-12 at 12 40 54 PM

3. I noticed that there was an occasional 1px discrepancy between the dialog and the container of the video player so I added a 1px buffer so you don't see the background of the dialog bleed through.
 
Screen Shot 2022-06-12 at 12 43 39 PM

## Testing Instructions

1. Visit `http://localhost:8000/demo/rh-dialog/` in Chrome or Safari
2. Open the video dialog example
3. Resize your browser between different aspect ratios to ensure the video player is completely visible at all times.


bennypowers (approved):
Lather greatly, these men!

Thanks for catching this @heyMP

PR #390, opened 2022-06-14
feat: change screen size queries to min-width

zeroedin:
## What I did

1. Converted media queries to `min-width` based queries for standard breakpoints.  I feel this is the proper approach for responsively designed websites.  Breakpoints matching css cascading to larger sizes so that it is additive : 
```
[no-break-point] 
 ⬇️
[320px] /* mobilePortraitBreakpoint */
 ⬇️
[576px] /* mobileTabletBreakpoint */
 ⬇️
[768px] 
 ⬇️
[992px]
 ⬇️
[1200px] 
 ⬇️
etc...
```
```css

/* no breakpoint */
div {
 color: blue;
}

@media and screen(min-width: 320px) {
  /* mobilePortraitBreakpoint */
  div {
    padding: 0.5em;
  }
}

@media and screen(min-width: 576px) {
  /* mobileTabletBreakpoint */
  div {
    padding: 1em;
  }
}

etc...

```

2. Ordered tokens smallest to biggest, and added comments for what I think might be special breakpoints tokens which are based on `max-width` such as `mobile` and `mobileXL`.  
3. Added `matches` property to return all current  matching `BreakPointKey`.  Resulting in a stack such as `['mobile', 'mobilePortrait', 'mobileLandscape', 'mobileLandscape', 'tabletPortait']`  if the screen size is `< 992px` as defined in the token value: 
```
export const tabletLandscapeBreakpoint = unsafeCSS`992px`;
```

The implementation of `matches` allows you to set a class based on if a specific breakpoint is or isn't in the stack, then we know we've gone beyond the size in which the interface needs to react.

```
/* Looking for if `tabletLandscape` breakpoint has been reached. If not we add `is-mobile` to classes */
const mobile = !this.#screenSize.matches.has('tabletLandscape');
const classes = { 'is-mobile': mobile };
```

The reasoning behind this change is that the `size` attribute is only set to the current breakpoint giving nothing to base the logic of checking all matched breakpoints from if the current state returned was less than the `tabletLandscapeBreakpoint`.  Because `min-width` is additive, the `matches` will include all breakpoints that match.

This is also working around the idea of `this.mobile` in the context of the controller which is based on the max-width breakpoint value of `700px` defining what is or isn't mobile.  I'm not convinced we should have a specific definition of `mobile` at a specific breakpoint, rather being more content-specific on when things shift.  

In the case of `secondary-nav` that content specific happens under 992px per spec due to limited width for content at a smallish screen size.  Not that 992px specifies what is and what isn't `mobile`.

### DIFF
--- a/lib/RHDSScreenSizeController.ts
+++ b/lib/RHDSScreenSizeController.ts
@@ -26,11 +26,11 @@ export class RHDSScreenSizeController implements ReactiveController {
 
   static queries = new Map<BreakpointKey, MediaQueryList>([
     ['mobile', matchMedia(`screen and (max-width: ${mobileBreakpoint})`)],
-    ['mobilePortrait', matchMedia(`screen and (max-width: ${mobilePortraitBreakpoint})`)],
-    ['mobileLandscape', matchMedia(`screen and (max-width: ${mobileLandscapeBreakpoint})`)],
-    ['tabletPortait', matchMedia(`screen and (max-width: ${tabletPortaitBreakpoint})`)],
-    ['tabletLandscape', matchMedia(`screen and (max-width: ${tabletLandscapeBreakpoint})`)],
     ['mobileXl', matchMedia(`screen and (max-width: ${mobileXlBreakpoint})`)],
+    ['mobilePortrait', matchMedia(`screen and (min-width: ${mobilePortraitBreakpoint})`)],
+    ['mobileLandscape', matchMedia(`screen and (min-width: ${mobileLandscapeBreakpoint})`)],
+    ['tabletPortait', matchMedia(`screen and (min-width: ${tabletPortaitBreakpoint})`)],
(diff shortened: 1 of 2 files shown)


bennypowers:
@zeroedin happy for you to fold this into your secondary-nav PR, so long as it doesn't break any other components.

zeroedin:
> @zeroedin happy for you to fold this into your secondary-nav PR, so long as it doesn't break any other components.

As it stands right now, the only other component that is actively using `RHDSScreenSizeController` is `rh-dialog`.  It uses the mobile max-width breakpoint, which this PR does not change.

I'll fold this into `feat/secondary-nav`.  I'll leave this PR open as a draft for any further comments or considerations.

zeroedin:
Closing this as it has been implemented.

PR #396, opened 2022-06-16
docs: build status tables from component docs

bennypowers:
## What I did

1. copy the data from https://drive.google.com/u/0/open?id=1RINxMTusHWQ7m7Dljis96c6D44CHgtHprQ7mUhWd-Hc into component docs pages
2. remove the rhds-component-status webcomponent
3. add a `componentStatus` shortcode

## Testing Instructions

1. check the deploy preview, ensure status tables on component pages look correct

## Notes to Reviewers
I'd like to review the categories here, i'm not sure how accurately the table headings (PFE repo, adobe target, etc) reflect our plans anymore.


bennypowers:
> @coreyvickery @marionnegp Did we decide to go with X marks instead of Check marks? Also, we might be considering changing the marks used in the future. It is possible we might change it to a circle with a check-in from our icon set. @bennypowers swapping the mark for an RH-icon replacement should be considered. Will that be an issue later?
> 
>     * The current X marks seem a tad off-center and to the right.
> 
>     * I would  move the "Last updated: 6/10/2020" text to the right of the table

Date updated text moved! 
ATM it just dumps the text contents of the cell into the template, but changing that later wouldn't be hard. Perhaps we can save the x=>${thing} update for a later addition?

bennypowers:
@marionnegp `link with icon` and `link` should work as expected now

marionnegp:
@bennypowers, the `link` page looks good now!

Issue #392, opened 2022-06-14
[bug] Minor design edit to `rh-footer`


coreyvickery:
### Description
* Remove the horizontal rule from under the logo and social media icons on small screen sizes

### Before
![_Footer on small screens - before](https://user-images.githubusercontent.com/45767595/173675873-5fe7d68b-1b8d-41a9-96d4-134fede00e0c.png)

### After
![_Footer on small screens - after](https://user-images.githubusercontent.com/45767595/173675902-3ab7fc0a-6524-4232-88c4-368cf3c78fc8.png)

### Which Component?

rh-footer

### Steps to reproduce

Go [here](https://ux.redhat.com/components/footer/) to see image examples or go [here](https://rh-footer-demo.vercel.app/) and play squishy window.

### Expected behaviour

Horizontal rule should be not be visible on small screen sizes

### Operating System (OS)

macOS

### Browser

Chrome

Issue #371, opened 2022-06-07
[feat] rh-footer should be able to display global footer only


heyMP:
### Description

We missed this feature in the initial release of rh-footer but there are some cases where only the global footer that contains the legal links is needed.  Currently we don't have a method of doing that easily.  This is because the 'header' section of the footer has fallback content if no slots are provided and because the 'main' section of the component has height due to grid gap.  This means that even if there is no content for 'main' and 'header' slots defined by the user, they will still see those sections.

### Suggested solution

We should use the SlotController to hide sections that don't contain user provided content for the 'header', 'main', and 'footer' sections of the rh-footer component.

### Additional context

https://deploy-preview-341--red-hat-design-system.netlify.app/components/footer/#global-footer

![Screen Shot 2022-06-07 at 9 34 12 AM](https://user-images.githubusercontent.com/3428964/172393639-f5427b5c-0926-4afb-b8d2-a0b138890a49.png)

bennypowers:
What if 
```html
<!-- happy path -->
<rh-footer>
  ... insert normal footer content here
  <rh-global-footer slot="footer">...insert global content here</rh-global-footer>
</rh-footer>

<!-- global only -->
<rh-global-footer>...insert global content here</rh-global-footer>
```

heyMP:
Could do that too.  Normally I would go with a dedicated component but it looks like we're leaning towards using slots to do more of the heavy lifting. 

Example: https://github.com/RedHat-UX/red-hat-design-system/issues/281#issuecomment-1134698422

bennypowers:
yeah i still think that's the way to go, generally speaking

but in this particular case the request was for *only* the global part
so we can save some bytes on those pages by vending a separate component

heyMP:
@bennypowers good point about the saving bytes consideration.  We'll also have to make sure that we do the following:

- [ ] Ensure that the `rh-footer-global` is not bundled with `rh-footer.js`

bennypowers:
will be a la carte just like the rest of the lib if we
```js
import '@rhds/elements/rh-footer/rh-global-footer.js';
```

Will it need lightDOM styles?

heyMP:
It will need lightdom styles since there are nested lists.  We have two options.

[...]

Issue #312, opened 2022-05-24
[feat]  <rh-footer> add font-size revert instead of px value for


heyMP:
### Description

rh-footer explicitly sets the font-size of the :host component to `18px`.  This was a bandaid to account for sites that had different root font-sizes.

### Suggested solution

We should do this instead:

```
What we came up with was reverting the font-size: revert on :host will set the components base back to the users setting.  I figure we can trap door this with var(--rh-component-base-font-size, revert) or whatever
```
cc: @zeroedin @markcaron

heyMP:
@zeroedin is checking on the best corse of action on this.

bennypowers:
do we need a variable for this? Users that want to change the default should just
```css
rh-footer {
  font-size: 25em; /* 🌶 */
}
```

zeroedin:
The fix I worked in for `rh-secondary-nav` resets the font-size to user preference, then uses `em` to style component sizing.

```
:host {
  font-size: var(--rh-secondary-nav-base-font-size, initial);
}
```
This resets to initial but also allows a trap door if someone is adamant about setting that value.

As for your question benny, do we need a variable? No probably not.   Your solution is most likely more valid just override directly in light dom.

bennypowers:
Closed by #406

PR #406, opened 2022-06-22
feat(footer): add rh-global-footer and UI updates

heyMP:
## What I did

1. #371 
2. #312 
3. #392

### Setup
1. `git checkout origin/feat/rh-global-footer`
2. `npm ci`
3. `npm run dev`
4. Visit http://localhost:8000/demo/rh-footer/

### Global footer
- [ ] The global footer in both the "Kitchen Sink" and the "Global Footer" demos should look and function identically.

### Remove space between header and mobile menu
- [ ] Verify that the mobile menu is now flush with the header according to the design specs.
- [ ] Verify that the mobile menu still works as expected. (See screenshots in. #392)

### Move from `px` based units to `ems` for accessibility
- [ ] Verify that the fonts-sizes still match the design specs.
- [ ] Verify that the user can control the size of the text by changing the browsers default font-size.
Example: 

https://user-images.githubusercontent.com/3428964/175962372-e43a5d49-48bc-4596-b8ad-db2ebb82ea38.mov


elements/rh-footer/README.md
```diff
@@ -11,6 +11,10 @@ npm install @rhds/elements
 import '@rhds/elements/rh-footer/rh-footer.js';
 ```
 
+```html
+<link rel="stylesheet" href="https://unpkg.com/@rhds/elements/elements/rh-footer/rh-footer-lightdom.css" />
```
heyMP, line 15: Adding example for how to pull in css file off of a CDN

elements/rh-footer/README.md
```diff
@@ -20,8 +24,8 @@ Via CDN with import maps (_recommended_):
 <script type="importmap">
 {
   "imports": {
-    "@rhds/elements": "https://ga.jspm.io/npm:@rhds/elements@1.0.0-beta.5/rhds.min.js",
-    "@rhds/elements/rh-footer/rh-footer.js": "https://ga.jspm.io/npm:@rhds/elements@1.0.0-beta.5/elements/rh-footer/rh-footer.js"
+    "@rhds/elements": "https://ga.jspm.io/npm:@rhds/elements@1.0.0-beta.9/rhds.min.js",
+    "@rhds/elements/rh-footer/rh-footer.js": "https://ga.jspm.io/npm:@rhds/elements@1.0.0-beta.9/elements/rh-footer/rh-footer.js"
```
heyMP, line 28: Update docs to next version.

elements/rh-footer/RhFooter.ts
```diff
@@ -42,38 +42,13 @@ function isHeader(tagName: string) {
  * @csspart logo
  * @slot    social-links
  * @csspart social-links
- * @slot    social-links-start
- * @csspart social-links-start
- * @slot    social-links-end
- * @csspart social-links-end
```
heyMP, line 48: I'm removing these regions as social links will no longer be default in the component.

elements/rh-footer/RhFooter.ts
```diff
@@ -42,38 +42,13 @@ function isHeader(tagName: string) {
...
- * @slot    footer-secondary-start
- * @csspart footer-secondary-start
- * @slot    footer-secondary-end
- * @csspart footer-secondary-end
- * @slot    footer-links-primary
- * @csspart footer-links-primary
- * @slot    footer-links-secondary
- * @csspart footer-links-secondary
- * @slot    footer-tertiary
- * @csspart footer-tertiary
```
heyMP, line 76: Moving everything with footer- prefix into the `rh-global-footer`.

elements/rh-footer/rh-footer-lightdom.css
```diff
@@ -1,40 +1,42 @@
...
-  font-size: var(--_font-size);
-  text-decoration: none;
+  color: var(--link-color, #fff) !important;
+  font-size: var(--link-font-size, 0.875em);
 }
 
-rh-footer :is([slot^=links], [slot=footer-links-primary], [slot=footer-links-secondary]) a:is(:hover, :focus-within) {
-  text-decoration: underline;
+rh-global-footer [slot^=links] a {
+  font-size: inherit;
```
heyMP, line 23: We are telling these links that they should inherit there font-size from inside the web component

elements/rh-footer/rh-footer-block.css
```diff
@@ -15,13 +15,13 @@
 
 ::slotted(*) {
   color: #fff;
-  font-size: 14px;
+  font-size: 0.875em;
```
heyMP, line 18: We're replacing all px values to `ems` for a11y.

elements/rh-footer/rh-footer-lightdom.css
```diff
@@ -1,40 +1,42 @@
-rh-footer :is([slot^=links]) a {
-  gap: var(--rh-footer-links-gap, 10px);
+/* ensure links fully wrap img tags */
+:is(rh-footer, rh-global-footer) a[slot^=logo] {
+  display: block;
+}
+
+rh-footer [slot^=links] a {
+  gap: var(--rh-footer-links-gap, 0.625em);
```
heyMP, line 7: I'm not overlay thrilled about setting gap / padding / margins to `ems`.  Seems like it would make it more difficult to translate the design specs to code. 🤷‍♂️

elements/rh-footer/rh-footer-responsive.css.ts
```diff
@@ -92,20 +92,11 @@ export const responsiveStyles = css`
...
-    /* Add a bit more margin to the primary content on mobile */
-    .main-primary {
-      margin: calc(
-          var(--pf-global--spacer--2xl, 48px) -
-            var(--pf-global--spacer--xl, 32px)
-        )
-        0;
-    }
-
-    .footer-logo {
```
heyMP, line 104: Since we are collapsing the space between header and the accordion on mobile, we no longer need this.

elements/rh-footer/rh-footer.css
```diff
@@ -5,9 +5,6 @@
   --_accent-color: var(--rh-footer-accent-color, #e00);
   --_section-side-gap: var(--rh-footer-section-side-gap, var(--pf-global--spacer--xl, 32px));
 
-  /* PFE 2.0 ContextControllerAPI: set context for child components */
-  --context: dark;
-
```
heyMP, line 10: This is a deprecated API.

elements/rh-footer/rh-footer.css
```diff
@@ -16,18 +13,19 @@
...
   /* apply sensible defaults based on redhat standards. */
   color: #fff;
   font-family: 'Red Hat Text', 'RedHatText', 'Overpass', Overpass, Arial, sans-serif;
-  line-height: 125%;
+  line-height: 150%;
   font-weight: 300;
 
   /* set at 18px for margin and padding standardization */
-  font-size: 18px;
+  font-size: initial;
```
heyMP, line 28: This is the meat and potatoes of making ems work properly. It overrides the websites base font-size and uses the user's browser settings.

elements/rh-footer/rh-global-footer.ts
```diff
@@ -0,0 +1,110 @@
...
+@customElement('rh-global-footer')
+export class RhGlobalFooter extends LitElement {
+  static readonly styles = [style, responsiveStyles];
+
+  @colorContextProvider()
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette: ColorPalette = 'darker';
+
+  render() {
+    return html`
+      <div class="section global-base" part="section base">
```
heyMP, line 49: You might be wondering "why are only the class names prefixed with `global`?".  It's because we are using a shared .css file so we unfortunately have to do that.  But it's all internal to the shadowdom so it's not that bad I don't think. :)

elements/rh-footer/test/rh-footer.spec.ts
```diff
@@ -2,15 +2,26 @@ import { html } from 'lit';
 import { fixture, expect, nextFrame, aTimeout } from '@open-wc/testing';
 import { setViewport } from '@web/test-runner-commands';
 import { RhFooter } from '../RhFooter.js';
+import { RhGlobalFooter } from '../rh-global-footer.js';
```
heyMP, line 5: I'm adding some basic tests for the global footer.  This element would benefit greatly with some visual regression snapshots. 😀

lib/MatchMediaController.ts
```diff
@@ -25,6 +25,7 @@ export class MatchMediaController implements ReactiveController {
   }
 
   hostConnected() {
+    this.evaluate();
```
heyMP, line 28: This fixes the problem of the mobile accordion not showing up on initial load.  The user would have to resize the screen to desktop and then back to mobile width to render the mobile accordion menu.

elements/rh-footer/rh-footer-block.css
bennypowers: Grab this from tokens?

lib/MatchMediaController.ts
bennypowers: WDYT about working with @zeroedin to replace this with screensize controller?

elements/rh-footer/README.md
bennypowers: I'm thinking about adding a prepublishOnly step which copies all the `/elements/*/*-lighdom.css` to `$1/$2-lightdom.css`. Doesn't fill me with joy though

elements/rh-footer/rh-footer-block.css
heyMP: I think we need to break this out as a discussion.  The issue is that currently we can't use `rems` on RHDC.  RHDC sets an odd root font-size that will mess with all of our `rem` value tokens like so:

Screen Shot 2022-06-28 at 2 31 50 PM

heyMP: Moved to the Design Tokens discussion: https://github.com/RedHat-UX/red-hat-design-system/discussions/385#discussioncomment-3042873

elements/rh-footer/README.md
bennypowers: https://github.com/RedHat-UX/red-hat-design-system/issues/418

elements/rh-footer/RhFooter.ts
```diff
@@ -178,65 +130,7 @@ export class RhFooter extends LitElement {
               </div>
             </slot>
           </div>
-          <div class="section footer" part="section footer">
```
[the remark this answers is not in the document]
bennypowers: yeah i'm good with this too

bennypowers:
Sent some changes today after reviewing:
- documented slots and parts in jsdoc
- documented css properties (was this intended?)
- fixed the part name for social-links
- updated the README
- removed `@pfelement`
- changed `--link-header-font-size` to `--rh-footer-link-header-font-size` and documented it
- preferred ecmascript `#private` fields
- privatized isMobile
The biggest change I made was that last one, to remove the `is-mobile` public attribute and instead set an `isMobile` class on the container `<footer>`, along with attendant shadow CSS changes. This removed the need to track and react to state changes with `@property() isMobile` as well as the need to use the `onChange` callback from MatchMediaController

@heyMP if these changes suit you, I'm satisfied, just please *squash* and merge with something like
```
feat(footer): add `<rh-global-footer>` 
```

bennypowers (approved):
Learning gastronomy: tastes meaty!

elements/rh-footer/rh-footer.css
coreyvickery: @heyMP Not sure where to put this feedback, but the column titles (`Products`, `Tools`, `Try, buy, sell`, etc.) are rendering `"Red Hat Display"` instead of `"Red Hat Text"`. Everything else looks fantastic.

coreyvickery (commented):
@heyMP Left feedback below line 30, not sure where else I should have put it. Once complete, approved.

elements/rh-footer/rh-footer.css
bennypowers: I checked this using the local dev server, and this was the result:
Screen Shot 2022-07-05 at 10 04 44

In this screenshot, the heading elements use the `"Red Hat Display"` font, as specified in the page's (light DOM) `main.css`.

I added a CSS rule which sets the font-family for slotted headings to the `--rh-font-family-heading` token, using `!important` to overrule page CSS. The result was this screenshot, where you can see the headings use the `"RedHatDisplay"` font, which, confusingly, is a different font.

Screen Shot 2022-07-05 at 10 11 14

@heyMP I pushed this commit, but feel free to revert if you have a better solution. `!important` is perhaps a bit upsetting, but OTOH, users can always set `--rh-font-family-heading` - we also take this approach in button.

heyMP: Sounds good.  We've been using !important when absolutely necessary.  Makes sense to me.  Thank you~!

elements/rh-footer/rh-footer-responsive.css.ts
```diff
@@ -163,7 +163,7 @@ export const responsiveStyles = css`
   }
 
   @media screen and (min-width: ${mobileBreakpoint}) {
-    :host([is-mobile]) .links ::slotted(ul) {
+    .isMobile .links ::slotted(ul) {
```
heyMP, line 166: We moved `is-mobile` state into the shadowroot

elements/rh-footer/rh-footer.css
```diff
@@ -198,94 +198,94 @@ pfe-accordion {
   gap: 1.5em;
 }
 
-:not(.isMobile) .links {
+[part="base"]:not(.isMobile) .links {
```
heyMP, line 201: The above selector was not working for some reason: 🤔

zeroedin (approved):
LGTM

PR #416, opened 2022-06-28
Removing any mention of UI icons

coreyvickery:
## What I did

1. Removed any mention of UI icons because they are not included in the icon library anymore


marionnegp (approved):
Looks good

PR #438, opened 2022-07-20
chore: build with typescript

### LINKED ISSUE #429: [bug] include global footer in npm package
We neglected to include rh-global-footer.js in the latest release. 

_Originally posted by @bennypowers in https://github.com/RedHat-UX/red-hat-design-system/discussions/420#discussioncomment-3134921_

bennypowers:
esbuild has [difficulty](https://github.com/evanw/esbuild/issues/708) with cases like "bundle these entry points, but don't bundle these specific files, treat them as separate imports, and don't bundle them into any other files that import them". This makes it hard to load `rh-gllobal-footer.js` separately.

Closes #429 
See https://github.com/RedHat-UX/red-hat-design-system/discussions/420#discussioncomment-3174377 (nice)
cc @eyevana

## What I did

1. This commit replaces the per-element esbuild builds with a typescript build that uses a
[typescript transform](https://npm.im/ttypescript) to replace import specifiers ending with `css` with `css.js`, 
2. uses the node filesystem api to directly build shadow DOM css into their own javascript modules.

## Testing Instructions

```
npm run build
```

check that the build is sane, try copying it into a separate node project and see that it works as expected. check that 11ty builds as expected

### DIFF
--- a/.changeset/cool-parents-rescue.md
+++ b/.changeset/cool-parents-rescue.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": minor
+---
+
+Ship components unbundled
--- a/.eleventy.cjs
+++ b/.eleventy.cjs
@@ -128,8 +128,10 @@ module.exports = function(eleventyConfig) {
     [`${path.dirname(require.resolve('@patternfly/pfe-styles'))}/*.{css,css.map}`]: 'assets'
   });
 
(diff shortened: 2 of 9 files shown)


zhawkins:
Ship it 🚢

zeroedin (approved):
Linux Gnu Torvalds Makefile

PR #440, opened 2022-07-21
chore: add lib directory to npm package

zeroedin:
## What I did

1. Added lib directory to `package.json` files block

## Notes to Reviewers
I was unsure if I needed to run `npm i --package-lock-only` to update the package-lock.json.  I don't believe the files block is referenced there.


bennypowers (approved):
Lib gets tarballed, mate

PR #444, opened 2022-07-26
[rh-stat]: accessibility & font color fixes

brianferry:
## What I did

1. I have added two new demos for the proposed font color changes for rh-stat for accessibility on dark mode backgrounds. 

## Testing Instructions

1.


elements/rh-stat/rh-stat.css
```diff
@@ -1,5 +1,5 @@
 :host([on='dark']) {
-    --accent-color: var(--rh-color-brand-red-on-dark, #ff3333);
+    --accent-color: var(--rh-color-brand-red-on-dark, #fff);
```
bennypowers, line 2: Instead of changing the default value for the `--rh-color-brand-red-on-dark` token, which is defined in the tokens repo, you should set the value directly on `--accent-color`. I also suggest renaming to either `--rh-stat-accent-color` (if intended to be public) or `--_accent-color` (if intended to be private). In the latter case, I also suggest assigning the property on the shadow container, instead of the host.

This should maybe wait until after https://github.com/RedHat-UX/red-hat-design-system/pull/471 then rebase on top of that

brianferry: Updated this based on the feedback & rebased.

bennypowers (changes requested):
please add a changeset

bennypowers:
looks like this needs to be rebased

.changeset/proud-ears-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Fixing rh-stat dark theme font color for accessibility.
```
bennypowers, line 5:
```suggestion
Stat: Fixes dark theme font color contrast.
```

bennypowers (approved):

PR #457, opened 2022-08-04
fix(secondary-nav): css style fixes

zeroedin:
## What I did

1.  Fixed missing font-family stacks when base css isn't applied
2.  Added rh-token CSS custom properties for font family stacks
3.  Fixed regression in spacing for slotted cta


heyMP (approved):
LGTM 👍

Issue #432, opened 2022-07-14
[bug] <rh-stat> rh-stat has an accessibility issue with gray backgrounds


brianferry:
### Describe the bug

The `rh-stat` component has an accessibility issue with gray backgrounds (specifically #3C3F42) where the red font is not legible from the gray background. 

### Which Component?

rh-stat

### Steps to reproduce

1. Pull down the RHDS project
2. Build the project and start the dev server
3. Change one of the demo backgrounds in `rh-stat` to #3C3F42

### Expected behaviour

The text needs to pass the accessibility tests for gray backgrounds

### Operating System (OS)

macOS

### Browser

Chrome

coreyvickery:
@brianferry Here are the design fixes we are proposing.

https://xd.adobe.com/view/bf14fb6e-fd1e-4869-98ee-14deeb2428b3-ea47/

brianferry:
@coreyvickery @marionnegp 

I have added demos so that you can take a look at what the options would look like for the demo.  These demos are both overriding the font color for all dark background rh-stat elements.

Link: https://deploy-preview-444--red-hat-design-system.netlify.app/components/statistic/ 

The first one uses #FDDBDB which I got from https://red-hat-design-tokens.netlify.app/ which is the token $rh-color-brand-red-lightest. 

The second uses #D2D2D2 outlined in the UX doc which has the token $rh-color-text-secondary-on-dark

coreyvickery:
@brianferry 

On `#3c3f42` backgrounds:
- [ ] Stat should be white
- [ ] Body text should be #d2d2d2

Then we're done.

coreyvickery:
@brianferry Let's make everything white for dark theme including the icon, title, stat, and body text. Can you please update the demo page to reflect this?

brianferry:
@coreyvickery @marionnegp - The demos have now been updated to have a #fff font everywhere.

coreyvickery:
@brianferry Can we close this or do we need to update the documentation page or somewhere else?

brianferry:
@coreyvickery - Sorry, should've closed this before, if we need to add it to the docs we can open a separate issue for that.  Thanks!

PR #459, opened 2022-08-08
docs(stat): add rh-stat docs page

marionnegp:
## What I did

1. Added a Statistic page.

## Testing Instructions

1.

## Notes to Reviewers
- The dark theme stat styling might change, depending on #432, so those examples may have to be updated.


package.json
```diff
@@ -101,6 +101,7 @@
     "@patternfly/pfe-styles": "next",
     "@patternfly/pfe-tabs": "next",
     "@patternfly/pfe-toast": "next",
+    "async": "^3.2.4",
```
bennypowers, line 104: are the package.json changes necessary for this pr? can we remove them:

marionnegp:
@coreyvickery, I've incorporated you and Ronnie's feedback in this PR and updated the dark theme stat. I made the edits while coding, so the XD link isn't up-to-date.

coreyvickery:
@marionnegp My feedback is below. It looks like a lot, but they're all small changes that should take no time at all to make.

- [ ] Global: Use `h4` instead of `h3`
- [ ] Global: Change notation on white from `#0066cc` to `#ff00ff`
- [ ] Global: Change notation on black to from `#ffffff` to `#ff40ff`
- [ ] `Sample component` section: Left align stat image
- [ ] `Anatomy` section: Change notation color to pink (`#f0f`)
- [ ] `Anatomy` section: Remove period from image example under the body text
- [ ] `Color` subheading: Remove period from image example
- [ ] `Sizes` subheading: I think the two images can be combined into one with notation underneath that says Default and Large underneath the image, doing this will allow you to combine the separate lines of text into one paragraph
- [ ] `Usage` section: Is another word that can replace the `Layout` subheading? Same question for the `Layout` subheading in the `Style` section, it looks weird that two of the same words are so close to each other, so change one or the other
- [ ] `Layout` subheading in `Usage` section: I can barely see the grid behind the first image so make more opaque, maybe 5% opacity instead of 2% or 3%
- [ ] `Pairing statistics with footnotes` subheading: move `[1]` and `Fortune 500 . . .` much closer together
- [ ] `Vertical height` subheading: Is this image supposed to be in a container? Seems like the border is missing
- [ ] `Dark theme - secondary background` subheading: Change background color to `#212427`
- [ ] `Large screens` and `Small screens` subheadings: Make the grids a tiny bit more opaque, maybe 5% opacity instead of 2% or 3%
- [ ] Remove `<hr>` above `Best practices` heading
- [ ] `Including additional elements` subheading: Move `However, a call to action . . .` text and image somewhere else
- [ ] `Best practices` section: Add subheading above `Don't use an icon . . .`, maybe `Unrelated icon`?
- [ ] `Character count` subheading: Separate `Title: . . .`, `Number: . . .`, etc. into their own blocks of content and add a subheading for each
- [ ] Do the same as above for the `Content guidelines` blocks of content
- [ ] Remove `<hr>` above `Spacing` heading
- [ ] Adding subheading to each block of `Spacing` content
- [ ] Remove `<hr>` above links

marionnegp:
> `Including additional elements subheading`: Move `However, a call to action . . .` text and image somewhere else
Changed this to a helpful tip because I wasn't sure where else it would fit.

marionnegp:
@coreyvickery, could you take another look? The only edit I haven't made yet is changing the color of the dark theme secondary bg.

> `Sample component` section: Left align stat image

I put it in a div with a max-width, but by default one stat would be centered on the page.

> `Dark theme - secondary background` subheading: Change background color to `#212427`

The Color page lists #212427 as a tertiary background color. Should this or the color page be updated?

> `Character count` subheading: Separate `Title: . . .`, `Number: . . .`, etc. into their own blocks of content and add a subheading for each

Let me know if the changes I made here were what you were thinking.

coreyvickery (commented):
@marionnegp Everything looks great, just let me know when to approve and merge.

docs/components/statistic.njk
```diff
@@ -220,7 +220,7 @@ tags:
 {%- endcall %}
 
 <h4>Dark theme - secondary background</h4>
-<div class="example" style="background: #3C3F42; border-radius: 3px;">
+<div class="example" style="background: #212427; border-radius: 3px;">
```
bennypowers, line 223: We have access to all the tokens, so why not use the semantic token name?

marionnegp: Ha, good point. I ended up adding the tokens with their fallback values in the Accessibility section. @coreyvickery, if I pulled some inline styles from another doc page, I left it alone until the redesign. An example would be the helpful tip formatting that I found in the new nav pages.

coreyvickery: @marionnegp Approved.

coreyvickery (approved):

PR #466, opened 2022-08-10
Added analytics data attributes to demo code

### LINKED ISSUE #393: [feat] <rh-footer> discovery - analytics tracking
### Description

We need to add data attribute tracking to the rh-footer. This issue will track on the "discovery" work that Benny agreed to take on as an output of our meeting with SDI on 6/14. In essence there could be a way to modify the current structure calling for click events that's more simplified.

### Suggested solution

- setup spandx so I [Benny] can test element shadowdom with data-attrs according to Rickey's specifications
- potentially modify Wes' scripts so that components don't need to fire custom events

### Example API

_No response_

### Additional context

Resources:
Codepen:[ https://codepen.io/heyMP/pen/oNEmzbr?editors=1011](https://codepen.io/heyMP/pen/oNEmzbr?editors=1011)
Demo:[ https://github.com/heyMP/rh-footer-demo](https://github.com/heyMP/rh-footer-demo)
Docs site: https://ux.redhat.com/components/footer/
[Demo blob](https://github.com/heyMP/rh-footer-demo/blob/89f67aa18feb0e3045572cc891ed1ddd7fae41f3/index.html#L50-L59)

alypilkons-RH:
## What I did

1. JVP requested I update this file with example data-analytics attributes. I followed the structure for data analytics attributes that is being used currently in the main redhat.com nav.

## Testing Instructions

1. Visit: https://deploy-preview-466--red-hat-design-system.netlify.app/components/secondary-nav/demo/
2. Verify that the data-analytic attributes are correct.

## Notes to Reviewers

1. This will need to be looked over by the analytics team to ensure the structure meets their requirements.

Closes #393

### DIFF
--- a/elements/rh-secondary-nav/demo/rh-secondary-nav.html
+++ b/elements/rh-secondary-nav/demo/rh-secondary-nav.html
@@ -3,124 +3,113 @@
 <script type="module" src="rh-secondary-nav.js"></script>
 
 <div class="mock-nav"></div>
-<rh-secondary-nav role="navigation">
+<rh-secondary-nav role="navigation" data-analytics-region="secondary nav">
   <a href="#" slot="logo" id="logo-id">
     Red Hat Ansible Automation Platform
   </a>
   <ul slot="nav" aria-labelledby="logo-id">
     <li>
       <rh-secondary-nav-dropdown>
-        <a href="#" slot="link">Explore</a>
+        <a href="#" data-analytics-level="1" data-analytics-text="Explore" slot="link">Explore</a>
         <rh-secondary-nav-menu slot="menu">
           <rh-secondary-nav-menu-section>
             <h2 slot="header" id="ansible-platform">
-              <a href="#">Why Red Hat Ansible Automation Platforms</a>
+              <a href="#" data-analytics-level="2" data-analytics-text="Why Red Hat Ansible Automation Platforms" data-analytics-category="Explore">
+                Why Red Hat Ansible Automation Platforms
+              </a>
             </h2>
-            <ul slot="links" aria-labelledby="ansible-platform"">
+            <ul slot="links" aria-labelledby="ansible-platform">
               <li>
-                <a href="#">
+                <a href="#" data-analytics-level="3" data-analytics-text="Integrations" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                   Integrations
                   <pfe-icon icon="web-icon-new-window" color="info" size="sm"></pfe-icon>
                 </a>
               </li>
-              <li><a href="#">Custom success stories</a></li>
               <li>
-                <a href="#">
+                <a href="#" data-analytics-level="3" data-analytics-text="Custom success stories" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
+                    Custom success stories
+                </a>
+              </li>
+              <li>
+                <a href="#" data-analytics-level="3" data-analytics-text="What is ansible" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                   What is ansible
                   <pfe-icon icon="web-icon-new-window" color="info" size="sm"></pfe-icon>
                 </a>
               </li>
-              <li><a href="#">Why choose Red Hat for automation?</a></li>
-              <li><a href="#">Edge automation with Red Hat Ansible Automation Platform</a></li>
+              <li>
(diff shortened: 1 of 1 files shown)


alypilkons-RH:
tagging @heyMP

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -3,168 +3,157 @@
 <script type="module" src="rh-secondary-nav.js"></script>
 
 <div class="mock-nav"></div>
-<rh-secondary-nav role="navigation">
+<rh-secondary-nav role="navigation" data-analytics-region="secondary nav">
```
jvprh, line 6: Since this demo is using Ansible as the example, can we make this *-region="Ansible secondary nav" ?

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -3,168 +3,157 @@
...
   <a href="#" slot="logo" id="logo-id">
     Red Hat Ansible Automation Platform
   </a>
   <ul slot="nav" aria-labelledby="logo-id">
     <li>
       <rh-secondary-nav-dropdown>
-        <a href="#" slot="link">Explore</a>
+        <a href="#" data-analytics-level="1" data-analytics-text="Explore" slot="link">Explore</a>
```
jvprh, line 13: I'm not sure if *-level="" is still being used, but we can leave it for now. Rickey (SDI) and Marcia can review and verify. No action needed.

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -3,168 +3,157 @@
...
         <rh-secondary-nav-menu slot="menu">
           <rh-secondary-nav-menu-section>
             <h2 slot="header" id="ansible-platform">
-              <a href="#">Why Red Hat Ansible Automation Platforms</a>
+              <a href="#" data-analytics-level="2" data-analytics-text="Why Red Hat Ansible Automation Platforms" data-analytics-category="Explore">
```
jvprh, line 17: I think this category value, and others at this level, would also be *-category="Explore|Why Red Hat Ansible Automation Platforms" since it's within the same region, even though it's not inside of the <ul>

elements/rh-secondary-nav/demo/rh-secondary-nav.html
```diff
@@ -173,11 +162,11 @@ <h2 slot="header">
         </rh-secondary-nav-menu>
       </rh-secondary-nav-dropdown>
      </li>
-    <li><a href="#">Documentation</a></li>
-    <li><a href="#">Extra Placeholder</a></li>
+    <li><a href="#" data-analytics-level="1" data-analytics-text="Documentation">Documentation</a></li>
```
jvprh, line 165: Also need to verify with Marcia & Rickey if a *-category attr is required at this level.

elements/rh-secondary-nav/demo/rh-secondary-nav.html
alypilkons-RH: Got it. Yeah Allen had mentioned that the main nav uses them, but the footer does not, so he wasn't sure.

elements/rh-secondary-nav/demo/rh-secondary-nav.html
alypilkons-RH: Level 1 elements in the main nav don't have category attributes so followed that structure here.

elements/rh-secondary-nav/demo/rh-secondary-nav.html
alypilkons-RH: Good call!

elements/rh-secondary-nav/demo/rh-secondary-nav.html
alypilkons-RH: @jvprh Are you thinking the a tag inside the h2 tag should include its own text as part of its data-analytics category? Based on the main nav example and talking with Allen that didn't seem needed.

bennypowers:
I suggest moving this to it's own demo file, e.g. instead of changing the existing `elements/rh-secondary-nav/demo/rh-secondary-nav.html` file, I suggest creating a new `elements/rh-secondary-nav/demo/analytics.html`. There are two advantages to this approach:

1. keeping the analytics attributes in a specific place for each component makes it easy for users who need analytics attrs to find it, without cluttering other demo files
2. We could add some automated visual feedback to files called `analytics.html`, e.g. an output pane which logs syntax-highlighted JSON payloads for each relevant click

PR #469, opened 2022-08-11
[rh-alert]: Updating README.md

brianferry:
Updating the README.md file to contain information about the `rh-alert` component.

### DIFF
--- a/elements/rh-alert/README.md
+++ b/elements/rh-alert/README.md
@@ -1,11 +1,70 @@
 # Alert
-Add a description of the component here.
+An Alert is a banner used to notify a user about a change in status or communicate other information. It can be generated with or without a user triggering an action first.
 
 ## Usage
-Describe how best to use this web component along with best practices.
+An alert is used by wrapping html elements with an `rh-alert` element as slots in order to format the data in the desired manner.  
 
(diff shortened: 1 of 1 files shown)


elements/rh-alert/README.md
```diff
@@ -1,11 +1,70 @@
...
+
+If using npm/bundlers:
+```bash
+npm install @rhds/elements
+```
+
+Then once installed, import it to your application:
+
+```js
+import '@rhds/elements/rh-alert';
```
bennypowers, line 28:
```suggestion
import '@rhds/elements/rh-alert/rh-alert.js';
```

bennypowers (changes requested):
Big improvement!

bennypowers:
FYI i set this (and the stat pr) to squash with a commit type and scope `docs(stat)`. the `chore` scope is for things like dependency updates, cicd workflows, etc. README falls under `docs`, and since these commits are specific to individual elements, they should be scoped appropriately. Commit scopes should be the tag name without the `rh-` prefix.

These are perhaps pedantic distinctions, but they do help significantly in reading and understanding the commit history

bennypowers (approved):

PR #471, opened 2022-08-15
refactor(stat)!: internalize state

bennypowers:
I had an opportunity to take rh-stat for a test drive, which gave me an itch to do some refactors. 

## What I did

- BREAKING: remove `@pfelement` decorator
- BREAKING: remove `is-mobile` attribute
- BREAKING: remove `description` slot, replace with anonymous slot
- replace `MatchMediaController` with `ScreenSizeController`
- document slots and css properties
- hide empty slots
- make `updateIcons` method private, and refactor it to run on slot
  change and affect only the icon slotted into `icon`
- initialize properties in class field initializers

## Testing Instructions

1. Manually test on the deploy preview


bennypowers:
@marionnegp in this PR, I removed the "Placeholder Description" content, because I don't want end users to see strings like that on pages where developers forgot to include required content.

Could we instead issue a console warning if a stat element is added to the page without that required content? Which slots are absolutely required? stat and title?

marionnegp:
@bennypowers, a warning sounds fine! The required slots would be stat and description.

bennypowers:
Ok I added the warning, thanks @marionnegp 

@brianferry tests are failing for mobile breakpoints. Is it me?

brianferry:
@bennypowers  - Check this now, I fixed some of the tests and I think that it's a mix of there need to be more tests added to rh-stat to accommodate the mixing of light and shadow dom mixed with some tests that didn't work properly before that are being fixed with this MR now.

brianferry:
updated the importMap and the styles to fix a problem I was seeing with the icon.  Still seeing an issue where top="statistic" isn't putting the stat in the correct order above the title.

bennypowers:
Last commit styles almost everything in Shadow DOM, which should solve your comment @brianferry

docs/_data/importMap.cjs
```diff
@@ -34,6 +34,7 @@ const PFE_DEPS = [
   '@patternfly/pfe-core/functions/debounce.js',
   '@patternfly/pfe-core/functions/deprecatedCustomEvent.js',
   '@patternfly/pfe-core/functions/random.js',
+  '@lrnwebcomponents/code-sample',
```
brianferry, line 37: Adding this comment to inform that I'll be approving this PR and will resolve this issue pending @marionnegp design approval

brianferry (approved):

docs/_data/importMap.cjs
bennypowers: Thanks @brianferry. @marionnegp if  you're good, resolve this conversation to automerge. IMO, it's ok to merge this without first providing the video icon, we can open an new issue to deal with that.

bennypowers: @marionnegp in the interest of progress, I'll merge now. If you have further review, it's very much welcome, but open a new issue. Thanks.

PR #472, opened 2022-08-15
created analytics.html file

alypilkons-RH:
This is the demo HTML file with analytics data attributes added. Benny recommended putting this in its own HTML file rather than replacing the original demo file. I followed the structure for data analytics attributes that is being used currently in the main redhat.com nav.

## What I did

1. Took the original demo HTML file and added analytics data attributes to it. Created a new analytics file in the folder so that users can view the demo code with or without the analytics attributes. 

## Testing Instructions

1. Verify that the data analytic attributes are correct

## Notes to Reviewers

1. This will need to be looked over by the analytics team to ensure the structure meets their requirements.

### DIFF
--- a/elements/rh-secondary-nav/demo/analytics.html
+++ b/elements/rh-secondary-nav/demo/analytics.html
@@ -166,7 +166,7 @@
     <li><a href="#" data-analytics-text="Extra Placeholder">Extra Placeholder</a></li>
   </ul>
   <pfe-cta slot="cta">
-    <a href="#" data-analytics-text="Get started" data-analytics-LinkType="cta">Get started</a>
+    <a href="#" data-analytics-text="Get started" data-analytics-linktype="cta">Get started</a>
   </pfe-cta>
 </rh-secondary-nav>
 
--- a/elements/rh-secondary-nav/demo/analytics.html
+++ b/elements/rh-secondary-nav/demo/analytics.html
@@ -3,111 +3,111 @@
 <script type="module" src="../rh-secondary-nav.js"></script>
 
 <div class="mock-nav"></div>
-<rh-secondary-nav role="navigation" data-analytics-region="Ansible secondary nav">
+<rh-secondary-nav role="navigation" data-analytics-region="secondary-navigation-Ansible">
   <a href="#" slot="logo" id="logo-id">
     Red Hat Ansible Automation Platform
   </a>
   <ul slot="nav" aria-labelledby="logo-id">
     <li>
       <rh-secondary-nav-dropdown>
-        <a href="#" data-analytics-level="1" data-analytics-text="Explore" slot="link">Explore</a>
+        <a href="#" data-analytics-text="Explore" slot="link">Explore</a>
         <rh-secondary-nav-menu slot="menu">
           <rh-secondary-nav-menu-section>
             <h2 slot="header" id="ansible-platform">
-              <a href="#" data-analytics-level="2" data-analytics-text="Why Red Hat Ansible Automation Platforms" data-analytics-category="Explore">
+              <a href="#" data-analytics-text="Why Red Hat Ansible Automation Platforms" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                 Why Red Hat Ansible Automation Platforms
               </a>
             </h2>
             <ul slot="links" aria-labelledby="ansible-platform">
               <li>
-                <a href="#" data-analytics-level="3" data-analytics-text="Integrations" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
+                <a href="#" data-analytics-text="Integrations" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                   Integrations
                   <pfe-icon icon="web-icon-new-window" color="info" size="sm"></pfe-icon>
                 </a>
               </li>
               <li>
-                <a href="#" data-analytics-level="3" data-analytics-text="Custom success stories" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
+                <a href="#" data-analytics-text="Custom success stories" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                     Custom success stories
                 </a>
               </li>
               <li>
-                <a href="#" data-analytics-level="3" data-analytics-text="What is ansible" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
+                <a href="#" data-analytics-text="What is ansible" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                   What is ansible
                   <pfe-icon icon="web-icon-new-window" color="info" size="sm"></pfe-icon>
                 </a>
               </li>
               <li>
-                <a href="#" data-analytics-level="3" data-analytics-text="Why choose Red Hat for automation?" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
+                <a href="#" data-analytics-text="Why choose Red Hat for automation?" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                 Why choose Red Hat for automation?
                 </a>
               </li>
               <li>
-                <a href="#" data-analytics-level="3" data-analytics-text="Edge automation with Red Hat Ansible Automation Platform" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
+                <a href="#" data-analytics-text="Edge automation with Red Hat Ansible Automation Platform" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
                 Edge automation with Red Hat Ansible Automation Platform
                 </a>
               </li>
             </ul>
             <pfe-cta slot="cta">
-              <a href="#" data-analytics-linktype="cta" data-analytics-level="3" data-analytics-text="Section Level CTA" data-analytics-category="Explore|Why Red Hat Ansible Automation Platforms">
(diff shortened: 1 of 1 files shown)


bennypowers (approved):
LGTMK. The lightdom css and secondary nav script *might* need to be referenced with another `../` as in:
```html
<link rel="stylesheet" href="../demo.css">
<link rel="stylesheet" href="../../rh-secondary-nav-lightdom.css">
<script type="module" src="../rh-secondary-nav.js"></script>
```
on account of how we generate the demo pages.

see https://deploy-preview-472--red-hat-design-system.netlify.app/components/secondary-nav/demo/analytics/ and observe network panel in dev tools

You can check this in the deploy preview

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
+<link rel="stylesheet" href="demo.css">
```
bennypowers, line 1:
```suggestion
<link rel="stylesheet" href="../demo.css">
```

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
...
+<link rel="stylesheet" href="../rh-secondary-nav-lightdom.css">
```
bennypowers, line 2:
```suggestion
<link rel="stylesheet" href="../../rh-secondary-nav-lightdom.css">
```

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
...
+<script type="module" src="rh-secondary-nav.js"></script>
```
bennypowers, line 3:
```suggestion
<script type="module" src="../rh-secondary-nav.js"></script>
```

bennypowers:
I've set this PR to auto-merge once the above comments are resolved. Please check against the deploy preview linked above before resolving them. Thanks for the contribution @alypilkons-RH!

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
...
+
+<div class="mock-nav"></div>
+<rh-secondary-nav role="navigation" data-analytics-region="Ansible secondary nav">
```
bennypowers, line 6: @jvanpelt leaving this comment here to block auto-merge on your approval. Please resolve this comment (and any other that come up for your review) to enable auto merge

alypilkons-RH: Thanks @bennypowers - I committed the suggestions you added, and the preview appears correctly now.

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
...
+  <a href="#" slot="logo" id="logo-id">
+    Red Hat Ansible Automation Platform
+  </a>
+  <ul slot="nav" aria-labelledby="logo-id">
+    <li>
+      <rh-secondary-nav-dropdown>
+        <a href="#" data-analytics-level="1" data-analytics-text="Explore" slot="link">Explore</a>
```
jvprh, line 13: Are `data-analytics-level="1"` attributes still valuable for us to capture?

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
...
+        <rh-secondary-nav-menu slot="menu">
+          <rh-secondary-nav-menu-section>
+            <h2 slot="header" id="ansible-platform">
+              <a href="#" data-analytics-level="2" data-analytics-text="Why Red Hat Ansible Automation Platforms" data-analytics-category="Explore">
```
jvprh, line 17: @Rickey Messick - looking at the demo, these header links are in the same layout region as the ul links that follow. Should their `*-category=""` values also be `Explore|Why Red Hat Ansible Automation Platforms` vs just `Explore`? I would suggest they should, but this is consistent with the current RHDC main nav. (See "Services & support" for an example in production.)

elements/rh-secondary-nav/demo/analytics.html
rmessickSDI: No we should remove all those.  `data-analytics-level`

elements/rh-secondary-nav/demo/analytics.html
rmessickSDI: agree value for the category should be 'Explore|Why Red Hat Ansible Automation Platforms'

elements/rh-secondary-nav/demo/analytics.html
rmessickSDI: Can you make the region something like `data-analytics-region="secondary-navigation"`.  Do not want it to be site specific since we will use same tracking across all.

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
+<link rel="stylesheet" href="../demo.css">
+<link rel="stylesheet" href="../../rh-secondary-nav-lightdom.css">
+<script type="module" src="../rh-secondary-nav.js"></script>
+
+<div class="mock-nav"></div>
+<rh-secondary-nav role="navigation" data-analytics-region="Ansible secondary nav">
```
rmessickSDI, line 6: @jvprh Can you make the region something like data-analytics-region="secondary-navigation". Do not want it to be site specific since we will use the same tracking across all.

jvprh: @rmessickSDI - The secondary navs will not be site specific, they will be site _section_ specific. There are currently at least 6 different iterations of this nav that will be replaced by this component. The best examples are those found on the RHEL, Ansible and Openshift product pages.  Does that change the recommendation?

elements/rh-secondary-nav/demo/analytics.html
```diff
@@ -0,0 +1,209 @@
...
+            </li>
+          </ul>
+        </rh-secondary-nav-menu>
+      </rh-secondary-nav-dropdown>
+     </li>
+    <li><a href="#" data-analytics-level="1" data-analytics-text="Documentation">Documentation</a></li>
+    <li><a href="#" data-analytics-level="1" data-analytics-text="Extra Placeholder">Extra Placeholder</a></li>
+  </ul>
+  <pfe-cta slot="cta">
+    <a href="#" data-analytics-level="1" data-analytics-text="Get started">Get started</a>
```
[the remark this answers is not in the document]
jvprh: Good catch, thank you @rmessickSDI !

elements/rh-secondary-nav/demo/analytics.html
rmessickSDI: Oh, that makes sense.  Can we make it like this with the section being the last?  That way we can set one rule for all and have the option to get specific with section if we need to.  `data-analytics-region="secondary-navigation-ansible"`

jvprh: Yes, but now I have to ask about your use of dashes here vs no dashes in *-category

rmessickSDI: Sure.  The category is data values that will be put directly into analytics.  The region is just used for logic in Launch.  That logic is text string based prefer no spaces.  So for this instance just like the homepage sections, the hierarchy is the other way to make it easy for regex functions.
example:
secondary-navigation-ansible
secondary-navigation-Openshift
secondary-navigation-RHEL
I can make 1 launch condition to start with "secondary-navigation-" and have event fire across all of them.  Then also have the option to fire the specific events if business needs all the way down to section level.

Did the same with redhat.com home page sections:
homepage-offer-card-featured

jvprh: Got it. 👍

elements/rh-secondary-nav/demo/analytics.html
alypilkons-RH: Got it! I updated this one, and also the 'additional features' heading to match this logic.

elements/rh-secondary-nav/demo/analytics.html
alypilkons-RH: updated!

elements/rh-secondary-nav/demo/analytics.html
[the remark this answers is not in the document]
alypilkons-RH: Good call, this is updated!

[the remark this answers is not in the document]
bennypowers: just wanting to note that data attrs are generally transformed from dash to camel case

https://codepen.io/bennyp/pen/wvmRKWP?editors=1111

```html
<div data-analytics-LinkType="cta"></div>
```
```js
div.dataset;
/*
{
  "analyticsLinktype": "cta"
}
*/
```

so this should be equivalent to 
```html
<div data-analytics-linktype="cta"></div>
```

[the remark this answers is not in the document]
alypilkons-RH: @rmessickSDI @jvprh - see Benny's comment above - any concerns/thoughts about the camel case in LinkType?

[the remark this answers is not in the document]
rmessickSDI: We account for both so good to have it lower.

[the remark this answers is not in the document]
bennypowers: updated

elements/rh-secondary-nav/demo/analytics.html
bennypowers: Having received @rmessickSDI's approval via email, resolving this and merging

PR #477, opened 2022-08-18
refactor(alert)!: remove pfelement

bennypowers:
and refactor icons

## What I did

1. remove `@pfelement` decorator from alert
2. refactor icons in alert (eventually, we'll use a similar technique to generate icon sets for `<rh-icon>`)

## Notes to Reviewers
I dont think this _necessarily_ needs a changeset, but if you like, add one and squash.


brianferry (approved):
Looks good to me, especially like the icons variable improvement.

PR #489, opened 2022-08-23
fix(cta): design tweaks

bennypowers:
## What I did

1. updated border width
2. updated inner border offset

## Testing Instructions

1. run the demo, check the DP

## Notes to Reviewers

Brick font weight on demo:
image

Looks like the browser is applying `font-weight: 400;` to the link. @marionnegp, would you please give me some more information on the case you've identified which is not rendering properly?

> Video icon isn't showing up on demo.

This is because we don't have that icon asset yet. The closest I found was in fontawesome, but it's not an exact match. Pending redevelopment of `<pfe-icon>` and implementation of `<rh-icon>`, we'll revisit this, but I don't want to hold up `<rh-cta>` release until then


bennypowers:
looks like it's because its using `RedHatDisplay` instead of `Red Hat Display`. see https://github.com/RedHat-UX/red-hat-design-tokens/issues/17. I'd prefer to solve that there

bennypowers:
@marionnegp PTAL at the padding. The fonts, lets defer that fix until https://github.com/RedHat-UX/red-hat-design-tokens/issues/17

marionnegp:
I'd prefer for the padding value to not be dependent on font size, but I don't know if there was a reason for using rems in the tokens. If we want to change the tokens, I'm ok with the CTAs as is and dealing with the space tokens in a different issue.

bennypowers:
https://deploy-preview-489--red-hat-design-system.netlify.app/components/call-to-action/demo/, now

but the values are not as important as the token names. the values we'll automatically fix later, so if the token names are ok, just approve

bennypowers:
ok the above link is fixed now

the problem was with netlify's builders using a cached version of pfe tools. we don't need to change any code to fix it.

https://deploy-preview-489--red-hat-design-system.netlify.app/components/call-to-action/demo/

marionnegp (approved):
Token names look good!

PR #497, opened 2022-08-25
Add design tokens to rh-blockquote

Nouveau:
## What I did

1. Replaced hardcoded values with design tokens.

## Testing Instructions

1. Check that Blockquote still renders as expected.

## Notes to Reviewers

We have a few one off values currently in Blockquote that aren't in our tokens. 
- Author is body text with a weight of 500
- The left highlight border is 8px
- The left highlight border color is #43ADAF

### DIFF
--- a/.changeset/tough-apes-doubt.md
+++ b/.changeset/tough-apes-doubt.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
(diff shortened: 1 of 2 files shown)


bennypowers (approved):
Nice!

Some of the values are out of date, *but* i have a working stylelint rule that can automatically update them, so I don' think we need to fix those here.

PR #500, opened 2022-08-25
chore: autoupdate tokens

bennypowers:
## What I did

1. add [rhds tokens stylelint plugin](https://github.com/RedHat-UX/red-hat-design-tokens/pull/21)
2. update linting rules and scripts

## Testing Instructions

1. do some linting!

## Notes to Reviewers

Closes #tokens


bennypowers:
Thanks, @brianferry 

I adjusted the lint rules to align with the token values. we can always reevaluate those preferences later.

zeroedin:
> we can always reevaluate those preferences later.

I think we can definitely iron out the details on the preferences.  

Otherwise, I think this PR is great, I ran it on the whole project and targeted a specific component.  It found lots of spacing mistakes that wouldn't catch my eye otherwise.  This will definitely be handy to clean up formatting issues with the `-fix` flag, as well as more complex issues that would need more review like `no-descending-specificity`.

bennypowers:
Great. If we merge today I can put together a PR that lints the files tomorrow, and we'll go from there

zeroedin (approved):

PR #494, opened 2022-08-24
refactor(secondary-nav): remove font based scaling

zeroedin:
## What I did

  - Removed font-based component scaling with em-based lengths. Text only will now scale with the user font setting
  - Added token-based length support
  - Fixed token-based font-family stack
  - Fixed nav bar height not adjusting when logo text wraps to 3 lines
  - Fixed button background on color-palette="darker"

## Testing Instructions

1.  View on deploy preview [demo](https://deploy-preview-494--red-hat-design-system.netlify.app/components/secondary-nav/demo)

## Notes to Reviewers

  - Please look for any inconsistencies in spacing around elements.  I converted a lot of margin/borders from em -> px and applied token vars, please pay close attention to anything in the diff that might have been human error.

PR #505, opened 2022-08-30
style: apply stylelint

bennypowers:
## What I did

1. autofix all fixable linting errors in element stylesheets
2. adjust some of the linting rules in stylelintrc
3. manually fix the remaining errors (may involve visual regressions):
  1. [x] refactor `rh-alert` styles and apply tokens
  2. [x] change rule order in `rh-stat`
  4. [x] apply design tokens to `rh-cta` (should fix @marionnegp's design notes on cta)
  5. [x] apply design tokens to `rh-footer`
  6. [x] apply design tokens to `rh-tooltip`
  7. [x] apply design tokens to `rh-secondary-nav`, lint, change rule order. may close #494

## Testing Instructions

1. visual regression testing required for above-mentioned elements.

## Notes to Reviewers
In cases where I could not find a corresponding token value to the value specified in css, I added a `WARNING` comment. These may be considered for inclusion in tokens, or the comment may be updated to reflect the reasons why a non-token value was chosen.


heyMP:
@bennypowers I need to create RHDS specific CSS overrides for all of the `rem` values.

Screen Shot 2022-08-30 at 7 20 26 PM

bennypowers:
@heyMP I added some font-size and other overrides: line height, etc.

I wondering to what degree this is necessary though, shouldn't we just use the tokens as-is? does the goal have to be pixel-perfect accuracy with the previous version, even if that means using values not present in the tokens? perhaps @RedHat-UX/designers can weigh in.

Either way, here's what the updated demo looks like:

image

bennypowers:
fair warning, i force-pushed the last commit

bennypowers:
last commits add red-hat font to all demos, so keep that in mind

coreyvickery:
> I wondering to what degree this is necessary though, shouldn't we just use the tokens as-is? does the goal have to be pixel-perfect accuracy with the previous version, even if that means using values not present in the tokens? perhaps @RedHat-UX/designers can weigh in.

@bennypowers @heyMP I would advocate for using tokens as we have defined them. We should always try and get as close as possible to a given design, but if 100% pixel perfection means that we are neglecting our tokens or doing too many custom things, that is unnecessary. If we feel that tokens are missing, we can add them.

bennypowers:
ok @heyMP @coreyvickery with your permission we'll either revert that change or open a new issue for it

bennypowers:
@coreyvickery i've updated [footer](https://deploy-preview-505--red-hat-design-system.netlify.app/components/footer/demo/) to use token values
@marionnegp ditto for [cta](https://deploy-preview-505--red-hat-design-system.netlify.app/components/call-to-action/demo/). LMK if I'm using the right token names for variant padding.

Chatting with @heyMP, he clarified for me: i thought he meant he needed to maintain the non-token values in footer here in rhds, but instead, he'll solve the problem on rhdc with rem by overriding the token values with ems instead of rems. in this case, tokens actually help solve the problem.

brianferry:
@bennypowers - Looking at tooltip and alert both look good to me, alert looks slightly different with the description text using the 0.875rem value for the font size now rather than the px value of 14px but that goes with our design system so I think that's good.

brianferry:
@bennypowers rh-stat also looks good to me with the reordering of the css.

bennypowers:
Nice! Thanks @brianferry. Please leave an approval. The PR is not set to auto-merge.

brianferry (approved):
Took a look at `rh-stat`, `rh-tooltip`, and `rh-alert`, outside of the `rh-alert` font size being different everything looked good to me.  In `rh-alert` currently we're setting the description slot to be a font of `14px` whereas the new code uses the token / `0.875rem` which is the correct value from the token but we can overwrite it if we want to from a design perspective.

elements/rh-footer/rh-footer-lightdom.css
```diff
@@ -1,48 +1,51 @@
...
   display: contents;
 }
 
-:is(rh-footer, rh-global-footer) [slot^=links] a {
+:is(rh-footer, rh-global-footer) [slot^="links"] a {
   display: block;
-  color: var(--link-color, #fff) !important;
-  font-size: var(--link-font-size, 0.875em);
+  color: var(--rh-color-text-primary-on-dark, #ffffff) !important;
+  font-size: var(--rh-footer-link-font-size, var(--rh-font-size-body-text-sm, 0.875rem));
```
bennypowers, line 46: Following up on https://github.com/orgs/RedHat-UX/discussions/385#discussioncomment-3523950 here,

screenshot

@coreyvickery can you tell me which specific token name is preferred here?

The token `--rh-font-size-body-text-sm` has the value `0.875rem`, which [corresponds](https://codebeautify.org/rem-to-px-converter) to `14px` at a 16px base.

Because the font size is defined in rems, it will adjust depending on your browser font size settings or based on the page font size.

If this is in fact the right size, but is rendering contrary to expectations due to user settings or page root font size, we can adopt ems instead, but I think that's outside the scope of this PR, and we'd have to take it up in the discussion thread above.

Alternately, we can provide a non-token font size here, but your earlier comments in this thread advised against that.

bennypowers: @coreyvickery I pushed a commit just now which sets the site-wide base font size to 16px across ux-dot, including the demos.

Please check again, as I suspect this will solve your issue.

image

coreyvickery: @bennypowers The demo looks good, thank you for adjusting.

elements/rh-cta/rh-cta.css
```diff
@@ -5,16 +5,17 @@
...
   cursor: pointer;
+
   --rh-context-background-color: var(--rh-cta-background-color) !important;
 }
 
 ::slotted(*) {
   white-space: normal;
   display: inline;
-  padding: 0.6rem, 0 !important;
+  padding: var(--rh-space-md, 8px) 0 !important;
```
bennypowers, line 18: Reserving this comment for @marionnegp's review

elements/rh-secondary-nav/rh-secondary-nav.css
```diff
@@ -26,29 +30,41 @@ nav {
...
-  row-gap: 0;
   height: fit-content;
   z-index: var(--rh-secondary-nav-z-index, 102);
-  background-color: var(--rh-color-surface-light, #F0F0F0);
+  background-color: var(--rh-color-surface-light, #f0f0f0);
+
+  --_r: var(--_nav-height) max-content max-content;
+  --_c: 1fr max-content;
+
+  grid-template:
```
bennypowers, line 51: reserving this comment for @zeroedin's review

elements/rh-cta/rh-cta.css
```diff
@@ -187,33 +197,41 @@ svg {
...
+  /* --rh-color-border-interactive-on-light with 10% opacity */
+  --rh-cta-focus-container-background-color: #0066cc1a;
   --rh-cta-focus-border-color: transparent;
   --rh-cta-focus-color: var(--rh-context-light-color-focus, #0066cc);
   --rh-cta-focus-inner-border-color: transparent;
   --rh-cta-focus-text-decoration: none;
-  --rh-cta-active-container-background-color: #0066cc1a; /* --rh-color-border-interactive-on-light with 10% opacity */
+
+  /* --rh-color-border-interactive-on-light with 10% opacity */
+  --rh-cta-active-container-background-color: #0066cc1a;
```
marionnegp, line 209: Out of curiosity, what does the 1a at the end of the hex code mean?

marionnegp (commented):
CTA tokens look good to me!

elements/rh-cta/rh-cta.css
bennypowers: That's the alpha channel

https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color

bennypowers: we might consider vending hsl tokens as well, which would make this easier to validate in css.

elements/rh-secondary-nav/rh-secondary-nav.css
bennypowers: revert and allow individual grid properties

heyMP (approved):
Let's Roll 🎸

PR #507, opened 2022-09-01
fix: ship lib files to npm

bennypowers:
## What I did

1. add build artifacts from lib to the tarball


heyMP (approved):
LGTM!

Screen Shot 2022-09-01 at 8 11 19 AM

Issue #476, opened 2022-08-17
[bug] footer: implementation bugs


heyMP:
### Describe the bug

- [ ] secondary start slot not being used but still showing gap (bug in rh-footer)
- [ ] two column breakpoint on global footer not present (bug in rh-global-footer)
- [ ] tertiary slot in global footer has a margin-start padding that needs to be removed (bug in rh-global-footer)
- [ ] max width on footer-block
- [ ] icon size changed to 24px down from 28px (band aide)
- [ ] h3 line height cascading. added 1.1rem (band aide)
- [ ] side gap was reduced to 32px. side gap variable was used to overwrite.

### Which Component?

rh-footer, rh-global-footer

### Steps to reproduce

TBD

### Expected behaviour

TBD

### Operating System (OS)

macOS

### Browser

Chrome

heyMP:
Need to gather more info on each.

heyMP:
Closing this issue to resolve Abra release.

PR #509, opened 2022-09-01
fix (footer): add em based tokens example

heyMP:
## What I did

1.  Added CSS parts for accordion-header and accordion-panel
2. Added RHDC specific `em` based overrides to the proxy demo
3. Cleaned up some of the previously incorrect grid gap settings around primary links

## Note to reviewers

Some of the margins in the rh-global-footer are visibly incorrect.  These are being fixed in #476.

## Testing Instructions

1. Visit `http://localhost:8000/components/footer/demo/`
2. Verify that, on desktop, the primary links have a 
  - column gap of 32px
  - row gap of 16px
  
## Screenshots

Desktop primary links gaps

Screen Shot 2022-09-03 at 10 54 02 AM

RHDC Desktop

Screen Shot 2022-09-03 at 11 07 21 AM

RHDC Mobile

localhost_8001_en (6)

### DIFF
--- a/elements/rh-footer/RhFooter.ts
+++ b/elements/rh-footer/RhFooter.ts
@@ -158,7 +158,7 @@ export class RhFooter extends LitElement {
       <slot name="links"></slot>
       ` : html`
       <pfe-accordion on="dark" color-palette="darkest">${children.map((child, index) => staticHtml`
-        <pfe-accordion-${unsafeStatic(child.type)} part="links-accordion-${unsafeStatic(child.type)}">
+        <pfe-accordion-${unsafeStatic(child.type)} part="links-accordion-${child.type}">
           <slot name="links-${index}"></slot>
          </pfe-accordion-${unsafeStatic(child.type)}>`)}
       </pfe-accordion>
--- a/elements/rh-footer/RhFooter.ts
+++ b/elements/rh-footer/RhFooter.ts
@@ -59,7 +59,6 @@ function isHeader(tagName: string) {
  * @cssprop --rh-footer-border-color - {@default #6a6e73}
  * @cssprop --rh-footer-accent-color - {@default #e00}
  * @cssprop --rh-footer-section-side-gap - {@default 32px}
- * @cssprop --rh-footer-links-column-gap - {@default 32px}
  * @cssprop --rh-footer-links-gap - {@default 8px}
  * @cssprop --rh-footer-link-header-font-size - {@default 0.875em}
  */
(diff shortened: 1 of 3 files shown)


elements/rh-footer/rh-footer.css
```diff
@@ -206,6 +206,10 @@ pfe-accordion {
   display: contents;
 }
 
+.isMobile .links {
+  --rh-footer-link-header-font-size: var(--rh-footer-link-header-mobile-font-size, var(--rh-font-size-body-text-lg, 1.125rem));
```
bennypowers, line 210: Can you add this to the docblock?

bennypowers (changes requested):

elements/rh-footer/rh-footer.css
heyMP: Yeah, I'm still deciding on whether to add it or not. I'd prefer to solve this will a part instead of a state-specific variable.

bennypowers: agreed

heyMP: I added new parts for accordion-header and accordion-panel.  I like that much better :)

elements/rh-footer/RhFooter.ts
```diff
@@ -155,7 +158,7 @@ export class RhFooter extends LitElement {
       <slot name="links"></slot>
       ` : html`
       <pfe-accordion on="dark" color-palette="darkest">${children.map((child, index) => staticHtml`
-        <pfe-accordion-${unsafeStatic(child.type)}>
+        <pfe-accordion-${unsafeStatic(child.type)} part="links-accordion-${unsafeStatic(child.type)}">
```
bennypowers, line 161: the partname should take a static, as it's an AttributePart

```suggestion
        <pfe-accordion-${unsafeStatic(child.type)} part="links-accordion-${child.type}">
```

FWIW you could also make the partnames a list `part="links accordion ${child.type}`

then

```css
rh-footer::part(accordion header) { ... }
```

bennypowers: otherwise, LGTM

bennypowers (changes requested):

elements/rh-footer/RhFooter.ts
heyMP: I do like that approach but there isn't a good `:not()` approach.  So, `rh-footer::part(links accordion):not(header)`.  I suppose that's indicative of needing a better naming convention though.

PR #515, opened 2022-09-07
refactor(secondary-nav): token values and focus fix

zeroedin:
## What I did

1.  Replaced em-based space values with space tokens.
2. Fixed issue with focus closing menu when a click happens inside of secondary-nav fly out menu closing menu

## Testing Instructions

1.  View the netlify demo link.  Adjust browser font-size preference, only fonts should scale.
2. Click / tab around menu make sure menu opens and closes appropriately

## Notes to Reviewers

Please double-check any design spec spacing mistakes with [XD specs](https://xd.adobe.com/view/72d9148d-c898-4d7d-beaa-b9f746acdbbf-278a/specs/) for user error in copying/changing values to token usage.

### DIFF
--- a/elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
+++ b/elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
@@ -63,7 +63,7 @@ rh-secondary-nav > [slot="logo"] {
   font-size: var(--rh-font-size-body-text-md, 1rem);
   letter-spacing: var(--rh-letter-spacing-body-text, 0.01125em);
   line-height: 1.6875em; /* WARNING: not a recognized token value */
-  margin-inline-start: var(--rh-length-lg, 16px);
+  margin-inline-start: var(--rh-space-lg, 16px);
   font-family: var(--rh-font-family-heading, RedHatDisplay, "Red Hat Display", "Noto Sans Arabic", "Noto Sans Hebrew", "Noto Sans JP", "Noto Sans KR", "Noto Sans Malayalam", "Noto Sans SC", "Noto Sans TC", "Noto Sans Thai", Overpass, Helvetica, Arial, sans-serif);
   font-weight: var(--rh-font-weight-heading-medium, 500);
 }
@@ -82,9 +82,9 @@ rh-secondary-nav > [slot="nav"] {
   background-color: var(--rh-color-surface-lightest, #ffffff);
   margin: 0;
   padding: 
-    var(--rh-length-2xl, 32px)
-    var(--rh-length-lg, 16px)
-    var(--rh-length-lg, 16px);
+    var(--rh-space-2xl, 32px)
+    var(--rh-space-lg, 16px)
+    var(--rh-space-lg, 16px);
 }
 
 rh-secondary-nav > [slot="cta"] {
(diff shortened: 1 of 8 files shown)


elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
```diff
@@ -1,26 +1,27 @@
 rh-secondary-nav {
+  --_rh-secondary-nav-max-height: max-content;
```
bennypowers, line 2: 'private' vars don't need prefixes, since unlike 'public' vars, they are assigned on the host (that distinction and the 'why' should go in the wiki)
```suggestion
  --_max-height: max-content;
```
etc

elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
```diff
@@ -60,9 +61,9 @@ rh-secondary-nav > [slot="logo"] {
   max-width: var(--rh-secondary-nav-logo-max-width, 10em);
   text-decoration: none;
   font-size: var(--rh-font-size-body-text-md, 1rem);
-  letter-spacing: 0.01125em; /* WARNING: not a recognized token value */
-  line-height: 1.6875em; /* WARNING: not a recognized token value */
-  margin-inline-start: var(--rh-space-2xl, 32px);
+  letter-spacing: var(--rh-letter-spacing-body-text, 0.01125em);
```
bennypowers, line 64: is that the correct value?

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -212,12 +212,17 @@ export class RhSecondaryNav extends LitElement {
   @bound
   private _focusOutHandler(event: FocusEvent) {
     const target = event.relatedTarget as HTMLElement;
-    if (target === null || (target && !target.closest('rh-secondary-nav'))) {
-      if (this._compact) {
-        this._mobileMenuExpanded = false;
+    if (target) {
```
bennypowers, line 215: no `if { if`. use optional chaining

bennypowers (changes requested):
few small nits, otherwise looks good

elements/rh-secondary-nav/rh-secondary-nav.ts
zeroedin: Fixed in cd9062ab3f49fafe880c1f763b96f6c4ca81f52c

elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
zeroedin: The original value for letter spacing was taken from the XD file of 0.18px, so at 16px as base 0.01125em would be correct.    The CSS property `--rh-letter-spacing-body-text` however isn't the same as in the tokens project, I must have [misread it](https://github.com/RedHat-UX/red-hat-design-tokens/blob/main/tokens/font/letter-spacing.yml) when copying token/values as it is set to `0.0125rem`.  Will remove the token.

zeroedin: Fixed in 9f012b6d97cbebb79034124e6e8fd8ce1b19143e

elements/rh-secondary-nav/rh-secondary-nav-lightdom.css
zeroedin: Fixed in eb2256e4152e17de05b8cbd4bb5997cf55fbcbdd

coreyvickery:
@zeroedin

One tiny thing, the `Menu` text is `Display` instead of `Text` on small breakpoints.

Screen Shot 2022-09-07 at 2 23 09 PM

zeroedin:
> One tiny thing, the `Menu` text is `Display` instead of `Text` on small breakpoints.

Good catch.  Fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/515/commits/548ba52a32fe9765b7193ab38ea016696e367c1b

coreyvickery (approved):

bennypowers (approved):
LGTM. I renamed some of the 'private' vars to fit in with the convention elsewhere. there was a naming conflict with one of them (chevron) so i prepended `nav`

Issue #516, opened 2022-09-08
[bug] rh-alert font issue on buttons


dcaryll:
### Describe the bug

The buttons in the alert pattern are showing the wrong font-family. 

@zeroedin mentioned this is likely an issue with not using design tokens. 

### Which Component?

rh-alert

### Steps to reproduce

1. Go to https://ux.redhat.com/components/alert/demo/
2. View the buttons (Dismiss/Confirm)
3. Inspect > font-family is Arial

### Expected behaviour

The font family should follow the design spec from https://xd.adobe.com/spec/3e62e93c-8338-4f31-5040-3b81f0ed5c71-4853/screen/84de1e95-5aca-4782-8eba-5a98804f93a6/specs/

### Screenshots

![Screen Shot 2022-09-08 at 1 48 31 PM](https://user-images.githubusercontent.com/1249747/189191287-23857260-787f-4ff8-9c00-0807ef00734d.png)

### Element HTML

```HTML
<button slot="actions" data-action="dismiss">Dismiss</button>
```

### Operating System (OS)

macOS

### Browser

Chrome

zeroedin:
After inspection it does use design tokens, just doesn't explicitly set a font-family value.

brianferry:
Yup, and in addition to this we will need to set the cursor styles for anything slotted in the actions which we can also do in the css.

brianferry:
Set up a demo with the fixed fonts and cursor pointer @dcaryll https://deploy-preview-517--red-hat-design-system.netlify.app/components/alert/demo/

dcaryll:
Thank you @brianferry! @coreyvickery or @marionnegp would one of you mind doing a design review of this pattern again?

marionnegp:
@brianferry, I can take a look.

PR #517, opened 2022-09-08
fix: rh-alert not using correct font-family, cursor pointer

brianferry:
Fixes issues related to font-family and cursor pointer for rh-alert described https://github.com/RedHat-UX/red-hat-design-system/issues/516

### DIFF
--- a/elements/rh-alert/rh-alert.css
+++ b/elements/rh-alert/rh-alert.css
@@ -45,7 +45,7 @@
 }
 
 #container {
-  border-width: var(--rh-border-width-lg, 3px);
+  border-width: var(--rh-border-width-md, 2px);
   border-color: var(--_border-color);
   border-style: solid;
   background-color: var(--_background-color);
@@ -74,6 +74,7 @@ header {
 }
 
 header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
+  font-size: var(--rh-font-size-body-text-sm, 0.875rem);
   margin: 0;
 
   --rh-font-family-heading: var(--_font-family);
@@ -84,7 +85,6 @@ header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
 }
 
 #header {
-  font-size: var(--rh-font-size-body-text-sm, 0.875rem);
   color: var(--_header-color);
   font-weight: var(--rh-font-weight-heading-medium, 500);
   flex: 1 1 auto;
@@ -114,6 +114,11 @@ header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
 
 #description {
   font-size: var(--rh-font-size-body-text-sm, 0.875rem);
+}
+
+#description > ::slotted(*) {
+  margin-top: var(--rh-space-md, 8px);
+  margin-bottom: var(--rh-space-lg, 16px);
 }
 
 footer {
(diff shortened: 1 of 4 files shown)


elements/rh-alert/rh-alert.css
```diff
@@ -73,6 +75,8 @@ header {
 
 header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
   margin: 0;
+
+  --rh-font-family-heading: var(--_font-family);
```
marionnegp, line 79: The XD kit's version is showing that text in the header should be using Red Hat Text and --rh-font-size-body-text-sm. Font weight's correct.

elements/rh-alert/rh-alert.css
```diff
@@ -52,6 +53,7 @@
   display: grid;
   grid-template-columns: min-content 1fr;
   gap: var(--rh-space-xs, 4px);
+  font-family: var(--_font-family);
 }
```
marionnegp, line 57: Border width for the container should be --rh-border-width-md for all variants

elements/rh-alert/rh-alert.css
```diff
@@ -123,15 +127,21 @@ footer ::slotted([slot="actions"]) {
...
 
+footer ::slotted([slot="actions"]:hover) {
+  cursor: pointer;
+  text-decoration: underline;
+  color: var(--rh-color-interactive-blue-darkest, #004080);
+}
+
 :host(:not([variant])) #container,
 :host([toast]) #container {
```
marionnegp, line 146: I believe the token for the box shadow on the toast variant should be --rh-box-shadow-lg. That's the closest to the design specs.

marionnegp (commented):
The changes you made today look good. I found a few things from the previous iteration of rh-alert that weren't consistent with what we have in XD.

elements/rh-alert/rh-alert.css
brianferry: Updated the font size for the header, looks like the font-family was already red hat text?

elements/rh-alert/rh-alert.css
brianferry: Border width has been updated ✅

elements/rh-alert/rh-alert.css
brianferry: Updated this to --rh-box-shadow-lg w/ the token values

brianferry:
@marionnegp - Also noticed the margin between the header & the description looked off compared to the UX docs.  In the UX docs it looks like the description is 8px away from the header and 16px away from the action links so I have updated those in the demo as well to use the relevant tokens.

marionnegp:
Awesome, @brianferry! The changes look good to me.

Does it matter that the component width changed in the demo?
Screen Shot 2022-09-09 at 1 23 59 PM

brianferry:
@marionnegp - That's a good question, so in the mockups I see it's 396px width and in the component itself we set a max width of 40vh which probably isn't what we want from the demos.  Would we want to set the width to 396px regardless of the content inside so that it always appears the same?  If we set it to a max-width of 396px then we could end up with something like this when we have less content

image

marionnegp:
@brianferry, only the toast variant should have a max-width of 555px. The other two variants don't need a max-width.

brianferry:
@marionnegp - Got it, I was incorrect in thinking that all the rh-alerts had a max-width of 40vh, that was a demo specific css and not indicative of the component itself so I removed it.  I also added the max-width for the toast variant.

marionnegp:
No prob! I wasn't sure at first either, @brianferry.

marionnegp (commented):
This looks good to me, but I'll let you go ahead and merge it!

marionnegp (approved):

PR #547, opened 2022-09-27
fix(secondary-nav): nav level cta color context on dark

### LINKED ISSUE #546: [bug] <secondary-nav> mobile nav level cta color context
### Describe the bug

When the `rh-secondary-nav` has it's `color-palette` set to `darker` the nav level cta when viewed in the mobile dropdown has the incorrect color context.  

### Which Component?

rh-secondary-nav

### Steps to reproduce

1. Go to the [Secondary Nav Demo](https://ux.redhat.com/components/secondary-nav/demo/dark-variant/) for the Dark variant
2. Set the viewport < 992px for the mobile compact nav view.
3. Open the mobile menu

### Expected behaviour

The cta color context should be `on="light"` as the mobile menu dropdown is a white background

### Screenshots

Screen Shot 2022-09-27 at 11 15 51 AM

### Element HTML

_No response_

### Operating System (OS)

iOS, macOS, Linux, Windows

### Browser

### More browser information

_No response_

### Additional Info

_No response_

zeroedin:
## What I did

1. Wrapped the `<slot name="cta"><slot>` with `<rh-context-provider>` and apply correct `color-palette` attribute when in mobile view.
2. Closes #546

## Testing Instructions

1. Go to the Secondary Nav Deploy Preview for the [Dark variant](https://deploy-preview-547--red-hat-design-system.netlify.app/components/secondary-nav/demo/dark-variant/) and the [Default](https://deploy-preview-547--red-hat-design-system.netlify.app/components/secondary-nav/demo/)
3. Set the viewport < 992px for the mobile compact nav view.
4. Open the mobile menu

## Notes to Reviewers

Look for any possible combination of the mobile menu open/closed and at changing to different screen sizes to see if the color context can get out of sync.

### DIFF
--- a/elements/rh-secondary-nav/demo/demo.css
+++ b/elements/rh-secondary-nav/demo/demo.css
@@ -18,7 +18,7 @@ rh-secondary-nav-menu#custom-grid::part(sections) {
 }
 
 /* Dark demo with long logo text */
-@media screen and (min-width: 768px) {
+@media screen and (min-width: 358px) {
   rh-secondary-nav[color-palette="darker"] {
     --rh-secondary-nav-logo-max-width: 13.4em;
   }
--- a/elements/rh-secondary-nav/rh-secondary-nav.ts
+++ b/elements/rh-secondary-nav/rh-secondary-nav.ts
@@ -135,6 +135,10 @@ export class RhSecondaryNav extends LitElement {
   firstUpdated() {
     // after update the overlay should be available to attach an event listener to
     this._overlay.addEventListener('click', this._overlayClickHandler);
+    // if compact menu and dark variant then set cta color to lightest
+    if (this.colorPalette === 'darker' && this._compact) {
+      this._ctaColorPalette = 'lightest';
+    }
   }
 
   render() {
@@ -434,9 +438,6 @@ export class RhSecondaryNav extends LitElement {
       this._mobileMenuExpanded = false;
     } else {
       this._mobileMenuExpanded = true;
-      if (this.colorPalette === 'darker') {
-        this._ctaColorPalette = 'lightest';
-      }
     }
     this.dispatchEvent(new SecondaryNavOverlayChangeEvent(this._mobileMenuExpanded, this));
   }
--- a/elements/rh-secondary-nav/rh-secondary-nav.ts
+++ b/elements/rh-secondary-nav/rh-secondary-nav.ts
@@ -4,6 +4,8 @@ import { classMap } from 'lit/directives/class-map.js';
 import { bound, observed } from '@patternfly/pfe-core/decorators.js';
 import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
 
+import '../rh-context-provider/rh-context-provider.js';
+
 import './rh-secondary-nav-dropdown.js';
 import './rh-secondary-nav-menu.js';
 import './rh-secondary-nav-menu-section.js';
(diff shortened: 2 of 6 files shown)


elements/rh-secondary-nav/demo/demo.css
```diff
@@ -18,7 +18,7 @@ rh-secondary-nav-menu#custom-grid::part(sections) {
 }
 
 /* Dark demo with long logo text */
-@media screen and (min-width: 768px) {
+@media screen and (min-width: 358px) {
```
brianferry, line 21: Is this a permanent change or just one made for the demo purposes of this bug?

brianferry:
Looks good to me, I see the color context switching for the mobile view.  I see it starting in `on="dark"` for mobile if you start in mobile view but as soon as you open the menu it flips to the correct value.  👍

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -418,6 +434,9 @@ export class RhSecondaryNav extends LitElement {
       this._mobileMenuExpanded = false;
     } else {
       this._mobileMenuExpanded = true;
+      if (this.colorPalette === 'darker') {
```
brianferry, line 437: Not sure if we necessarily need this check though it would limit the number of times this value is updated potentially.

zeroedin: You are correct about this check happening too often.   The need for that check is due to initialization in mobile view not setting the value for the slotted cta.   After a change to the screen size it would reset, but not on initial load.

I corrected this in e1f98ed90a14a8c3ae2fb04f860c05d7159bc4ad by moving that check to `firstUpdated()`

elements/rh-secondary-nav/demo/demo.css
zeroedin: This is just an implementation example, the change has nothing to do with this PR specifically.  Just correcting a missed opportunity in the demo to show how a long text logo can have a custom breakpoint to break into 3 lines.

bennypowers:
:thinking: we shouldn't need the context provider, should we? cta should just pick it up

zeroedin:
@bennypowers the cta exists in the context of the nav level in desktop view.  It appropriately gains the context in desktop view for the context of the component `color-palette`. 

However, in mobile view the cta is part of the mobile dropdown.  The dropdown's background color is white.   Adding the context provider allows me to manipulate that change, otherwise for a dark variant `<rh-secondary-nav color-palette="darker">` the cta is on a white background in the dropdown menu but is still in context `on="dark"`.

elements/rh-secondary-nav/rh-secondary-nav.ts
```diff
@@ -106,6 +108,12 @@ export class RhSecondaryNav extends LitElement {
   @colorContextProvider()
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette: NavPalette = 'lighter';
 
+  /**
+   * If the host color-palette attr is set to lighter, the cta color context should be lighest
+   * Otherwise use the the same as value as the host color-palette attr as default set value.
+   */
+  @state() private _ctaColorPalette = this.colorPalette === 'lighter' ? 'lighest' : this.colorPalette;
```
bennypowers, line 115: please add a little more info here about the problem, so that we can be confident in removing the workaround down the line if we're able.
```suggestion
  // NB: yadda yadda mobile dropdown context, and that this can be removed if design adds a dark bg
  @state() private _ctaColorPalette = this.colorPalette === 'lighter' ? 'lighest' : this.colorPalette;
```

bennypowers (changes requested):

bennypowers (approved):

PR #594, opened 2022-10-19
feat(accordion): add rh-accordion

### LINKED ISSUE #303: [feat]: <rh-accordion> ready to migrate
### Tag Name

rh-accordion

### RHDS Only

- [ ] The component will be **RHDS only**, delete the PFE component once published

### Additional Info

XD link: https://xd.adobe.com/view/0cefed4d-4b3c-4e19-85b9-5bbbe09868ca-59bc/

- caret icon is lighter and points down when accordion is collapsed
- updated font sizes
- new variant: small size

CC: @coreyvickery

brianferry:
## What I did

1.  Created a new `rh-accordion` element which extends off of `BaseAccordion` in the patternfly repo.
2.  Also included with this are two child components `rh-accordion-panel` and `rh-accordion-header` which functions as the content wrapper and toggle respectively. 
3. Updated references to `pfe-accordion` in the repo to `rh-accordion`. 
4. Updated ImportMap to include latest version of `pfe-accordion@2.0.0-next.7`.
5. Closes #303

## Testing Instructions

1.  Go to https://deploy-preview-594--red-hat-design-system.netlify.app/components/accordion/demo/
2. Need to test the accessibility and functionality of the accordion components with the variants **default**, **large**, **dark mode**
3. go to https://deploy-preview-594--red-hat-design-system.netlify.app/components/accordion/demo/color-context/
4. test color context support

## Notes to Reviewers
1. The demos should match the design specs here 
- Large light mode https://xd.adobe.com/view/d20a1c17-375c-4c7f-6827-5fe2f8c2d92f-a43b/screen/d1bad3d6-75c3-455a-a26f-c135517ed131/specs/
- Large dark mode https://xd.adobe.com/view/d20a1c17-375c-4c7f-6827-5fe2f8c2d92f-a43b/screen/6f381695-556c-443c-9424-bd30f0fe3f69/specs/
- Small light mode https://xd.adobe.com/view/d20a1c17-375c-4c7f-6827-5fe2f8c2d92f-a43b/screen/6f381695-556c-443c-9424-bd30f0fe3f69/specs/
- Small dark mode https://xd.adobe.com/view/d20a1c17-375c-4c7f-6827-5fe2f8c2d92f-a43b/screen/98097ce7-3504-462a-89f5-9b3828a5f8a6/specs/


elements/rh-accordion/rh-accordion-header.css
```diff
@@ -15,6 +15,7 @@
   --pf-c-accordion__toggle--active-text--Color: #151515;
   --pf-c-accordion__toggle--expanded-text--Color: #151515;
   --pf-c-accordion__toggle--expanded-icon--Rotate: -180deg;
+  --pf-c-accordion__toggle--before--Top: -1px;
```
bennypowers, line 18: ideally, we'd  move the pf styles to pfe-accordion's scss  files  and redo all these with tokens or hardcoded values

brianferry: Agreed, this was after playing around with closing the gap between the panel and the header for the after element but this should be done in pfe by default because it'll need to happen over there as well.

bennypowers: Ok. For now, I suggest changing this PR so that it assumes there are no pf vars in baseaccordion, which will probably mean reimplementing all that values here

brianferry: Wait, are you saying we rename all the variables in BaseAccordionHeader.scss in pfe so that it doesn't have the --pf-c beginning?  Like just doing `--accordion__toggle--before--Top`?

brianferry: or do you more mean that all of these that are set not using tokens but instead using hardcoded values like `--pf-c-accordion__toggle--active-text--Color: #151515;` should be `--pf-c-accordion__toggle--active-text--Color: var(--rh-color-text-primary-on-light, #151515);

brianferry:
@bennypowers - Having trouble with pfe-accordion on the components/accordion page not being able to pull in the pfe-icons since it's looking for the ./assets/icons/fas/angle-down.js file inside of our local directory rather than from the pfe-icon directory.  Should I open an issue for this or is there one already?

marionnegp (changes requested):
@brianferry, found some small changes for accordion, but it's looking great!

- [x] This is nit-picky, but is there a way to hide the bit of box-shadow that's showing at the top of `<rh-accordion-panel>`?
Screen Shot 2022-11-01 at 10 07 12 AM
- [ ] The header text should wrap. It might also need a max-width. For the large accordion, the header text gets truncated, but the small accordion just keeps expanding to accommodate the text.
_Large_
Screen Shot 2022-11-01 at 10 26 54 AM
_Small_
Screen Shot 2022-11-01 at 10 29 24 AM
- [ ] The header text for the large interior accordions should have a font size of 18px.
- [ ] For expanded dark theme accordions, the right border color should be #6A6E73.

bennypowers:
> @bennypowers - Having trouble with pfe-accordion on the components/accordion page not being able to pull in the pfe-icons since it's looking for the ./assets/icons/fas/angle-down.js file inside of our local directory rather than from the pfe-icon directory. Should I open an issue for this or is there one already?

check the pfe icon docs, there's a solution for it with addIconSet. IIRC `/docs` should already have that worked out though. do a project-search for `PfeIcon`

brianferry:
@marionnegp 

All the notes from your previous review should be fixed now, I

- Added a cascades on the large attribute for rh-accordions, so those nested accordions should now automatically take on that property. 
- Updated the borders on the dark theme accordions
- Added a word wrap so that anything that goes beyond the current size goes to the start of the next line
- Update the box-shadow so that hopefully it won't show on the top anymore, this is the piece I think probably needs the most looking at.

Thank you again for taking a look at this!

marionnegp (changes requested):
Everything from my last review looks fixed! One last thing, our dark theme brand red changed to #FF442B, so the left border color needs to be updated. The `--rh-color-brand-red-on-dark` token's updated if you want to use it.

elements/rh-accordion/rh-accordion-header.css
```diff
@@ -0,0 +1,153 @@
...
+  --_rhds-c-accordion__toggle--BackgroundColor: #ffffff;
+  --_rhds-c-accordion__toggle--FontSize: var(--rh-font-size-body-text-md, 1rem);
+  --pfe-icon--size: 16px;
+
+  color: var(--rh-context-light-color-text, #151515);
+  background-color: var(--_rhds-background-color, var(--rh-color-surface-lightest, #ffffff));
+}
+
+:host([on="dark"]) {
+  --_rhds-c-accordion__toggle--Color: #ffffff;
```
bennypowers, line 11: these names should align with the conventions. see wiki please

bennypowers (changes requested):

marionnegp (commented):
Looks good to me!

elements/rh-accordion/demo/rh-accordion.html
```diff
@@ -0,0 +1,170 @@
...
+        <h2>Item Three</h2>
+    </rh-accordion-header>
+    <rh-accordion-panel>
+        Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et
+        dolore magna
+        aliqua.
+    </rh-accordion-panel>
+</rh-accordion>
+
+<pfe-band size="large" color-palette="darkest">
```
zeroedin, line 117: Use `rh-context-provider` instead of `pfe-band`

elements/rh-accordion/rh-accordion.ts
```diff
@@ -0,0 +1,57 @@
+import type { ColorPalette, ColorTheme } from '../../lib/context/color.js';
+import { colorContextProvider } from '../../lib/context/color.js';
+import { cascades } from '@patternfly/pfe-core/decorators/cascades.js';
+import { customElement, property } from 'lit/decorators.js';
+
+import styles from './rh-accordion.css';
+import { BaseAccordion } from '@patternfly/pfe-accordion/BaseAccordion.js';
+import { colorContextConsumer } from '@patternfly/pfe-core/decorators.js';
```
zeroedin, line 8:
```Suggestion

File: elements/rh-accordion/demo/rh-accordion.js [hunk 5]
```diff
@@ -0,0 +1,8 @@
+import '@patternfly/pfe-band';
+import '@patternfly/pfe-accordion';
```
zeroedin, line 2:
```Suggestion
import '@patternfly/pfe-accordion';
import '@rhds/elements/rh-context-provider/rh-context-provider.js';
```

elements/rh-accordion/rh-accordion-header.ts
```diff
@@ -0,0 +1,48 @@
+import type { ColorTheme } from '../../lib/context/color.js';
+
+import { customElement, property } from 'lit/decorators.js';
+
+import styles from './rh-accordion-header.css';
+import { BaseAccordionHeader } from '@patternfly/pfe-accordion/BaseAccordionHeader.js';
+
+import { colorContextConsumer } from '@patternfly/pfe-core/decorators.js';
```
zeroedin, line 8:
```Suggestion
import type { ColorTheme  } from '../../lib/context/color.js';

import { customElement, property } from 'lit/decorators.js';

import styles from './rh-accordion-header.css';
import { BaseAccordionHeader } from '@patternfly/pfe-accordion/BaseAccordionHeader.js';

import { colorContextConsumer  } from '../../lib/context/color.js';
```

brianferry: Great points @zeroedin, updated the header & demo!

bennypowers:
Hoping @brianferry will agree to block this on https://github.com/patternfly/patternfly-elements/pull/2266 :pray: 

I'm v. happy to do the work here.

elements/rh-demo/rh-demo.ts
```diff
@@ -47,16 +47,16 @@ export class RhDemo extends LitElement {
     return html`
       <div class="rh-demo-preview-area"></div>
       <div part="code" class="rh-demo-code-area">
-        <pfe-accordion>
-          <pfe-accordion-header>
+        <rh-accordion>
```
bennypowers, line 50: we need to remove the dynamic `pfe-accordion` import and replace it with a static rh-accordion import

elements/rh-footer/RhFooter.ts
```diff
@@ -158,11 +158,11 @@ export class RhFooter extends LitElement {
     return !(isMobile && children) ? html`
       <slot name="links"></slot>
       ` : html`
-      <pfe-accordion on="dark" color-palette="darkest">${children.map((child, index) => staticHtml`
```
bennypowers, line 161: we need to replace the imports

bennypowers (changes requested):

bennypowers:
Should wait for https://github.com/patternfly/patternfly-elements/pull/2282 before releasing

as well, there's a bug in context here. rh-accordion is both a provider and a consumer. if rh-accordion is `on` a dark context, but doesn't specify its own `color-palette`, we should expect it's children to inherit it's parent's dark context, but they don't.  see the demo

elements/rh-accordion/rh-accordion-header.ts
```diff
@@ -35,7 +35,9 @@ export class RhAccordionHeader extends BaseAccordionHeader {
 
   static readonly styles = [...BaseAccordionHeader.styles, styles];
 
-  @colorContextConsumer() private on: ColorTheme = 'light';
+  @colorContextConsumer()
+  @property({ reflect: true })
```
bennypowers, line 39: Why does this need to reflect?

brianferry: This brought up a good question I had yesterday that I wasn't able to figure out till this morning.  Previously I had issues getting the attributes to pass down to the child components so I added the reflects back and it worked.  It looks like though I would only need that on the `rh-accordion` element parent but the `panel` just needed a few css updates & a wrapper to have the class internally and it didn't need to reflect anymore either.

elements/rh-accordion/rh-accordion.css
```diff
@@ -0,0 +1,35 @@
...
+
+::slotted(rh-accordion-header:not(:first-child)) {
+  display: block;
+  border-bottom: 1px solid var(--_border-color);
+}
+
+::slotted(rh-accordion-header:is([expanded])) {
+  display: block;
+  border-bottom: 0;
+  box-shadow: 0 2px 4px #15151533;
```
zeroedin, line 27: Should this be using a token value?  I do notice there is a alpha difference 33% vs 20% in the token.

```Suggestion
  box-shadow: var(--rh-box-shadow-sm, 0 2px 4px rgba(21, 21, 21, 0.2));
```

elements/rh-accordion/rh-accordion.css
```diff
@@ -0,0 +1,35 @@
...
+}
+
+::slotted(rh-accordion-panel:is([expanded])) {
+  display: block;
+  border-bottom: 1px solid var(--_border-color);
+  box-shadow: 0 2px 4px #15151533;
```
zeroedin, line 33: Same as above, should this be using a token value?

```Suggestion
  box-shadow: var(--rh-box-shadow-sm, 0 2px 4px rgba(21, 21, 21, 0.2));
```

elements/rh-accordion/rh-accordion-header.css
```diff
@@ -0,0 +1,113 @@
...
+:host([large]) {
+  --_font-size: var(--rh-font-size-body-text-lg, 1.125rem);
+  --_padding-block-start: var(--rh-space-lg, 16px);
+  --_padding-inline-end: var(--rh-space-xl, 24px);
+  --_padding-block-end: var(--rh-space-lg, 16px);
+  --_padding-inline-start: var(--rh-space-xl, 24px);
+}
+
+:host([expanded]) {
+  border-right: 2px solid var(--_border-right-color, #d2d2d2);
```
zeroedin, line 39: I wonder if this should be `border-inline-end` instead of `border-right`.

When the accordion is presented under `dir="rtl"` should the borders also adjust sides?  I noticed that what looks like a brand red border when expanded isn't actually a border `--_panel-content-body-accent-color` so might be a bit more tricky to just switching to logical properties.

bennypowers (changes requested):

elements/rh-accordion/rh-accordion.css
```diff
@@ -5,11 +5,10 @@
   background-color: var(--rh-color-white, #ffffff);
 }
 
-#container { display: contents; }
-
-.dark {
-  --_border-color: var(--rh-color-border-subtle-on-light, #d2d2d2);
+:host([on="dark"]) {
```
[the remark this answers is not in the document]
brianferry: I've updated this in the PR along with the attribute false, not sure what we were seeing earlier today where the attribute false ended up fixing the styling on the child components because I cannot get that fix to work anymore.  It looks like the color context sets the dark property and then overwrites it right afterwards with a null property

elements/rh-accordion/rh-accordion-header.css
brianferry: Updated these with logical properties

elements/rh-accordion/rh-accordion.css
brianferry: Good call, went and updated this & made sure the other css variables were correct too

elements/rh-accordion/rh-accordion-panel.ts
```diff
@@ -26,7 +26,6 @@ export class RhAccordionPanel extends BaseAccordionPanel {
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
 
   @colorContextConsumer()
```
bennypowers, line 28:
```suggestion
  @colorContextConsumer({ attribute: false })
```

elements/rh-accordion/rh-accordion.ts
```diff
@@ -38,7 +38,6 @@ export class RhAccordion extends BaseAccordion {
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
 
   @colorContextConsumer()
```
bennypowers, line 40:
```suggestion
  @colorContextConsumer({ attribute: false })
```

bennypowers (changes requested):
CSS changes look really nice

elements/rh-accordion/rh-accordion.css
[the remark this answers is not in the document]
bennypowers: I'll take a look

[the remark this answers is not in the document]
bennypowers: sent some commits. needs a bit of polish on the typescript side, but runtime I *think* is correct.

bennypowers:
see also https://github.com/patternfly/patternfly-elements/pull/2308

elements/rh-accordion/rh-accordion-header.css
```diff
@@ -0,0 +1,114 @@
+:host {
+  --_padding-block-start: var(--rh-space-lg, 16px);
+  --_padding-inline-end: var(--rh-space-xl, 24px);
+  --_padding-block-end: var(--rh-space-lg, 16px);
+  --_padding-inline-start: var(--rh-space-xl, 24px);
```
bennypowers, line 5: would be good to have a RTL demo. there's an example in CTA

elements/rh-accordion/rh-accordion.css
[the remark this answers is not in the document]
bennypowers: [Context should be g2g now](https://deploy-preview-594--red-hat-design-system.netlify.app/components/accordion/demo/color-context/)

elements/rh-accordion/rh-accordion-header.css
bennypowers: I added an [RTL demo](https://deploy-preview-594--red-hat-design-system.netlify.app/components/accordion/demo/right-to-left/). it looks ok-but-not-great. We can merge this and loop back, provided this comment gets spun out to a new issue

TODO:
- adjust rotate direction
- align texts (there's maybe some flexbox thing going on)
- change border alignment

elements/rh-accordion/rh-accordion-panel.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @slot
+ *       The content of the accordion panel can be any basic markup including but not limited to div, paragraph, or nested accordion panels.
+ */
+@customElement('rh-accordion-panel')
+export class RhAccordionPanel extends BaseAccordionPanel {
+  static readonly version = '{{version}}';
+
+  static readonly styles = [...BaseAccordionPanel.styles, styles];
+
+  @colorContextConsumer() on?: ColorTheme;
```
bennypowers, line 24:
```suggestion
  @colorContextConsumer() private on?: ColorTheme;
```

bennypowers (approved):

bennypowers:
This is waiting on @marionnegp 's review

marionnegp (approved):

PR #595, opened 2022-10-19
feat(alert): dismissable alerts

brianferry:
## What I did

1. Added a click handler inside of rh-alert that will set the host element to hidden and send a composed event out on close.
2. New event is a `close` event on the alert that can be built off of using `addEventListener('close')` on the rh-alert element.


brianferry:
Closing until meeting next week

elements/rh-alert/rh-alert.ts
```diff
@@ -25,6 +28,12 @@ const ICONS = {
   }
 };
 
+export class AlertCloseEvent extends ComposedEvent {
+  constructor() {
+    super('alert-close');
```
bennypowers, line 33:
```suggestion
    super('close');
```

elements/rh-alert/rh-alert.ts
```diff
@@ -54,6 +63,11 @@ export class RhAlert extends LitElement {
 
   @property({ reflect: true, type: Boolean }) toast = false;
 
+  @bound private _closeHandler() {
```
bennypowers, line 66:
```suggestion
  #closeHandler() {
```
no need for the decorator if it's attached in the template

elements/rh-alert/rh-alert.ts
```diff
@@ -62,13 +76,19 @@ export class RhAlert extends LitElement {
         </div>
         <div id="middle-column">
           <header>
-            <div id="header"><slot name="header"></slot></div>
+            <div id="header">
+              <slot name="header"></slot>
+            </div>
             <div id="header-actions">
-              <button id="close-button" aria-label="Close" confirm>${ICONS.get('close')}</button>
+              <button id="close-button" aria-label="Close" confirm @click=${this._closeHandler}>${ICONS.get('close')}</button>
```
bennypowers, line 83:
```suggestion
              <button id="close-button" aria-label="Close" confirm @click=${this.#closeHandler}>${ICONS.get('close')}</button>
```

.changeset/dirty-bears-fetch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Adding functionality to rh-alert to allow the internal close button to set the host element to hidden and to bubble up a close event.
```
bennypowers, line 5:
```suggestion
Added behaviour to `<rh-alert>`'s close button: clicking it now removes the element.
```

Usually we don't need technical details like "this event bubbles" in the changelog, and they should be past-tense

bennypowers (changes requested):

elements/rh-alert/rh-alert.ts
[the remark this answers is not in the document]
brianferry: That could've been a piece that I had missed in our conversation, though I think that makes sense thinking on it in retrospect since we don't necessarily need it all the times for example on a toast message.

elements/rh-alert/demo/rh-alert.js
```diff
@@ -3,5 +3,10 @@ import '@rhds/elements/rh-alert/rh-alert.js';
 
 const root = document.querySelector('[data-demo="rh-alert"]')?.shadowRoot ?? document;
 
-root.querySelector('rh-alert');
+const dismissiableAlerts = root.querySelectorAll('rh-alert[dismissable]');
+
+dismissiableAlerts.forEach(alert => {
+  // eslint-disable-next-line no-console
+  alert.addEventListener('close', () => console.log('close'));
```
bennypowers, line 10: Let's leave one example in the demo where the default is prevented, to be sure that's possible, maybe run a timer and a spinner and then close to demo async workflows.

elements/rh-alert/rh-alert.ts
```diff
@@ -63,9 +63,13 @@ export class RhAlert extends LitElement {
...
   @property({ reflect: true, type: Boolean }) toast = false;
 
-  @bound private _closeHandler() {
-    this.hidden = true;
-    this.dispatchEvent(new AlertCloseEvent());
+  @property({ reflect: true, type: Boolean }) dismissable = false;
+
+  #closeHandler() {
+    const event = new AlertCloseEvent();
+    if (this.dispatchEvent(event)) {
```
bennypowers, line 70: We need unit tests for this, too

bennypowers (changes requested):
Looking great! Almost there!

elements/rh-alert/demo/rh-alert.js
brianferry: Added

elements/rh-alert/rh-alert.ts
bennypowers: should be able to just `expect(element).to.be.undefined` for that

elements/rh-alert/demo/rh-alert.html
```diff
@@ -157,4 +157,12 @@ <h3 slot="header">Toast dismissable</h3>
     <button slot="actions" data-action="dismiss">Dismiss</button>
     <button slot="actions" data-action="confirm">Confirm</button>
   </rh-alert>
+
+  <rh-alert dismissable prevent-default>
+    <h3 slot="header">Dismissiable With Prevent Default</h3>
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam eleifend elit sed est
+      egestas, a sollicitudin mauris tincidunt.</p>
+    <button slot="actions" data-action="dismiss">Dismiss</button>
```
bennypowers, line 165: I like the slotted buttons. if we have data-action, do we need `dismissable`?

elements/rh-alert/rh-alert.ts
brianferry: You would think so but I was not able to get the element to be undefined when I would run through it even though I could verify that the event was being called / not called when adding other attributes.  Figured there might have been something else I needed to do to update the component or the fixture wasn't removing it for some reason.

bennypowers: you can also check [element.isConnected](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected)

https://lit.dev/playground/#gist=b0639065ab09e87708e56a6c8d2fdc2f

brianferry: Looks like that was it, was able to remove the sinon spy

elements/rh-alert/demo/rh-alert.html
```diff
@@ -5,7 +5,7 @@
   <h2 slot="header">Alert</h2>
 
   <rh-alert state="default">
-    <h3 slot="header">Default</h3>
+   <h3 slot="header">Default</h3>
     <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam eleifend elit sed est
       egestas, a sollicitudin mauris tincidunt.</p>
     <button slot="actions" data-action="dismiss">Dismiss</button>
```
bennypowers, line 11: reserving this comment to block merge pending @brianferry 's review with design re: header actions vs footer actions. resolve to automerge

bennypowers (approved):
Great work @brianferry 

Living Gators Tread Mercifully

elements/rh-alert/demo/rh-alert.html
brianferry: Resolving this, talked to Marionne and we agreed to keep both the header actions (close button) and the footer actions.

PR #597, opened 2022-10-20
feat(tag): add <rh-tag>

### LINKED ISSUE #398: [feat]: <rh-tag> Ready to migrate
### Tag Name

rh-tag

### RHDS Only

- [ ] The component will be **RHDS only**, extends from PFE component 

### Additional Info

pfe-label demo: https://patternflyelements.org/components/label/

zeroedin:
## What I did

1. Created `rh-tag` based on `BaseLabel` from pfe-label
2. Closes #398  
 
### Design spec:
https://xd.adobe.com/view/d20a1c17-375c-4c7f-6827-5fe2f8c2d92f-a43b/screen/6a1953ee-4519-43d3-b02e-13453611c015/specs/

### Todo:

- [x] write spec tests
- [x] add changeset
- [x] rename `rh-tag`

## Testing Instructions

1. `npm start`
2. `npm run watch:docs`
3. View deploy preview


bennypowers:
i'm concerned about the use of the word `label` here.

Can we make a FACE that acts like `<label>`?

```html
<rh-label for="switch">AT reads me as label for switch</rh-label>
<rh-switch id="switch"></rh-switch>
```

Until we're sure we can do that (perhaps with cross-root aria and element internals?) I think we should consider renaming this element.

cc @markcaron @hellogreg @nikkimk

zeroedin:
> i'm concerned about the use of the word `label` here.


`rh-tag` maybe to remove the confusion with a potential FACE element?  

I'm trying to figure out why we'd need to build `rh-label` as a FACE element that mimics the standard `<label>` tag, but enforcement of styles maybe enough of a reason.

bennypowers:
> why we'd need to build `rh-label` as a FACE element that mimics the standard `<label>` tag

`label` means something specific on the web. it's critical that we preserve the semantic meaning of label as "element or attribute containing text which describes a control or other element".

zeroedin:
@coreyvickery when you have a chance could you weigh in here?  We would need to revise the naming in the XDs and future docs on `ux.` so you have the veto power here on this.  We would like to resolve this as soon as we can so can move forward with this PR as it's been sitting for a while now.

The plan is `rh-tag` (or whatever we name this component) would still extend `pfe-label`, opening up `rh-label` namespace for a future new component that would be a Form Associated Custom Element, using element internals that more resembles the `<label>` HTML tag and use case.

coreyvickery:
@zeroedin Fine with that, as long as `rh-tag` can incorporate the styles seen in the link at the top.

elements/rh-tag/README.md
```diff
@@ -1,11 +1,11 @@
-# Label
+# Tag
 Add a description of the component here.
```
bennypowers, line 2: this file needs some attention

elements/rh-tag/demo/rh-tag.html
```diff
@@ -0,0 +1,31 @@
+<link rel="stylesheet" href="./demo.css"/>
+<script type="module" src="./rh-tag.js"></script>
+
+
+  <h2 slot="header">Filled Color</h2>
```
bennypowers, line 5:
```suggestion
  <h2>Filled Color</h2>
```
etc

elements/rh-tag/demo/rh-tag.html
```diff
@@ -0,0 +1,31 @@
...
+  <rh-tag color="blue">Blue</rh-tag>
+  <rh-tag color="green">Green</rh-tag>
+  <rh-tag color="orange">Orange</rh-tag>
+  <rh-tag color="red"><span>Red <span class="visually-hidden-class">Hat</span></span></rh-tag>
+  <rh-tag color="purple">Purple</rh-tag>
+  <rh-tag color="cyan">Cyan</rh-tag>
+  <rh-tag>Grey</rh-tag>
+  <div class="example">
+      Example:
+      <pfe-codeblock><div codeblock-container code-language="html"><code>&lt;rh-tag color="blue"&gt;Blue&lt;/rh-tag&gt;</code></div></pfe-codeblock>
```
bennypowers, line 15: let's prefer `<zero-md>`

bennypowers (changes requested):

PR #599, opened 2022-10-26
feat(badge): add <rh-badge>

marionnegp:
## What I did

1. Extended 'pfe-badge' and applied styling from the final [`rh-badge` design](https://xd.adobe.com/view/b9ff51f7-08cd-4e68-b290-a64601198600-ecac/)

## Testing Instructions

1.


bennypowers:
https://www.npmjs.com/package/@patternfly/pfe-spinner

apparently spinner next-6 got published under the latest dist tag :facepalm: 

oh well, soon it'll all be /elements and there will be slightly fewer headaches

bennypowers:
a potential fix for those failing tests: https://github.com/RedHat-UX/red-hat-design-system/pull/622

elements/rh-badge/rh-badge.css
```diff
@@ -0,0 +1,38 @@
+/* stylelint-disable color-hex-length */
+
+:host {
+  border-radius: var(--pf-c-badge--BorderRadius, var(--pf-global--BorderRadius--lg, 180em));
```
brianferry, line 4: You'll want to update these to the rhds variables / default tokens https://red-hat-design-tokens.netlify.app/

brianferry: For example, for 

`min-width: var(--pf-c-badge--MinWidth, var(--pf-global--spacer--xl, 2rem));`

You would want to use 

`min-width: var(--rh-badge-MinWidth, var(--rh-space-2xl, 32px));`

brianferry (commented):
This looks great @marionnegp, awesome work!  Just a few nitpicks left in the comments, please feel free to message me or reply any questions

elements/rh-badge/rh-badge.css
bennypowers: better `--rh-badge-min-width` per [conventions](https://github.com/RedHat-UX/red-hat-design-system/wiki/CSS-Styles#tokens-and-naming-conventions)

even better don't expose min-width as a token at all, unless specifically needed, users could just set `min-width` themselves

brianferry:
Looking at the code everything looks good to me, we might want to add some spec tests just to check on the padding and the border radius but since it's not really interactive (no buttons or moving pieces) it might not be too important.  Ran it locally and checked it out at https://deploy-preview-599--red-hat-design-system.netlify.app/components/badge/demo/ and it matches the specs in both places.  Looks great @marionnegp !

elements/rh-badge/test/rh-badge.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { expect, html } from '@open-wc/testing';
+import { createFixture } from '@patternfly/pfe-tools/test/create-fixture.js';
+import { RhBadge } from '@rhds/elements/rh-badge/rh-badge.js';
+
+const element = html`
+  <rh-badge></rh-badge>
+`;
```
brianferry, line 7: Leaving this comment for if we want to go ahead and add some tests for styling into the spec test file.  If not, please resolve.

brianferry (changes requested):

brianferry (approved):

elements/rh-badge/rh-badge.ts
```diff
@@ -0,0 +1,30 @@
+import { customElement } from 'lit/decorators.js';
+
+import { BaseBadge } from '@patternfly/pfe-badge/BaseBadge.js';
+
+import styles from './rh-badge.css';
+
+export type state = (
```
bennypowers: blocked by https://github.com/patternfly/patternfly-elements/pull/2259

elements/rh-badge/rh-badge.ts
```diff
@@ -0,0 +1,28 @@
...
+export class RhBadge extends BaseBadge {
+  static readonly version = '{{version}}';
+
+  static readonly styles = [...BaseBadge.styles, styles];
+
+  /**
+   * Denotes the state-of-affairs this badge represents
+   * Options include read and unread
+   */
+  @property({ reflect: true }) state?: 'info'|'success'|'moderate'|'important'|'critical';
```
bennypowers, line 21: this will fail currently. blocked by https://github.com/patternfly/patternfly-elements/pull/2259

When two subclasses need to declare different properties with the same name, the property needs to have an abstract declaration on the baseclass with a type wide enough to encompass both cases.

bennypowers (changes requested):

elements/rh-badge/rh-badge.ts
bennypowers: I pushed some code which should fix this once we release pfe-badge and update the version here

bennypowers (approved):
Louisiana, Georgia, Tennessee, Mississippi

PR #616, opened 2022-11-03
Updated Danger alert values on Color page per DPO conversations

coreyvickery:
## What I did

1. Updated Danger alert values on Color page per DPO conversations.


bennypowers:
shouldn't this go in the tokens repo?

coreyvickery:
@bennypowers This is a quick fix for a tiny portion of the ux dot website only. Our new tokens will replace these values eventually.

marionnegp (approved):

PR #617, opened 2022-11-03
Updating breakpoints table per token conversations

coreyvickery:
## What I did

1. Updating breakpoints table per token conversations.

### DIFF
--- a/docs/foundations/grid.njk
+++ b/docs/foundations/grid.njk
@@ -132,55 +132,46 @@ tags:
         <table width="100%;">
             <thead>
                 <tr>
-                    <th>Device</th>
-                    <th>Window</th>
-                    <th>Range</th>
+                    <th>Breakpoint name</th>
+                    <th>Size range</th>
                     <th>Columns</th>
-                    <th>Gutters</th>
                 </tr>
             </thead>
             <tbody>
             <tr>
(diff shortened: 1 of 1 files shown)


bennypowers:
here too, shouldn't this be addressed in the tokens repo?

coreyvickery:
@bennypowers In this case, it is updating the page on ux dot with current design token values.

bennypowers:
shouldn't the ux page get it's values from the tokens repo? why keep two sets of tokens?

marionnegp (approved):

marionnegp:
@bennypowers @coreyvickery, we'll have to figure out how to make the breakpoint naming on [ux.redhat.com](https://ux.redhat.com/foundations/grid/#breakpoints) match what we have in the [tokens](https://red-hat-design-tokens.netlify.app/#breakpoints). Not sure which page would get the updates.

marionnegp:
@bennypowers @coreyvickery, we'll have to figure out how to make the breakpoint naming on [ux.redhat.com](https://ux.redhat.com/foundations/grid/#breakpoints) match what we have in the [tokens](https://red-hat-design-tokens.netlify.app/#breakpoints). Not sure which page would get the updates.

marionnegp (approved):

PR #622, opened 2022-11-09
fix(footer): icon dep

bennypowers:
## What I did

1. Update pfe-icon dependency
2. Update rh-footer tests, which depends on pfe-icon's apis 

## Testing Instructions

1. Try this with proxy, I think it's likely to break pages on RHDS, but OTOH, this is blocking [tests which depend on the pfe-icon implementation](https://github.com/RedHat-UX/red-hat-design-system/pull/599/checks?check_run_id=9380088070)

### DIFF
--- a/.changeset/thin-lemons-march.md
+++ b/.changeset/thin-lemons-march.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Updated `<rh-footer>` to use new `<pfe-icon>`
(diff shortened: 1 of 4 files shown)


heyMP (approved):
Tested locally.  It's been difficult to test this with the Proxy.  I'm getting timeout issues.  But RHDC will still use version pfe-icon 1.0 for now and this change is backward compatible. 👍

bennypowers:
mind nuking the package lock?

heyMP:
Sorry for the package-lock palooza. 🤦‍♂️

Issue #628, opened 2022-11-15
[bug] <rh-tooltip> demo broken


heyMP:
### Describe the bug

process.env on the CDN for popper.js

![Screenshot 2022-11-14 at 7 18 50 PM](https://user-images.githubusercontent.com/3428964/201795299-d4e584ff-8fad-4901-ad7e-2e4e7a585cc2.png)

### Which Component?

rh-footer

### Steps to reproduce

1. Go to https://ux.redhat.com/components/tooltip/demo/
2. Verify that the tooltips look like blocks and always render underneath the buttons
3. Open console to see the `process.env` error

### Expected behaviour

process.env needs to be mitigated.

### Operating System (OS)

macOS

### Browser

Chrome

brianferry:
I might have a fix for this that involves removing popper in general, need to get it tested more.  In the meantime I created a pull request with the fix at https://github.com/RedHat-UX/red-hat-design-system/pull/629

heyMP:
Sounds good, thanks for the quick fix! @brianferry

PR #629, opened 2022-11-15
fix(tooltip): demo broken - adding popperjs core back into the importmap file

brianferry:
## What I did

1. Added popperjs core back into the importmap as a static file until popperjs core can be removed. 

## Testing Instructions

1. Go to the tooltip demo in the deploy link and the demo should look correct and have correct stylings now. 

#628


heyMP (approved):
LGTM 👍

PR #633, opened 2022-11-18
New Design Tokens section

coreyvickery:
## What I did

1. Created a new section for our design tokens MVP.

## Testing Instructions

1. Go [here](https://deploy-preview-633--red-hat-design-system.netlify.app/design-tokens/) and click around the different pages.


bennypowers:
I'd like to close this pr and come back to it with one that pulls in the tokens directly and generates templates based on the values, documentation and metadata there, based on the work already done in https://red-hat-design-tokens.netlify.app/. We should not maintain two separate sources of truth for tokens, that was the whole point of establishing the tokens repo in the first place.

See https://github.com/RedHat-UX/red-hat-design-tokens/blob/main/tokens/breakpoints.yml which defines the token values and documentation

See https://github.com/RedHat-UX/red-hat-design-tokens/blob/main/lib/formats/docs.js which defines procedures for generating HTML docs from token sources

See https://github.com/RedHat-UX/red-hat-design-tokens/tree/main/docs which contains macros and templates for docs HTML

bennypowers (changes requested):
We should close this or rework it into something based entirely on the token sources

coreyvickery:
@bennypowers Maybe close it and open a new one when we can automate adding/updating values to ux dot. I think it would be best if the repo lived there. Of course, always happy to discuss.

bennypowers:
Closing to reevaluate the approach

PR #634, opened 2022-11-18
feat(tabs): add `<rh-tabs>`

zeroedin:
## What I did

1. Add `<rh-tabs>`

##Todo:

1. Box variant
2. Overflow styles

### DIFF
--- a/elements/rh-tabs/demo/rh-tabs.html
+++ b/elements/rh-tabs/demo/rh-tabs.html
@@ -1,0 +1,11 @@
+<link rel="stylesheet" href="./demo.css"/>
+<script type="module" src="./rh-tabs.js"></script>
+
+<rh-tabs>
+  <rh-tab id="users" slot="tab">Users</rh-tab>
+  <rh-tab-panel>Users</rh-tab-panel>
+  <rh-tab slot="tab">Containers</rh-tab>
+  <rh-tab-panel>Containers <a href="#">Focusable element</a></rh-tab-panel>
+  <rh-tab slot="tab">Database</rh-tab>
+  <rh-tab-panel>Database</rh-tab-panel>
+</rh-tabs
--- a/elements/rh-tabs/demo/rh-tabs.html
+++ b/elements/rh-tabs/demo/rh-tabs.html
@@ -1,11 +1,42 @@
 <link rel="stylesheet" href="./demo.css"/>
 <script type="module" src="./rh-tabs.js"></script>
 
-<rh-tabs>
-  <rh-tab id="users" slot="tab">Users</rh-tab>
-  <rh-tab-panel>Users</rh-tab-panel>
-  <rh-tab slot="tab">Containers</rh-tab>
-  <rh-tab-panel>Containers <a href="#">Focusable element</a></rh-tab-panel>
-  <rh-tab slot="tab">Database</rh-tab>
-  <rh-tab-panel>Database</rh-tab-panel>
-</rh-tabs
+<div class="input">
+  <fieldset>
+    <legend>Variant</legend>
+    <label><input type="radio" name="variant" value="default" checked> Default</label>
+    <label><input type="radio" name="variant" value="box"> Box</label>
+  </fieldset>
+
+  <fieldset>
+    <legend>Theme</legend>
+    <label><input type="radio" name="theme" value="lightest" checked> Lightest</label>
+    <label><input type="radio" name="theme" value="darkest"> Darkest</label>
+  </fieldset>
+</div>
+
+<div class="example">
+  <rh-context-provider color-palette="light">
+    <rh-tabs>
+      <rh-tab id="users" slot="tab">Users</rh-tab>
+      <rh-tab-panel>Users</rh-tab-panel>
+      <rh-tab slot="tab">Containers</rh-tab>
+      <rh-tab-panel>Containers <a href="#">Focusable element</a></rh-tab-panel>
+      <rh-tab slot="tab">Database</rh-tab>
+      <rh-tab-panel>Database</rh-tab-panel>
+    </rh-tabs>
+  </rh-context-provider>
+</div>
+
+<div class="example">
+  <rh-context-provider color-palette="light">
+    <rh-tabs vertical>
+      <rh-tab id="users" slot="tab">Users</rh-tab>
+      <rh-tab-panel>Users</rh-tab-panel>
+      <rh-tab slot="tab">Containers</rh-tab>
+      <rh-tab-panel>Containers <a href="#">Focusable element</a></rh-tab-panel>
+      <rh-tab slot="tab">Database</rh-tab>
+      <rh-tab-panel>Database</rh-tab-panel>
+    </rh-tabs>
+  </rh-context-provider>
+</div>
--- a/elements/rh-tabs/demo/rh-tabs.js
+++ b/elements/rh-tabs/demo/rh-tabs.js
@@ -1,5 +1,35 @@
 import '@rhds/elements/rh-tabs/rh-tabs.js';
+import '@rhds/elements/rh-context-provider/rh-context-provider.js';
 
-const root = document.querySelector('[data-demo="rh-tabs"]')?.shadowRoot ?? document;
+const page = document.querySelector('main');
 
-root.querySelector('rh-tabs');
+const root = page?.shadowRoot ?? document;
+
+const variantInput = root.querySelectorAll('input[name="variant"]');
+const themeInput = root.querySelectorAll('input[name="theme"]');
+
+const tabs = root.querySelectorAll('rh-tabs');
+const providers = root.querySelectorAll('rh-context-provider');
+
+function variantToggle(event) {
+  tabs.forEach(t => {
+    if (event.target.value === 'box') {
+      t.setAttribute(event.target.value, '');
+    } else {
+      t.removeAttribute('box');
+    }
+  });
+}
+variantInput.forEach(input => input.addEventListener('change', variantToggle));
+
+function themeToggle(event) {
+  providers.forEach(t => {
+    if (event.target.value === 'darkest') {
+      page.classList.add('dark');
+    } else {
+      page.classList.remove('dark');
+    }
+    t.setAttribute('color-palette', event.target.value);
+  });
+}
+themeInput.forEach(input => input.addEventListener('change', themeToggle));
--- a/elements/rh-tabs/demo/rh-tabs.js
+++ b/elements/rh-tabs/demo/rh-tabs.js
@@ -1,0 +1,5 @@
+import '@rhds/elements/rh-tabs/rh-tabs.js';
+
+const root = document.querySelector('[data-demo="rh-tabs"]')?.shadowRoot ?? document;
+
+root.querySelector('rh-tabs');
--- a/elements/rh-tabs/rh-tab.css
+++ b/elements/rh-tabs/rh-tab.css
@@ -1,0 +1,85 @@
+:host([active]),
+:host([active][on="light"]) {
+   --_link-color: var(--rh-tabs-link-color, var(--rh-color-text-primary-on-light, var(--rh-color-black-900, #151515)));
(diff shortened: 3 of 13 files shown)


bennypowers:
Blocked by color context fixes in #594

coreyvickery:
@marionnegp Please take a look too.

coreyvickery (approved):

elements/rh-tabs/rh-tab.css
```diff
@@ -0,0 +1,154 @@
...
+:host([box]) .dark {
+  --_inactive-tab-background-color: var(--rh-color-surface-dark, #3c3f42);
+}
+
+button {
+  color: var(--_button-text-color);
+  padding-block-start: var(--rh-tabs-link-padding-block-start, var(--rh-spacer-lg, 16px));
+  padding-inline-end: var(--rh-tabs-link-padding-inline-end, var(--rh-spacer-lg, 16px));
+  padding-block-end: var(--rh-tabs-link-padding-block-end, var(--rh-spacer-lg, 16px));
+  padding-inline-start: var(--rh-tabs-link-padding-inline-start, var(--rh-spacer-lg, 16px));
```
marionnegp, line 42: Default and box tabs should have 32px for inline padding. In the default variant only, the first tab should have `padding-inline-start: 0`, and the last tab should have `padding-inline-end: 0`.

marionnegp (changes requested):
- [ ] Dark theme doesn't seem to be working: 
Screenshot 2023-01-16 at 11 34 35 AM
Screenshot 2023-01-16 at 11 37 10 AM

- [x] Styling isn't loading when I view tabs in Safari: 
Screenshot 2023-01-16 at 11 31 05 AM

- [x] Borders for the first and last box tabs should look like this when tab group isn't overflowing:
Screenshot 2023-01-16 at 11 49 38 AM
Screenshot 2023-01-16 at 11 51 33 AM
I'm seeing:
Screenshot 2023-01-16 at 11 49 52 AM
Screenshot 2023-01-16 at 11 50 53 AM

- [x] Is there a way to get rid of the sliver of a tab that can be seen before the first box tab and after the last box tab in overflow?
Screenshot 2023-01-16 at 12 56 52 PM

- Where can I find specs for the different inset variants?

markcaron:
@zeroedin, a couple design review issues:

- We're missing the "disabled" fade on the left/right arrow when you've reached the ends
- The hover state should turn the arrow blue on the non-disabled arrow.
 
The hover state wasn't in the original mockup/spec, but @marionnegp is going to update them.

markcaron:
@nikkimk, Corey, Marionne and I were discussing the accessibility of some of the tabbing/focusing behaviors, and wanted to get your expertise:

- When using the arrow keys to toggle focus across the tab, they are automatically activated as you move side-to-side. Is this the expected behavior or should we be requiring the user to hit `Enter` or `Spacebar` to activate?
- When tabbing out of the tabset, and then back into them, the focus arrives on the activated tab. Wondering if the focus should always arrive on the first tab in the set?

The latter interaction is tied to the understanding that you can navigate side-to-side without activating.

Anyway, these interactions may have been purposeful, which is fine. Just new and different from PF's and previous iterations of PFE. We'll go with whatever you advise.

bennypowers:
Some advice from [WAI-ARIA practices: developing a keyboard interface](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_selection_follows_focus)

> ## Deciding When to Make Selection Automatically Follow Focus
>
> In composite widgets where only one element may be selected, such as a tablist or single-select listbox, moving the focus may also cause the focused element to become the selected element. This is called having selection follow focus. Having selection follow focus is often beneficial to users, but in some circumstances, it is extremely detrimental to accessibility.
> 
> For example, in a tablist, the selected state is used to indicate which panel is displayed. So, when selection follows focus in a tablist, moving focus from one tab to another automatically changes which panel is displayed. If the content of panels is present in the DOM, then displaying a new panel is nearly instantaneous. A keyboard user who wishes to display the fourth of six tabs can do so with 3 quick presses of the right arrow. And, a screen reader user who perceives the labels on tabs by navigating through them may efficiently read through the complete list without any latency.
> 
> However, if displaying a new panel causes a network request and possibly a page refresh, the effect of having selection automatically focus can be devastating to the experience for keyboard and screen reader users. In this case, displaying the fourth tab or reading through the list becomes a tedious and time-consuming task as the user experiences significant latency with each movement of focus. Further, if displaying a new tab refreshes the page, then the user not only has to wait for the new page to load but also return focus to the tab list.
> 
> When selection does not follow focus, the user changes which element is selected by pressing the Enter or Space key.

zeroedin:
Note: https://github.com/patternfly/patternfly-elements/issues/2323. /cc @marionnegp something I discovered when talking about some accessibility issues with @nikkimk.  Let me know if you can @coreyvickery have discussed what you want to do with the inset for RHDS.  By the updated design spec in the issue looks like maybe only 1 inset value?

zeroedin:
> Based on this it seems like we should default to active-follows-focus but provide a `manual` boolean attribute for users who generate content after the tab is activated.

@nikkimk and I discussed adding a manual activation option as well today.

nikkimk:
> * When using the arrow keys to toggle focus across the tabs, they are automatically activated as you move side-to-side. Is this the expected behavior or should we be requiring the user to hit `Enter` or `Spacebar` to activate?

Either is acceptable behavior. (See [Example of Tabs with Automatic Activation | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-automatic.html) and [Example of Tabs with Manual Activation | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html))

> * When moving focus out of of the tabs, and then back into them, the focus arrives on the activated tab. Wondering if the focus should always arrive on the first tab in the set?

Focus should go back to the active tab. (See examples from links above.)

(Also can I just say that talking about <kbd>Tab</kbd> while also talking about a tabs component, is making my head spin.)

zeroedin:
> Styling isn't loading when I view tabs in Safari:

@marionnegp so this is an issue upstream as well.  The polyfill for element internals which is needed for safari is not getting loaded.  I believe I have to switch to using the internals controller for the Base classes which then automatically loads the polyfills.

nikkimk (approved):
Lore Gimmick Trope McGuffin

elements/rh-tabs/rh-tab.ts
```diff
@@ -0,0 +1,57 @@
...
+  @observed
+  @property({ reflect: true, type: Boolean }) active = false;
+
+  @observed
+  @property({ reflect: true, type: Boolean }) disabled = false;
+
+  render() {
+    const { on = '' } = this;
+    return html`
+      <div id="rhds-container" class="${classMap({ [on]: !!on })}">${super.render()}</div>
```
nikkimk, line 48: This isn't updating on the dark background so the button text is not visible.

nikkimk (changes requested):

elements/rh-tabs/rh-tab.ts
zeroedin: Yup, unfortunately we need a fix here: https://github.com/RedHat-UX/red-hat-design-system/pull/710

elements/rh-tabs/rh-tab-panel.css
```diff
@@ -0,0 +1,21 @@
+#rhds-container {
+  display: block;
+  padding: var(--rh-space-2xl, 32px);
```
marionnegp, line 3: For vertical box tabs, the panel's padding-block value should be `--rh-space-3xl`.

elements/rh-tabs/rh-tab-panel.css
```diff
@@ -0,0 +1,21 @@
...
+}
+
+:host(:not([vertical]):not([box]):not([inset])) #rhds-container {
+  padding-inline-start: 0;
```
marionnegp, line 7: The panel tabs with this variant should be using padding: var(--rh-space-2xl, 32px) too so that the first tab's text lines up with the text in the panel.

elements/rh-tabs/demo/rh-tabs.html
```diff
@@ -0,0 +1,125 @@
...
+      <rh-tab slot="tab"> System         <pfe-icon slot="icon" icon="laptop" set="fas" size="sm"></pfe-icon></rh-tab>
+      <rh-tab-panel>System</rh-tab-panel>
+    </rh-tabs>
+  </rh-context-provider>
+</section>
+
+<section>
+  <rh-context-provider color-palette="lightest">
+    <rh-tabs vertical>
+      <rh-tab slot="tab"> Users          <pfe-icon slot="icon" icon="users" set="fas" size="sm"></pfe-icon></rh-tab>
```
bennypowers, line 113: pf-icon

elements/rh-tabs/demo/rh-tabs.js
```diff
@@ -0,0 +1,60 @@
...
+  });
+}
+
+function surfaceToggle(event) {
+  providers.forEach(surface => {
+    if (event.target.value === 'darkest') {
+      main.classList.add('dark');
+    } else {
+      main.classList.remove('dark');
+    }
```
bennypowers, line 27:
```suggestion
    main.classList.toggle('dark', event.target.value === 'darkest');
```

elements/rh-tabs/rh-tab-panel.css
```diff
@@ -0,0 +1,20 @@
+#rhds-container {
+  display: block;
+  padding: var(--rh-space-2xl, 32px);
```
bennypowers, line 3:
```suggestion
  padding: var(--rh-space-3xl, 48px);
```

elements/rh-tabs/rh-tabs.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  // cascade doesn't like undefined values as default
+  @cascades('rh-tab', 'rh-tab-panel')
+  @property({ reflect: true }) box?: 'box' | 'inset' | null = null;
+
+  @cascades('rh-tab', 'rh-tab-panel')
+  @property({ reflect: true, type: Boolean }) vertical = false;
+
+  @cascades('rh-tab')
+  @property({ attribute: 'border-bottom' }) borderBottom: 'true' | 'false' = 'true';
```
bennypowers, line 62: does this need to be `'true'|'false'` ? shouldn't it then be a boolean attr?

elements/rh-tabs/test/rh-tabs.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    it('should overflow if too wide', async function() {
+      const tabs = element.shadowRoot!.querySelector('[part="tabs"]')!;
+      const tabsOverflow = getComputedStyle(tabs).overflowX === 'auto';
+      expect(tabsOverflow).to.be.equal(true);
+    });
+
+    it('should have visible scroll buttons if overflowed', async function() {
+      await aTimeout(150);
```
bennypowers, line 156: can we remove this?

bennypowers (changes requested):
Almost there! few small nits

elements/rh-tabs/rh-tabs.ts
zeroedin: That actually needs to be removed completely.   That is PFE only, and shouldn't have been moved here.  Good catch.

elements/rh-tabs/rh-tab-panel.css
zeroedin: This is part of the conversation I'm going to have with @marionnegp later today.

elements/rh-tabs/test/rh-tabs.spec.ts
zeroedin: removed in [207d60b](https://github.com/RedHat-UX/red-hat-design-system/pull/634/commits/207d60b5aa48f6ea202a8675b5f73d93c19e0b19)

bennypowers (approved):
LGTM fire at will

marionnegp (approved):

nikkimk (approved):

Issue #625, opened 2022-11-09
[feat] Move colorContextConsumer attribute


marionnegp:
### Description

Move away from using `on` as a publicly available attribute.

Example:

```
/**
* Sets color theme based on parent context
*/
@colorContextConsumer()
@property({ reflect: true }) on?: ColorTheme;

render() {
    return html`
      <svg role="status" viewBox="0 0 100 100" aria-live="polite">
        <circle class="track" cx="50" cy="50" r="40" fill="none" vector-effect="non-scaling-stroke" />
        <circle class="dash" cx="50" cy="50" r="40" fill="none" vector-effect="non-scaling-stroke" />
      </svg>
      <slot></slot>
    `;
  }
```

### Suggested solution

Add it to a shadow container instead.

bennypowers:
@marionnegp I elaborated in OP somewhat

@brianferry @heyMP @zeroedin WDYT?

brianferry:
This makes sense to me, we'd be essentially wrapping any content we wanted to have a dark theme with a wrapper at that point which I think is pretty standard across the board now.

PR #649, opened 2022-12-04
fix(stat): display icon

bennypowers:
## What I did

1. set the icon size to `--rh-size-icon-05`
2. remove patternfly elements color context
3. add rhds color context
4. add `attribute: false` option to `@colorContextConsumer`
5. add `propertyName` protected property to `ColorContextConsumerController`

## Testing Instructions

1. run dev server or browse to dp

## Notes to Reviewers

DP https://deploy-preview-649--red-hat-design-system.netlify.app/components/statistic/demo/


marionnegp (commented):
Is the dark theme version supposed to have a white icon automatically?

elements/rh-stat/rh-stat.css
```diff
@@ -38,6 +36,10 @@ pfe-icon[size="md"],
   display: block;
   color: black;
```
brianferry, line 37: Looks like if we remove the color property for icon it will inherit the light / dark colors based on the theme.

brianferry (changes requested):

elements/rh-stat/rh-stat.css
marionnegp: This seems preferable to adding a `:host([on="dark"])` property. Would there be a case where a RH web property would have a different default icon color though?

bennypowers: The reason for the behaviour @brianferry observed was because rh-stat was using the colour context from pfe, which is slated to be deprecated and removed.

The last few commits to this branch replace the pfe context with the correct rhds context, and demonstrate how to remove the `on` attribute in favour of a private shadow class, as discussed in #625

.changeset/curly-seals-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
```
brianferry, line 3: Code looks good to me, demos should be updated as part of #594 so that they get the correct accordion styles for the view code on dark theme 👍. Leaving this comment here for design approval, feel free to resolve

brianferry (changes requested):

brianferry (approved):

marionnegp (approved):
Changed the icon color from black to #151515 in light theme. Everything else looks good to me!

Issue #656, opened 2022-12-09
[bug] rh-alert footer has margin when empty


dcaryll:
### Describe the bug

When using rh-alert, if no content is in the `footer` of the alert, it still maintains its margin. 

### Which Component?

rh-alert

### Steps to reproduce

1. Go to https://ux.redhat.com/components/alert/demo/
2. Remove content from the alert footer
3. See margin that remains

### Expected behaviour

The margin should go away if the footer is empty

### Screenshots

![Screenshot 2022-12-09 at 12 53 56 PM](https://user-images.githubusercontent.com/1249747/206767185-b23947a9-d44a-4695-84ad-9dcc5f89fbb4.png)

### Operating System (OS)

macOS

### Browser

Chrome, Safari

brianferry:
Created a PR here to fix this https://github.com/RedHat-UX/red-hat-design-system/pull/659

bennypowers:
Closed by #659

PR #659, opened 2022-12-09
fix(alert): Adding logic to check for empty actions for margin-top

brianferry:
## What I did

1.  Added a slot check for the Shadow DOM styles so the footer only gets the `margin-top` style when actions have been slotted in from the host. 

## Related Issue

#656 

## Testing Instructions

1. Go to the demo
3. Remove the actions from one of the alert elements
4. Inspect the Shadow DOM of that element, you should not see a `margin-top` on an empty footer now
5. You should still see the `margin-top` on the footers with actions slotted in.

### DIFF
--- a/.changeset/long-colts-prove.md
+++ b/.changeset/long-colts-prove.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Fixing an issue with rh-alert where the margin-top would still exist on the footer when no actions were slotted.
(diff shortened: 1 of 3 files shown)


.changeset/long-colts-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Fixing an issue with rh-alert where the margin-top would still exist on the footer when no actions were slotted.
```
bennypowers, line 5:
```suggestion
`<rh-alert>` fixed empty footer actions container still taking up blockwise space.
```

bennypowers (changes requested):
Quick changeset revision and we g2g

bennypowers (approved):
Less gratuitous template margin

PR #661, opened 2022-12-12
fix(cta): make `on` private

bennypowers:
## What I did

1. Made cta's `on` attribute private

## Testing Instructions

1. put cta on different contexts
2. make sure it looks right
3. toggle `on` attr with different values
4. make sure that has no effect

### DIFF
--- a/.changeset/shiny-bats-grab.md
+++ b/.changeset/shiny-bats-grab.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Made `on` attribute of CTA private
--- a/elements/rh-cta/rh-cta.css
+++ b/elements/rh-cta/rh-cta.css
@@ -212,7 +212,7 @@ svg {
   --rh-cta-active-text-decoration: none;
 }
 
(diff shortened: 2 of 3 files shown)


marionnegp (commented):
The private `on` attribute is working for me.

The following styling errors may need to be fixed in a different PR:

- The spacing is off for CTAs with icons. There should be an 8px space between the text and the icon.
Screen Shot 2022-12-14 at 1 49 02 PM

- The hover states for brick CTAs and secondary CTAs are off in dark theme.
Screen Shot 2022-12-14 at 1 49 43 PM
Screen Shot 2022-12-14 at 1 49 39 PM
Background color for secondary CTA on dark theme should be #fff. I'm trying to find out what the background color for brick CTA on dark theme should be #3C3F42.

- Padding for the left and right of primary, secondary, and brick CTA should be 32px. Right now, they're 24px.

bennypowers:
@marionnegp I spun out a new issue for those in #673. Would you mind dropping an approval here to merge this PR?

marionnegp (approved):
`On` attribute is working as expected.

PR #666, opened 2022-12-13
Fixing margin and font-weight issues

markcaron:
## What I did

Fixed the padding/margin to match the [design spec](https://xd.adobe.com/spec/3e62e93c-8338-4f31-5040-3b81f0ed5c71-4853/screen/f5c8e80c-83cb-4ae9-8e92-5991c39e738b/specs/).

Also updated the font-weight, since it seems `regular` is mapped to `300`.


brianferry (approved):
Lucifer Grants The Margins (LGTM)

PR #675, opened 2022-12-15
fix(context): makes `on` default to shadow-private

bennypowers:
## What I did

1. made `@colorContextConsumer` default to shadow-private
1. updated changesets

## Testing Instructions

1. check stat, alert, spinner in DP

### DIFF
--- a/elements/rh-context-provider/rh-context-provider.css
+++ b/elements/rh-context-provider/rh-context-provider.css
@@ -1,5 +1,5 @@
-:host { display: block; }
-div {
+:host {
+  display: block;
   background-color: var(--rh-context-background-color);
   color: var(--rh-context-text);
 }
--- a/.changeset/blue-carrots-march.md
+++ b/.changeset/blue-carrots-march.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tooltip>`: calculated the tooltip offset when the element is first updated
--- a/.changeset/cool-keys-unite.md
+++ b/.changeset/cool-keys-unite.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-blockquote>`: added colour context and updated style tokens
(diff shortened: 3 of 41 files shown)


zeroedin (changes requested):
Added a PR to this PR that addresses some issues I came across.

bennypowers:
@zeroedin I encorporated the demo you wrote in #678 here. I didn't add the other `color-palette` themes because that's a bigger design question (and because fixes made here to blockquote seem to solve the problem)

marionnegp (approved):
Color context changes are working, and the component styling looks correct to me.

elements/rh-context-provider/rh-context-provider.css
```diff
@@ -1,5 +1,5 @@
-:host { display: block; }
-div {
+:host {
```
zeroedin, line 1: I still find myself questioning if this component should provide the background color.

Are we creating a component that would be better named`rh-surface` or `rh-band` by doing this?  Should this component be a physical (for lack of better terms) component that modifies the visual styles or is it only providing context to child components?

```html
/** Mobile Screen size  (Simplified example) **/
<rh-secondary-nav color-palette="dark">
  /* Shadow DOM */
  <nav class="compact">
    <div id="container" class="expanded">
      ...
      <div id="cta">
        <rh-context-provider color-palette="lightest">
          <slot name="cta"></slot>
        </rh-context-provider>
      </div>
    </div>
  </nav>
</rh-secondary-nav>
```

```css
#container.expanded #cta {
  background-color: var(--rh-color-surface-lightest, #ffffff);
}
```

The above example shows how `rh-secondary-nav` implements the background color when a mobile menu is expanded.   The CSS provides the surface color while the `rh-context-provider` only provides the context to the child consumer components.  The mobile menu toggle event modifies the context provider with the switch of contexts.    

I guess my question is about which component should own those styles?  Should this component remain transparent and only provide context?  If not should we rename it?

zeroedin (approved):
Llamas Graze The Mountain

elements/rh-context-provider/rh-context-provider.css
bennypowers: I'm ok with removing the background colors. @zeroedin can you make that change? Make sure to check all the demo files

Issue #657, opened 2022-12-09
[bug]  rh-context-provider change causes visual style issue on rh-secondary-nav


zeroedin:
### Describe the bug

In commit b7aa79aa1aea83d6d3b9234c20d492d402c17aca a `<div>` element was added to `rh-context-provider`. Doing so has side effects on `rh-secondary-nav` as it expects no wrapper to be inside of `rh-context-provider`.   When something like a `rh-cta` is added and its container is styled as a flexbox and vertically centered it will no longer do so correctly with this surrounding `div`

If this container is needed in other areas could we add a `part` attribute to it to make it styleable?

```js
  render() {
    return html`<div part="container"><slot></slot></div>`;
  }
```

### Which Component?

rh-context-provider

### Steps to reproduce

1. use `rh-context-provider`
2. apply element inside its slot 

### Expected behaviour

rh-context-provider should not adjust layout of contents

### Screenshots

![image](https://user-images.githubusercontent.com/13666/206788263-c262ec71-1626-4356-93ac-1788e4dda3be.png)

### Operating System (OS)

iOS, macOS, Linux, Windows

### Browser

Chrome, Safari, Safari iOS, Microsoft Edge, Firefox

zeroedin:
I looked at this again and approached the fix differently than when I first looked.  No change needed to `rh-context-provider`

bennypowers:
i'd be happy to remove the wrapping div, but i found that without it, rcp didn't actually apply background or colour styles

zeroedin:
> i'd be happy to remove the wrapping div, but i found that without it, rcp didn't actually apply background or colour styles

Understood, and no worries, I have a solution to the div being present atm, but looking into a separate sorta related issue now but will ping you on the PR when it's ready for review.

bennypowers:
ok, let's make sure it works with all the demo pages because that's where i encountered it. we ought to write some unit tests, or maybe a visual regression

PR #677, opened 2022-12-15
fix(secondary-nav): update rh-context-provider styles

zeroedin:
## What I did

Changes to `<rh-secondary-nav>`:
 - Updated styles for `rh-context-provider` in shadowroot to ensure cta centering.
 - **BREAKING**: Fixed incorrect color map for dark variant `color-palette="darker"` to `color-palette="dark"`
 
Changes to `colorContextController`:
 - Added `light` and `dark` context-color css rules to match surface tokens `--rh-color-surface-light` and `--rh-color-surface-dark`

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/657

## Testing Instructions

1. View deploy preview
2. Ensure navbar slotted CTA is vertically centered 
3. Ensure cta on lighter variant nav bark is on `lighter` surface when in navbar and `lightest` when in mobile menu dropdown. 
4. Ensure cta on dark variant nav bark is on `dark` surface when in navbar and `lightest` when in mobile menu dropdown.


bennypowers:
i'm fine with this fix, but see https://github.com/RedHat-UX/red-hat-design-system/issues/657#issuecomment-1353388886

zeroedin:
> Do you want to do the remove-background-styles part in this PR or another one?

I'm not sure that I need to now yet.  I still have an issue with the color outstanding with color context not supporting `dark | light` I need to fix , I'm going to try and make that on this PR if possible.  If it gets involved with removing background-styles then we'll close this one and I'll open a new one for that.

zeroedin:
@bennypowers after talking with @coreyvickery we decided to add the `light` and `dark` to the `ColorContextController`.  I've done so in the last set of commits to this PR.  

This allows the `background-color` to stay as is on `rh-context-provider` and I've modified its use in secondary-nav to adjust.  Maybe we should revisit this in a discussion still: Should `rh-context-provider` now as it gives an actual surface be renamed to match its semantics?

lib/context/color.ts
```diff
@@ -19,8 +19,10 @@ export type ColorPalette = (
   | 'base'
   | 'accent'
   | 'complement'
+  | 'light'
   | 'lighter'
   | 'lightest'
+  | 'dark'
```
bennypowers, line 25: this change requires design sign-off

can't we just `[color-palette^="dark"]` in the css? https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#attrvalue_4

bennypowers (changes requested):

lib/context/color.ts
zeroedin: I'm ok with switching to the `^` selector.

@coreyvickery is aware these changes were coming, and has been added as a design review here now as well, forgot to do that.

zeroedin: > Rather than adding new color palettes

So you are saying don't add `light` and `dark` to the ColorPalette type, but DO add the attribute?  Can you explain that a bit more when you have the chance?

coreyvickery (approved):

coreyvickery (approved):

lib/context/color.ts
bennypowers: `ColorPalette` is a user-facing API, and it represents all the available palettes an element can have. So before, we said that a card could have base, accent, complement, lighter, lightest, darker, or darkest color palettes. That already struck me as too much, but now adding dark and light seems excessive to me. The point of having this type in our project isn't to permit more tests to pass, it's to communicate to developers using RHDS what the high level color theming concept contains. We should be trying to limit this type as much as possible, rather than expanding it to suit our needs as developers.

However, if design is certain that they want users to choose between 4 lightish colour palettes and 5 darkish color palettes when working with our components, well then i suppose this is ok

coreyvickery: @bennypowers @zeroedin I don't think we need `base`, `accent`, and `complement`.

I do think we need `lightest`, `lighter`, `light`, `darkest`, `darker`, and `dark` though. I can see other design teams using a similar structure, PatternFly might already, but the names might be different. Happy to confirm the color values for each of these as well.

bennypowers: If @coreyvickery 's good with it, then I'm good with it :)

bennypowers (approved):
Less than.. greater... man?
Solid pickup, @zeroedin !

.changeset/metal-chairs-pay.md
```diff
@@ -0,0 +1,9 @@
+---
+"@rhds/elements": patch
+---
+
+Changes to `<rh-secondary-nav>`:
+ - Updated styles for `rh-context-provider` in shadowroot to ensure cta centering.
+ - **BREAKING**: Fixed incorrect color map for dark variant `color-palette="darker"` to `color-palette="dark"`
+Changes to `colorContextController`:
```
bennypowers, line 8:
```suggestion
Changes to **Color Context**:
```

PR #678, opened 2022-12-15
fix(context): adds missing colors and demo

zeroedin:
## What I did

1. Added `dark` and `light` as `ColorPalette` context colors
2. Added demos for `rh-context-provider` with components that should be context-aware

## Todo

- [ ] Address `rh-blockquote` not picking up the correct context.  See ~~deploy preview~~ localhost demo. (Should we fix this in this PR or open an issue and fix with a followup?)

## Dev Notes:

`rh-secondary-nav` is by default `color-palette="lighter"`.  However, it also has a dark variant `color-palette="dark"`. 
    
This was incorrectly mapped as `darker` in prior releases, and was exposed by this change to `rh-context-provider`. So when `rh-context-provider` started providing actual background colors using a container div in it's shadow root this caused the slotted cta in `rh-secondary-nav` to display the wrong (although correct by name) color background due to it being wrapped by `rh-context-provider`. 

`rh-secondary-nav` needs this `rh-context-provider` to handle the switch between secondary-nav's cta context from desktop (lighter|dark) to mobile (lightest) when the mobile menu is opened.  So, in order to fix this, we need to provide the `dark` surface token value and add it to the `ColorPalette` type which is what this PR does.

However, this brings up the question should `rh-context-provider` provide the background surface?  My original thought was it was nearly a wrapper that provides the context (and the ability to switch it) and that the components themselves would apply the correct surface colors. `rh-context-provider` was a way to inform child elements which are context aware of that change that may only exist in CSS.  At some point, I think having `rh-context-provider` provide the background color is almost like recreating `pfe-band`, but without the additional layout content.  That said, I'm not sure I have a strong opinion either way, but maybe this should be a discussion thread?

Secondary Nav Desktop Example:
```
 rh-secondary-nav color-palette="dark"
 |
 |_ rh-context-provider color-palette="dark"
    |
    |_rh-cta component gets on `dark`
```

Secondary Nav on Mobile Nav Open Example:
```
 rh-secondary-nav color-palette="dark" /* mobile nav button clicked and dropdown nav opens */
 |
 |_ rh-context-provider color-palette="lightest" /* changed in JS as CSS changes but parent context stays the same */
    |
    |_rh-cta component gets on `light`
```

### DIFF
--- a/elements/rh-context-provider/demo/demo.css
+++ b/elements/rh-context-provider/demo/demo.css
@@ -1,0 +1,3 @@
+.example {
+    padding: 1rem;
+}
--- a/elements/rh-context-provider/demo/rh-context-provider.html
+++ b/elements/rh-context-provider/demo/rh-context-provider.html
@@ -1,0 +1,166 @@
+<link rel="stylesheet" href="demo.css">
+<script type="module" src="rh-context-provider.js"></script>
+
+<rh-context-provider color-palette="lightest">
+
+  <div class="example">
+    <h2>Lightest</h2>
+    <p>Here are some components that should be aware of their context:</p>
+
+    <rh-cta>Hello World</rh-cta>
+    <rh-spinner size="sm"></rh-spinner>
+    <rh-blockquote align="center">
+      <p>In open source, we feel strongly that to really do something well, you have to get a lot of people involved.</p>
+      <span slot="author">Linus Torvalds</span>
+      <span slot="title">Software Engineer</span>
+    </rh-blockquote>
+    <rh-stat>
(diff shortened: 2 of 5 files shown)


brianferry:
> My original thought was it was nearly a wrapper that provides the context (and the ability to switch it) and that the components themselves would apply the correct surface colors. 

I would agree with this, I would think that `rh-context-provider` will / can provide the data for the components but I would think the components themselves need to apply the correct styles.  

IMO if a component needed to style a child component differently than the rest of the elements I would use `rh-context-provider` and then style it either through the parent if it's a one off or through the child component if it doesn't already have those styles for that context.

bennypowers:
I'm very happy to revert the added `div` in context-provider, so long as all the demos work out.

PR #686, opened 2022-12-21
fix(context): request host for update

zeroedin:
## What I did

1. Added request to host for an update after  properties changed in `contextConsumer`

### DIFF
--- a/.changeset/five-jars-smell.md
+++ b/.changeset/five-jars-smell.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Added request to host for an update after property changed in contextConsumer
--- a/.changeset/five-jars-smell.md
+++ b/.changeset/five-jars-smell.md
@@ -2,4 +2,4 @@
 "@rhds/elements": patch
 ---
 
-Added request to host for an update after property changed in contextConsumer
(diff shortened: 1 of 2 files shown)


lib/context/color.ts
```diff
@@ -281,6 +281,7 @@ export class ColorContextConsumer extends ColorContextController implements Reac
         } else {
           this.host.setAttribute(this.attribute, next);
         }
+        this.host.requestUpdate();
```
bennypowers, line 284: Can you move this to lines 80 and 73?

bennypowers (changes requested):

lib/context/color.ts
zeroedin: Ok, gotcha as we don't need to request the update on an attribute change ✅

bennypowers (approved):
Good catch thanks for fixing this

.changeset/five-jars-smell.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Added request to host for an update after property changed in contextConsumer
```
bennypowers, line 5:
```suggestion
Fixed context features, making sure they update the host element
```

PR #674, opened 2022-12-14
fix(docs): remove spinner demo page from ux.redhat.com

marionnegp:
## What I did

1. Deleted Spinner page from the Component section.

## Testing Instructions

1. Check that Spinner doesn't show up in the deploy preview link. The Component nav should look like this:
Screen Shot 2022-12-14 at 5 31 49 PM

PR #689, opened 2022-12-22
test(tag): use tokens in tests

bennypowers:
## What I did

1. fix the rh-tag tests, using token names as specified


bennypowers:
unblocks https://github.com/RedHat-UX/red-hat-design-system/pull/674 for @marionnegp

heyMP (approved):
Dang, those tokens are looking nice! 👍

PR #690, opened 2023-01-03
doc: adding event listener example to rh-alert

brianferry:
We had discussed sometime back about adding documentation to the `rh-alert` README outlining how you could add your own custom event handler for `rh-alert` on a `close` event.  This is adding that code as well as a reasoning behind why you would use this.


elements/rh-alert/README.md
```diff
@@ -78,3 +78,15 @@ import '@rhds/elements/rh-alert/rh-alert.js';
...
 ```
+
+### Adding an Event Listener to a Basic Alert
+
+If you would like to add custom logic when the alert is closed, you can do so with JavaScript. 
+```js
+  // Query for the alert element
+  const alert = document.querySelector('rh-alert');
+  alert.addEventListener('close', () => {
+    // Add code to be executed when the alert element is closed.
```
bennypowers, line 89: This example should demonstrate using event.preventDefault() to keep the alert from closing

https://github.com/RedHat-UX/red-hat-design-system/blob/221a585ba610620953f940f2e72e66ba6314d006/elements/rh-alert/rh-alert.ts#L99

An example could be running some async function and only closing the alert when it resolves

bennypowers (changes requested):

elements/rh-alert/README.md
```diff
@@ -78,3 +78,15 @@ import '@rhds/elements/rh-alert/rh-alert.js';
...
+  });
```
bennypowers, line 90:
```suggestion
  alert.addEventListener('close', async function(event) {
    // Prevent the alert from closing
    event.preventDefault();
    // Perform some async task
    await alertClosed();
    // Close the alert when finished
    this.remove();
  });
```
I'm not sure why the alert's private `#closeHandler` sets the state to `'error'`, that might interfere with certain workflows. Perhaps we should remove that line?

elements/rh-alert/README.md
brianferry: Perhaps this could be a second example in addition to the basic one of adding code after the close event is fired?  I was thinking having a basic one would be helpful for those not accustomed yet to adding event listeners to elements.

elements/rh-alert/README.md
brianferry: Added this in as another example

elements/rh-alert/README.md
```diff
@@ -90,3 +90,16 @@ If you would like to add custom logic when the alert is closed, you can do so wi
...
+```js
+  // Query for the alert element
+  const alert = document.querySelector('rh-alert');
+  alert.addEventListener('close', () => {
+    // Prevent the alert from closing
+    event.preventDefault();
+    // Perform some async task
+    await alertClosed();
+    // Close the alert when finished
+    this.remove();
```
bennypowers, line 103: Needs to be a `function` for the `this` binding to point to the alert

brianferry: Great point, went ahead and updated.

bennypowers (approved):

PR #693, opened 2023-01-06
feat(tooltip): rh-tooltip to latest pfe-tooltip (floating-ui)

brianferry:
Updated the @patternfly/pfe-tooltip version to the latest from our updates, moved the styles over and updated the token values as well as the base component to include the Placement and content variable.

### DIFF
--- a/elements/rh-tooltip/demo/rh-tooltip.html
+++ b/elements/rh-tooltip/demo/rh-tooltip.html
@@ -6,7 +6,7 @@
 
   <p>
     <rh-tooltip position="left">
-      <pfe-button><button>Left Tooltip</button></pfe-button>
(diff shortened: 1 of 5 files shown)


zeroedin (approved):
Looks Great, Tooltips Merged

marionnegp (commented):
- [x] Add box-shadow (`--rh-box-shadow-sm`)
- [ ] Left and right tooltip text should be left-aligned

Everything else looks good!

zeroedin:
> * [ ]  Add box-shadow (`--rh-box-shadow-sm`)
> * [ ]  Left and right tooltip text should be left-aligned
> 
> Everything else looks good!

You'll want to consider RTL languages:  `text-align: start`

PR #710, opened 2023-01-19
fix(context): nested context

bennypowers:
## What I did

1. Added tests for nested contexts with undefined intermediary values

## Testing Instructions
- [Accordion demo](https://deploy-preview-710--red-hat-design-system.netlify.app/components/accordion/demo/)
- [Accordion context demo](https://deploy-preview-710--red-hat-design-system.netlify.app/components/accordion/demo/color-context/)
- [Provider demo](https://deploy-preview-710--red-hat-design-system.netlify.app/components/context-provider/demo/)
- [Provider nested demo](https://deploy-preview-710--red-hat-design-system.netlify.app/components/context-provider/demo/nested-contexts/) (do `import('/assets/rhds.min.js')` in console until #725 fixes the demo import paths)
- [Blockquote demo](https://deploy-preview-710--red-hat-design-system.netlify.app/components/blockquote/demo/)


bennypowers:
cc @zeroedin

zeroedin:
I was having trouble navigating around in the context given its organization in a single file.  I broke it down to logical parts to help me stay focused on the individual parts.

lib/context/provider.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /** Return the current CSS `--context` value, or null */
+  get #contextVariable(): ColorTheme | null {
+    return this.#style.getPropertyValue('--context').trim() as ColorTheme || null;
+  }
+
+  constructor(host: T, options?: ColorContextProviderOptions<T>) {
+    const { attribute = 'color-palette', ...rest } = options ?? {};
+    super(host, rest);
+    this.logger = new Logger(host);
```
bennypowers, line 41: maybe adding a private-to-the-provider ColorContextConsumer here would do the trick

bennypowers:
@zeroedin PTAL

bennypowers:
@zeroedin is this ready to go?

zeroedin:
> @zeroedin is this ready to go?

I'm still seeing the "Always Lightest" panel get switched to a dark context in the accordion demo when color-palette is changed to "dark" in the grand parent.
image

https://deploy-preview-710--red-hat-design-system.netlify.app/components/accordion/demo/color-context/

bennypowers:
We'll need a unit test to account for that

zeroedin:
> We'll need a unit test to account for that

We have that test but in terms of `rh-context-provider` and the `test-context-consumer` this test passes.  This is where we left off on our last pairing I believe.

```js
    beforeEach(async function() {
      await fixture(html`
        <rh-context-provider id="gp" color-palette="darkest">
          <rh-context-provider id="p" color-palette="lightest">
            <test-context-consumer id="c"></test-context-consumer>
          </rh-context-provider>
        </rh-context-provider>
      `);
      grandparent = document.getElementById('gp') as RhContextProvider;
      parent = document.getElementById('p') as RhContextProvider;
      child = document.getElementById('c') as ContextConsumer;
      await Promise.all([grandparent, parent, child].map(x => x.updateComplete));
    });
...
    describe('updating the grandparent context', function() {
      beforeEach(async function() {
        grandparent.colorPalette = 'darker';
        await Promise.all([grandparent, parent, child].map(x => x.updateComplete));
      });
      it('does not update the child context', function() {
        // Parent color context should override the grandparent context
        expect(child.on).to.equal('light');
      });
```

We can write one using a nested accordion, but then we run into testing `on` as it is private.   Worth testing for the shadow root class at that point?  Thoughts?

bennypowers:
OK, I'm hoping this'll do it:
- tested nested accordion case in context demo,
- added unit tests for above to context spec (combination provider/consumer)
- moved context files around
  - this structure should balance @zeroedin 's desire to separate files out without adding too many request chains
  - moved relevant types to their related files, which simplifies imports using `type` named import binding modifier
- fixed a context bug in blockquote

Along the way @zeroedin I noticed that secondary nav's dropdown always proactively sets a context. confirming that this is intentional. If so we should add a `// comment` explaining why so that no one makes that field optional.

zeroedin:
> Along the way @zeroedin I noticed that secondary nav's dropdown always proactively sets a context. confirming that this is intentional. If so we should add a `// comment` explaining why so that no one makes that field optional.

Yes, that is intentional, secondary-nav manages a rh-context-provider in the shadowroot to facilitate the cta slot being on the navbar context (lighter|darker) to the "lightest" dropdown menu when on mobile, this switch happens at a breakpoint in which the menu shifts to mobile styles. 

I'll add a comment

zeroedin:
This looks fixed to me.  Pulled tabs over in this branch locally to test, I'll need to make some adjustments but that was expected. As far as I can tell these changes do the job and do it well.

🍾 🎉  Nice work @bennypowers

zeroedin (approved):

PR #711, opened 2023-01-19
fix(accordion): adding right-to-left (rtl) styles

brianferry:
## What I did

1.  Added logical properties to .toggle:after for the red border on the accordion elements
2. Added --_isRTL to handle whether or not the rotation of the icon needs to happen clockwise (applicable for right-to-left text) or counter-clockwise (applicable for left-to-right text)

## Testing Instructions

- [ ] Go to the deploy link right-to-left demo
- [ ] Check that the following styles are correct
  - [ ] Text orientation
  - [ ] Red border position
  - [ ] Icon rotation on expand

### DIFF
--- a/.changeset/two-wolves-push.md
+++ b/.changeset/two-wolves-push.md
@@ -2,6 +2,4 @@
 "@rhds/elements": patch
 ---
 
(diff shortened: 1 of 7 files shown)


bennypowers (approved):
להיות גדול, תמשיך ממש

PR #730, opened 2023-02-07
fix(global-footer): correct logo href

heyMP:
## What I did

1. The fedora logo in `rh-global-footer` should always point to the Red Hat homepage no matter what site it's on.  I updated the logo url in the shadowroot to no longer be relative but absolute instead.
2. Added a test to check the logo slot for the absolute link to RHDC.

## Testing Instructions

1. Go to https://deploy-preview-730--red-hat-design-system.netlify.app/components/footer/demo/
2. Click on the gray Fedora in the universal footer (rh-global-footer)
3. Verify that you were redirected to the Red Hat .com homepage (https://redhat.com)

### DIFF
--- a/.changeset/fair-schools-deliver.md
+++ b/.changeset/fair-schools-deliver.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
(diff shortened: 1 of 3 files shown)


brianferry (approved):
Looks good, ran the tests & they passed

PR #735, opened 2023-02-15
chore: use node 18, wireit caching in ci

bennypowers:
## What I did

1. Upgrade CI node version to 18
2. Use the wireit cache action in CI
3. Enable linting to fail a CI run
4. lint element css, ts, docs cjs

## Testing Instructions

1. FAAFO


heyMP:
I don't know if this is related to this change but I noticed that the there are stale CSS files when using `npm run start`

Replicate:
1. npm run clean:nuke
2. npm ci
3. npm run build
4. npm run start
5. Go to http://localhost:8000/components/footer/demo/
6. Now try to edit css in the rh-footer.css file: `/elements/rh-footer/rh-footer.css`

bennypowers:
> I don't know if this is related to this change but I noticed that the there are stale CSS files when using `npm run start`
> 
> Replicate:
> 
>     1. npm run clean:nuke
> 
>     2. npm ci
> 
>     3. npm run build
> 
>     4. npm run start
> 
>     5. Go to http://localhost:8000/components/footer/demo/
> 
>     6. Now try to edit css in the rh-footer.css file: `/elements/rh-footer/rh-footer.css`

if the problem is that it doesn't seem to update the page, i think that's unrelated. does it clear up after a manual refresh, or do you have to kill the server each time?

brianferry:
> > I don't know if this is related to this change but I noticed that the there are stale CSS files when using `npm run start`
> > Replicate:
> > ```
> > 1. npm run clean:nuke
> > 
> > 2. npm ci
> > 
> > 3. npm run build
> > 
> > 4. npm run start
> > 
> > 5. Go to http://localhost:8000/components/footer/demo/
> > 
> > 6. Now try to edit css in the rh-footer.css file: `/elements/rh-footer/rh-footer.css`
> > ```
> 
> if the problem is that it doesn't seem to update the page, i think that's unrelated. does it clear up after a manual refresh, or do you have to kill the server each time?

The changes made to `css` outside of the `demo/` folder don't seem to pick up until you run `npm run start` again, however this does not appear to be a problem specific with this branch as it's also happening on `main`.

brianferry:
Code looks good, was able to run and do a lint check and no errors though lots of warnings still.  Will need to go through and check the components and make sure they all look good style wise still but this is looking good!

heyMP:
> > I don't know if this is related to this change but I noticed that the there are stale CSS files when using `npm run start`
> > Replicate:
> > ```
> > 1. npm run clean:nuke
> > 
> > 2. npm ci
> > 
> > 3. npm run build
> > 
> > 4. npm run start
> > 
> > 5. Go to http://localhost:8000/components/footer/demo/
> > 
> > 6. Now try to edit css in the rh-footer.css file: `/elements/rh-footer/rh-footer.css`
> > ```
> 
> if the problem is that it doesn't seem to update the page, i think that's unrelated. does it clear up after a manual refresh, or do you have to kill the server each time?

You have to restart the server.

heyMP:
Confirmed this is an issue on `main`.  Shouldn't be a blocker for this PR.

elements/rh-button/rh-button.css
```diff
@@ -499,3 +484,24 @@ button[disabled] {
...
   --_border-color: transparent;
 }
+
+/******************************
+ *                            *
+ *     SECONDARY-DANGER       *
+ *                            *
+ ******************************/
+
+:host([variant="secondary" i][danger]) button {
```
brianferry, line 494: What does the `i` do here in the css?

brianferry: or is it just a typo?

bennypowers: makes it case insensitive
https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#attrvalue_4

brianferry (approved):

PR #736, opened 2023-02-16
fix: remove husky from postinstall for users

bennypowers:
## What I did

1. remove husky step from postinstall when it's just users doing `npm i @rhds/elements`


heyMP (approved):
👍

PR #738, opened 2023-02-16
fix: bundle in package phase

bennypowers:
## What I did

1. ensure bundle is prepared when publishing

## Testing Instructions

1.


heyMP (approved):
LGTM

PR #756, opened 2023-03-05
docs: remove abspath hack in import map

bennypowers:
## What I did

1. investigated the problem in the import map related to abspaths and found it had something to do with `@lrnwebcomponents/code-sample`
2. alias that package specifically, and remove the previous hack

## Testing Instructions

1. check demo pages, including nested demos 
    - cta demo https://deploy-preview-756--red-hat-design-system.netlify.app/components/call-to-action/demo/
    - cta nested demo https://deploy-preview-756--red-hat-design-system.netlify.app/components/call-to-action/demo/analytics/

### DIFF
--- a/CONTRIBUTING_DEV.md
+++ b/CONTRIBUTING_DEV.md
@@ -54,3 +54,16 @@ npm start
 ```
 
 This starts a local dev server at http://localhost:8000 and the 11ty dev server for the docs site at http://localhost:8080
+
+To run *only* the components dev server, first run the build, then run the dev 
+server:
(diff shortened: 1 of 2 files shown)


brianferry:
Do we need this if we're removing `rh-demo` in https://github.com/RedHat-UX/red-hat-design-system/pull/766 or should we only remove the `@lrnwebcomponents/code-sample` specific code?

bennypowers:
would be good to update this PR after #766 lands, I'd still like to remove the hack if possible

PR #757, opened 2023-03-05
fix(subnav): this reference

bennypowers:
## What I did

1. fix a bad this reference in RhSubnav's static initializer block

## Testing Instructions

1. load up any page with rhds.min.js on it (including homepage)
2. There should be no console errors sourcemapped to `rh-subnav.ts`


zeroedin (approved):
Latency Gateway TCP/IP Megabytes

PR #766, opened 2023-03-09
fix: remove rh-demo

zeroedin:
## What I did

1. Removed `rh-demo`

## Testing Instructions

1. View [Deploy Preview]() ensure rh-demo is no longer used on `rh-stat`


bennypowers:
@brianferry maybe we should fold in #756 to this PR, like you suggested

bennypowers (approved):

PR #784, opened 2023-03-19
chore: don't error when bundling empty dirs

bennypowers:
## What I did

1. prevent `scripts/bundle.js` from erroring when it encounters empty directories. This will fix the compressed size action on PRs with new files.

## Testing Instructions

1. add an empty dir `elements/rh-jazz-hands` then `npm run build` - bundle script should not error.
1. merge this then update the audio-player PR. compressed size action should not fail


marionnegp (approved):
Running 'npm run build' worked for me!

PR #798, opened 2023-03-20
fix: revert glob v9 update

bennypowers:
## What I did

1. revert the upgrade to glob v9, which silently broke the 11ty build

## Testing Instructions

1. `npm start`


brianferry (approved):
Lemon Grape Tea Mix

zeroedin (approved):

PR #812, opened 2023-03-21
fix(accordion): remove overflow from headers

brianferry:
## What I did

1. Removing overflow on `span` element from `rh-accordion-header` as it's causing visual error on mobile screens for `rh-footer`.


heyMP (approved):
LGTM 👍

Issue #813, opened 2023-03-21
[feat] <axe-lightdom-tester>


nikkimk:
### Description

While we are able to use `axe-core` to accessibility test components, we can't account for what content authors are actually putting into light DOM. 

We should create a controller that uses `axe-core` and can be run on when content is in authoring mode or with analytics reporting, to check actual light DOM content for accessibility and log issues.

### Acceptance Criteria

- [ ] Design done
- [ ] Development done

### Other resources

[Axe Javascript Accessibility API](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#section-2-api-reference for configuration)

bennypowers:
We'd like to use something like @paceaux's selector hound to do static analysis instead.

PR #814, opened 2023-03-21
fix(footer): accessible logo in universal

nikkimk:
## What I did

1. Fixed `<rh-global-footer>` slotted default logo so that link uses `aria-label` (instead of incorrectly using `alt` attribute) and its SVG uses a `<title>` element (instead of incorrectly using `title` attribute).
2. Added a controller that checks slotted links and buttons to ensure light DOM has readable link text and button labels. 

## Testing Instructions

1. Use Axe Dev Tools to check `<rh-footer>` and `<rh-global-footer>` (@hellogreg only).
2. Change the footer and global footer demo HTML locally to include links that have no text, and verify that the `lightdomAccessibleTextController` places logger warnings in the console.

## Notes to Reviewers

We may want to use this else where or consider something ab bit more comprehensive like https://github.com/RedHat-UX/red-hat-design-system/issues/813


.changeset/big-shoes-kick.md
```diff
@@ -0,0 +1,8 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-global-footer>`: 
+- Changed slotted default logo so that link uses `aria-label` (instead of incorrectly using `alt` attribute) and its SVG uses a `<title>` element (instead of incorrectly using `title` attribute).
+- Added `lightdomAccessibleTextController` to check link text and button labels for accessibility
```
bennypowers, line 7:
```suggestion
`<rh-footer-universal>`: improved accessibility for screen reader users, and others
```

elements/rh-footer/lightdomAccessibleTextController.ts
bennypowers, line 1: Based on the call we had yesterday, let's defer adding this controller, and possible adopt the plan in #813 instead

bennypowers (changes requested):

bennypowers (approved):

elements/rh-footer/lightdomAccessibleTextController.ts
bennypowers: moved to https://github.com/RedHat-UX/red-hat-design-system/pull/816

Issue #795, opened 2023-03-20
[bug] <footer> Heading structure: primary links


nikkimk:
### Which Component 
rh-footer

### Steps to reproduce

See the [Initial RHDC Homepage Accessibility Evaluation - Heading structure](https://docs.google.com/document/d/1s6rWBTYdyI6mmWjZxDKgkLzHpKKiO7PMbYh3Ev5xbrg/edit#heading=h.jytbsv3ci72e)

PR #815, opened 2023-03-21
fix(footer): resolves heading structure #795

nikkimk:
## What I did

1. In `<rh-footer>`, added a visually hidden `h2` with a slot for footer heading text (default is "Red Hat footer").
2. In `<rh-universal-footer>` only if it is not already inside a `footer`, added a visually hidden `h2` with a slot for footer heading text (default is "Red Hat footer") and wrapped in a `footer`.
3. Added visually hidden styles to `rh-footer.css` for footer `h2` to visually hide it.

## Testing Instructions

1. From the `rh-footer` demo, use a screen reader use heading navigation to get to the "Red Hat footer" heading level 2. All subsequent headings should be h3 within the footer or its universal footer. (@hellogreg)
2. From the `rh-footer` demo, verify that an `h2` with slotted text "Red Hat footer" appears within the `rh-footer` shadowDOM's `footer` tag.
3. From the `rh-footer` demo, verify that no `footer` or `h2` appear within `rh-universal-footer`'s shadowDOM.
4. From the `rh-universal-footer` demo, use a screen reader use heading navigation to get to the "Red Hat footer" heading level 2. All subsequent headings should be h3 within the universal footer. (@hellogreg)
5. From the `rh-universal-footer` demo, verify that an `h2` with slotted text "Red Hat footer" appears within the `rh-universal-footer` shadowDOM's `footer` tag.

## Notes to Reviewers

- Footers content should be inside a `footer` element.
- An offscreen `h2` makes it easier for screen reader users who navigate by heading to skip to the footer content.
- Since `rh-universal-footer` can be used inside other footers, I added logic to only add `footer` and `h2` if it wasn't already nested inside another `footer`.


nikkimk:
@bennypowers it shouldn't be.  It just adds a visually hidden heading to shadow DOM that users wouldn't already have as a form of wayfinding for screenreader users.

bennypowers (approved):

bennypowers:
Looks like we got some aXe errors

nikkimk:
> Looks like we got some aXe errors

The tests had two footers on the page at the same time, which rightly failed the a11y audit since the footers now correctly include `<footer>`. I updated the test so that only one footer is added at the same time.

heyMP:
I'll resolve the conflict between this PR and my footer refactor PR.

PR #816, opened 2023-03-22
feat(footer): a11y audit controller

bennypowers:
## What I did

1. move footer html a11y audit controller from #814 to here
2. refactor

## Notes to Reviewers
This may be closed in favour of #813

### DIFF
--- a/elements/rh-footer/RhFooter.ts
+++ b/elements/rh-footer/RhFooter.ts
@@ -77,6 +77,9 @@ export class RhFooter extends LitElement {
     super.connectedCallback();
     // wire up accessibility aria-labels with unordered lists
     this.updateAccessibility();
+    window.requestIdleCallback?.(() =>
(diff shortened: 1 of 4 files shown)


elements/rh-footer/rh-footer-html-audit-controller.ts
nikkimk, line 1: Question from @zeroedin: Is this something we'll want to use on subnav and other nav components with links?

elements/rh-footer/RhFooter.ts
bennypowers, line 1: let's wait on merging this

markcaron:
@nikkimk and @bennypowers do either of you mind revisiting this? Wondering where we're at on it.

zeroedin:
@bennypowers @markcaron Should we close this PR?  Seems like the idea is no longer something we are pursuing or warranted at this time.

bennypowers:
We'll address this with selector hound

PR #835, opened 2023-03-23
fix(cta): use flex gap and remove inline padding

### LINKED ISSUE #833: [bug] <rh-cta>
### Describe the bug

Call to action when an SVG is used such as the default arrow has a `padding-inline`, however, when a `pf-icon` is used there is no padding.  See the play button vs arrow icon below:

### Which Component?

rh-cta

### Steps to reproduce

View the [cta demo](https://ux.redhat.com/components/call-to-action/demo/)

### Expected behaviour

Any icon should have the same spacing in releation to the cta text

### Screenshots

image

### Element HTML

_No response_

### Operating System (OS)

iOS, macOS, Linux, Windows

### Browser

_No response_

### More browser information

_No response_

### Additional Info

_No response_

zeroedin:
## What I did

1. Removed inline padding from shadowdom svg and pf-icon, added a flex gap replacement.

Closes #833

## Testing Instructions

1. View [deploy preview demo for cta](https://deploy-preview-835--red-hat-design-system.netlify.app/components/call-to-action/demo/), `<svg>` or `<pf-icons>` should be spaced the same using `var(--rh-space-md, 8px)`


elements/rh-cta/rh-cta.css
```diff
@@ -42,14 +42,13 @@
   border-color: var(--rh-cta-border-color, transparent);
   border-radius: var(--rh-border-radius-default, 3px);
   border-width: var(--rh-border-width-sm, 1px);
+  gap: var(--rh-space-md, 8px);
 
   --_context-background-color: var(--rh-cta-background-color) !important;
   --_arrow-size: 13px;
 }
 
-#container pfe-icon {
```
zeroedin, line 50: The actual bug was here, a usage of `pfe-` vs `pf-`.  I assumed this was a good opportunity to move to use `gap` on the flex container instead.

bennypowers (changes requested):

bennypowers (approved):
Labour generates true magnificence

PR #836, opened 2023-03-23
fix(footer): a11y test failures

nikkimk:
## What I did

1. Added time for the linked lightDOM CSS to load so that color contrast tests could pass.
2. Removed `it.skip('universal is accessible', function() { return expect(universalFooter).to.be.accessible(); });` on kitchen sink test because universal is part of the kitchen sink fixture represented by `element`. (`universalFooter` variable is not a fixture for this test.)

### DIFF
--- a/elements/rh-footer/test/rh-footer.spec.ts
+++ b/elements/rh-footer/test/rh-footer.spec.ts
@@ -142,7 +142,7 @@ const UNIVERSAL_FOOTER_TEMPLATE = html`
       <a href="#">*We’ve updated our privacy statement effective December 30, 202X.</a>
     </div>
   </rh-footer-universal>
-  <link rel="stylesheet" href="/elements/rh-footer/rh-footer-lightdom.css">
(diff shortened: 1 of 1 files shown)


elements/rh-footer/test/rh-footer.spec.ts
```diff
@@ -142,7 +142,7 @@ const UNIVERSAL_FOOTER_TEMPLATE = html`
       <a href="#">*We’ve updated our privacy statement effective December 30, 202X.</a>
     </div>
   </rh-footer-universal>
-  <link rel="stylesheet" href="/elements/rh-footer/rh-footer-lightdom.css">
+  <link rel="stylesheet" href="/elements/rh-footer/rh-footer-lightdom.css" />
```
bennypowers, line 145: please revert. [trailing slash is not necessary in HTML void elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element#self-closing_tags)

bennypowers (changes requested):

bennypowers (approved):

PR #843, opened 2023-03-24
fix(alert): prevent page styles from overriding component styles

### LINKED ISSUE #842: [bug]: alert heading is not inline with icon
Alert heading is not inline with icon

see https://deploy-preview-821--red-hat-design-system.netlify.app/elements/call-to-action/#navigation-vs.-action

_Originally posted by @nikkimk in https://github.com/RedHat-UX/red-hat-design-system/pull/821#discussion_r1146416838_

bennypowers:
## What I did

1. sprinkle `!important` on to slotted styles

Closes #842

### DIFF
--- a/elements/rh-alert/rh-alert.css
+++ b/elements/rh-alert/rh-alert.css
@@ -76,10 +76,10 @@ header {
 }
 
 header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
-  font-size: var(--rh-font-size-body-text-sm, 0.875rem);
-  margin: 0;
+  font-size: var(--rh-font-size-body-text-sm, 0.875rem) !important;
+  margin: 0 !important;
 
-  --rh-font-family-heading: var(--_font-family);
+  --rh-font-family-heading: var(--_font-family) !important;
 }
 
 #header-actions {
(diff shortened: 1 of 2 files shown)


elements/rh-alert/rh-alert.css
```diff
@@ -119,34 +119,33 @@ header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
...
-  margin-bottom: var(--rh-space-lg, 16px);
+  margin-block: var(--rh-space-md, 8px) var(--rh-space-lg, 16px) !important;
 }
 
 footer.hasActions {
   margin-top: var(--rh-space-lg, 16px);
 }
 
 footer ::slotted([slot="actions"]) {
-  margin-right: var(--actions-item--MarginRight, var(--rh-alert__actions--Gap, 24px)) !important;
```
heyMP, line 131: 👍

heyMP:
Can we make a hard and fast rule saying that all `::slotted()` rules need to include `!important`?

bennypowers:
> ~Can~ Should we make a hard and fast rule saying that all `::slotted()` rules need to include `!important`?

"hard and fast" is maybe too far, but yes, in cases like this where we know what the style should be in all cases, we're allowed to - and are supposed to - clobber light styles

elements/rh-alert/rh-alert.css
bennypowers: yeah i saw that and I was like 

"ay yay yay yay yay!"

brianferry (changes requested):
Resolve to merge

brianferry (approved):
Looks good, tested in docs/ia-cta branch and looks good to me

PR #841, opened 2023-03-24
fix(footer): match sidegap to the design specs

heyMP:
## What I did

1. Moved the sidegap spacing rules from the responsive.css sheet to the the lightdom.css file.
2. Added three levels of spacing changes based on breakpoint. 
 - default: (--rh-space-lg, 16px)
 - (--rh-breakpoint-sm, 768px): (--rh-space-2xl, 32px)
 - (~~--rh-breakpoint-lg, 1200px~~ --rh-breakpoint-xl, 1440): (--rh-space-4xl, 64px)
3. Introduced a **global custom property** called `--rhds-section-side-gap` that will be the single source of truth for elements that need to keep their content aligned with the footer content.

## Testing Instructions

1. Go here: https://deploy-preview-841--red-hat-design-system.netlify.app/elements/footer/demo/
2. Make sure that the space between the viewport and the content matches the [design specs](https://xd.adobe.com/view/835616bd-1374-483d-ab10-6ae92e0e343c-d605/screen/4c0466ca-93a2-4846-a0b5-07378eeadc87).

## Notes to Reviewers

**Why did you move it to the lightdom?**

There was a flash of unstyled content that would be a bit jarring.

**What's with this global variable --rhds-section-side-gap?**

Other sites are going to struggle lining up their headers with our footer if we don't expose our business logic for these three levels of spacing.  The alternative would be that every change to our internal sidebar spacing styles would in essence be a breaking change.

I'm not sure how I feel about this yet though so I would love your input!

**Are you positive about the breakpoint values?**

No, not really.  I would appreciate a design review to gut check me.

## Screenshots

Screenshot 2023-03-20 at 9 02 19 PM

Screenshot 2023-03-20 at 9 02 07 PM

Screenshot 2023-03-23 at 9 08 54 PM

PR #845, opened 2023-03-24
refactor(footer)!: align footer file structure to RHDS standards

heyMP:
## What I did

1. We were toying with the idea of providing side-effect free versions of our components. That is, the ability for our users to import just the class definition without registering it with `customElements.define()`.  Until we flesh that out I think we should align footer with the structure of the other components.

## Testing Instructions

1.  Should be no regressions at all.

## Notes to Reviewers

Requires #841


elements/rh-footer/rh-footer.ts
```diff
@@ -1,18 +1,210 @@
...
 import '@rhds/elements/rh-accordion/rh-accordion.js';
 
-import { RhFooter } from './RhFooter.js';
-import { RhFooterUniversal } from './rh-footer-universal.js';
-import { RhGlobalFooter } from './rh-global-footer.js';
+export { RhFooterUniversal } from './rh-footer-universal.js';
+export { RhGlobalFooter } from './rh-global-footer.js';
+export { RhFooterSocialLink } from './rh-footer-social-link.js';
+export { RhFooterLinks } from './rh-footer-links.js';
+export { RhFooterBlock } from './rh-footer-block.js';
```
heyMP, line 15: I'm not sure if we should do this or not.  I don't really see a downside to exporting all of these classes. 🤷‍♂️

.changeset/soft-teachers-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`rh-footer`: removed the side-effect free `RhFooter.js` file in favor of the `rh-footer.js` import.
```
bennypowers, line 5:
~~~suggestion
---
"@rhds/elements": major
---

`<rh-footer>`: removed the side-effect free `RhFooter.js` module.

If you were relying on this to extend `RhFooter`, instead import from `rh-footer` or use `whenDefined`, Note that `rh-footer` will be registered already, so choose a new custom element name.
BEFORE:
```js
import { RhFooter } from '@rhds/elements/rh-footer/RhFooter.js';
```
AFTER:
```js
import { RhFooter } from '@rhds/elements/rh-footer/rh-footer.js';
```
or 
```js
const RhFooter = await customElements.whenDefined('rh-footer');
```

**NOTE**: `<rh-footer>` is carefully developed to adhere to brand guidelines. If you find that your project has need to `extend RhFooter`, please reach out to the design systems team, so that we can work with you to use `<rh-footer>` as-is.
~~~

elements/rh-footer/rh-footer.ts
bennypowers: we'd have to support them at these paths. better not to export them and treat them as 'private'

elements/rh-footer/rh-footer.ts
```diff
@@ -1,18 +1,210 @@
...
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette: ColorPalette = 'darker';
+
+  override connectedCallback() {
+    super.connectedCallback();
+    // wire up accessibility aria-labels with unordered lists
+    this.updateAccessibility();
+  }
+
+  #compact = false;
+
```
bennypowers, line 92:
```suggestion
  #logger = new Logger(this);

  #compact = false;
  
  @colorContextProvider()
  @property({ reflect: true, attribute: 'color-palette' }) colorPalette: ColorPalette = 'darker';

  override connectedCallback() {
    super.connectedCallback();
    // wire up accessibility aria-labels with unordered lists
    this.updateAccessibility();
  }

```

Obsessive whitespace refactoring :shrug:

bennypowers (changes requested):

elements/rh-footer/rh-footer.ts
bennypowers: maybe keep `RhFooterUniversal` though

bennypowers (approved):

PR #870, opened 2023-03-28
feat(range): add rh-range

nikkimk:
## What I did

- [X]  Made `rh-range` its own branch so that `rh-audio-player` can continue without the component.

Resolve accessibility issues identified by axe DevTools:
   - [ ] `aria-label attribute cannot be used on a rh-range with no valid role attribute`
   - [ ] `aria-hidden element must not be focusable nor contain focusable elements`

## Testing Instructions

1. Test range with axe DevTools.
2. Set value of range using keyboard only.
3. Set value of range using mouse.


markcaron:
@nikkimk this may need some revisiting, since it's old. Do you mind closing if this needs to be? Or provide more context if we're keeping this open? Thanks!

zeroedin:
Closing this PR as stale

Issue #872, opened 2023-03-28
[docs] Missing component docs


zeroedin:
Components without docs on ux.redhat.com/elements/**

 - [x] `rh-tag`
 - [x] `rh-badge`
 - [x] `rh-subnav`
 - [x] `rh-spinner`
 - [x] `rh-context-provider`  - Are we documenting this helper component?
 - [x] `rh-pagination`

zeroedin:
Going to stub these pages out for release, then we can follow up and present detailed pages later.

bennypowers:
Each of these also needs a `screenshot.svg`, and so do:
- blockquote

coreyvickery:
@zeroedin @bennypowers That works, Marionne and I can work on these missing pages and assets next week. cc @marionnegp

bennypowers:
Nice. There are more here: https://ux.redhat.com/elements/

bennypowers:
- Screenshots have to live in `elements/*/docs/screenshot.svg`
- Summaries:
  - if the element exists, it has to be in a `@summary` jsdoc tag in the element's `.ts` file
  - if the element doesn't yet exist, it has to be in the frontmatter of elements.html

If there are elements which have a summary in JSDoc, but it's not appearing in the listing, that's a bug in our code and should be fixed, either in our 11ty config/plugins/templates, or in pfe-tools
| Element | Has Summary | Has Screenshot |
| ------- | ----------- | -------------- |
| accordion | :x: | :heavy_check_mark: |
| alert | :x: | :heavy_check_mark: |
| audio-player | :heavy_check_mark: | :heavy_check_mark:  |
| avatar | :x: | :heavy_check_mark: |
| badge | :x: | :x: |
| blockquote | :x: | :x: |
| button | :x: | :x: |
| card | :heavy_check_mark:  | :heavy_check_mark: |
| cta | :x: | :heavy_check_mark:  |
| dialog | :x: | :heavy_check_mark: |
| footer | :x: | :x: |
| jump-links | :heavy_check_mark: | :heavy_check_mark: |
| navigation-primary | :heavy_check_mark:  | :x: |
| navigation-secondary | :x: | :x: |
| pagination | :x: | :x: |
| popover | :heavy_check_mark:  | :heavy_check_mark:  |
| progress-steps | :heavy_check_mark:  | :heavy_check_mark: |
| spinner | :x: | :x: |
| statistic | :x: | :x: |
| subnav | :x: | :x: |
[...]

PR #873, opened 2023-03-28
docs: add stubbed out pages for missing elements

zeroedin:
## What I did

1. Stubbed out docs pages for missing components

Reference: https://github.com/RedHat-UX/red-hat-design-system/issues/872 

## What I did not do

1. Add stubs for `rh-context-provider` not sure if we are formally documenting that element. 
2. Create content for these components.  

## Testing Instructions

1. View deploy preview, missing components should now be listed in `/elements` although with little to no content, other than the code page generated by JSDoc.


bennypowers (approved):
YOLO!

Issue #874, opened 2023-03-29
[bug] Alias urls regression


zeroedin:
### Describe the bug

There is an issue with alias URLs as a regression in dev server and docs.

http://localhost:8000/components/navigation-(secondary)/demo/ should not contain parens.

http://localhost:8080/elements/navigation-secondary/demo/ are serving 404s.

11ty is creating to folders `/_site/elements/navigation-(secondary)` where the demos exist and `/_site/elements/navigation-(secondary)` where all the documentation pages exists.

### Which Component?

aliased components with parens

### Steps to reproduce

1. Go to https://ux.redhat.com/elements/navigation-secondary/demo/
2. Then go to https://ux.redhat.com/elements/navigation-(secondary)/demo/

The 2nd URL loads but other assets are then broken as they are looking in the same folder which they do not exist

### Expected behaviour

Only one folder for navigation-secondary should be created

### Screenshots

![Screenshot 2023-03-29 at 2 32 58 PM](https://user-images.githubusercontent.com/13666/228653721-786bab02-9392-44b9-870a-73b9e3401cd2.png)

### Operating System (OS)

macOS

zeroedin:
I believe the issue stems from upstream in `pfe-tools` specifically `Manifest.ts` in `getDemoMetadata()`, the slug should maybe not return with special characters like parenthesis.

```
const [, slug = ''] = permalink.match(/\/components\/(.*)\/demo/) ?? [];
```
should be updated to:
```
let [, slug = ''] = permalink.match(/\/components\/(.*)\/demo/) ?? [];
slug = slug.replace(/[()]/g, '');
```

Slugs should most likely always follow `snake-case` with nothing but `[a-zA-Z0-9]`.

zeroedin:
PR upstream: https://github.com/patternfly/patternfly-elements/pull/2460

bennypowers:
Consulting the deploy logs, `/elements/navigation-secondary/demo/index.html` exists, as do the subresources. As well, the [immutable url for the deploy](https://642940c52b512e000890b73f--red-hat-design-system.netlify.app/elements/navigation-secondary/demo/) contains the working demo. Perhaps this has to do with redirect rules, browser cache, or the github pages deploy?

PR #878, opened 2023-03-30
fix(navigation-secondary): realign color-palette with surface tokens

zeroedin:
## What I did

1. Realigned `color-palette` default to new surface token values

## Testing Instructions

1. View [Deploy Preview,](https://deploy-preview-878--red-hat-design-system.netlify.app/elements/navigation-secondary/demo/) navigation-secondary background should match `--rh-color-surface-lighter` value of #f2f2f2

### DIFF
--- a/elements/rh-navigation-secondary/rh-navigation-secondary.ts
+++ b/elements/rh-navigation-secondary/rh-navigation-secondary.ts
@@ -21,7 +21,7 @@ import { ScreenSizeController } from '../../lib/ScreenSizeController.js';
 import { colorContextProvider, type ColorPalette } from '../../lib/context/color/provider.js';
 
 export type NavPalette = Extract<ColorPalette, (
-  | 'light'
+  | 'lighter'
   | 'dark'
 )>;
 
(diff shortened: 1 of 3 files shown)


zeroedin:
DP is broken due to https://github.com/RedHat-UX/red-hat-design-system/issues/874

elements/rh-navigation-secondary/rh-navigation-secondary.ts
```diff
@@ -106,7 +106,7 @@ export class RhNavigationSecondary extends LitElement {
 
 
   @colorContextProvider()
-  @property({ reflect: true, attribute: 'color-palette' }) colorPalette: NavPalette = 'light';
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette: NavPalette = 'lighter';
 
   /**
    * If the host color-palette="light", the cta color context should be on="light"
```
brianferry, line 112:
```suggestion
   * If the host color-palette="lighter", the cta color context should be on="lighter"
```

brianferry: Same on line 114

marionnegp (approved):
👍

PR #908, opened 2023-04-11
fix(playground): added base.css to playground

nikkimk:
## What I did

Added `https://ux.redhat.com/assets/base.css` to each playground demo in `docs/_data/playground.cjs`.

## Testing Instructions

1. In [DP](https://deploy-preview-908--red-hat-design-system.netlify.app/) compare `elements/*/demo/` to the playground area in `elements/*`.
2. Styles should be the same.

## Notes to Reviewers
Right now I'm pointing to `https://ux.redhat.com/assets/base.css` but if we can find a better way to bundle  `docs/assets/base.css` with the playground files, we should do that instead.


bennypowers (approved):
Lovely Grouped Tryout Mockups

PR #947, opened 2023-04-20
docs: fix element listing titles

bennypowers:
## What I did

1. fix missing titles in the element listing in `/elements/`


nikkimk (changes requested):
In the DP, Call to Action is still being sorted by "Cta" instead of its alias.

nikkimk (approved):
lgtm

PR #717, opened 2023-01-26
docs: tokens

bennypowers:
Closes #719 
Closes #883 

## What I did

1. Added a single-page tokens entry to docs
2. Added tokens pages for single categories

## Testing Instructions

1. [Overview Page](https://deploy-preview-717--red-hat-design-system.netlify.app/tokens/)

## Notes to Reviewers
- [x] Blocked by https://github.com/patternfly/patternfly-elements/pull/2316
- [x] Blocked by #725 
- [x] Blocked by https://github.com/RedHat-UX/red-hat-design-tokens/pull/68
- [x] Blocked by #868

PR #949, opened 2023-04-20
docs: create autogenerated CSS properties tables

marionnegp:
## What I did

1.

## Testing Instructions

1.

### DIFF
--- a/docs/_plugins/shortcodes.cjs
+++ b/docs/_plugins/shortcodes.cjs
@@ -219,4 +219,21 @@ ${content ?? ''}
 
     `;
   });
+
+  eleventyConfig.addPairedShortcode('renderCssPropertiesTable', function(content, { tagName } = {}) {
+    /**
+     * NB: since the data for this shortcode is no a POJO,
+     * but a DocsPage instance, 11ty assigns it to this.ctx._
+     * @see https://github.com/11ty/eleventy/blob/bf7c0c0cce1b2cb01561f57fdd33db001df4cb7e/src/Plugins/RenderPlugin.js#L89-L93
+     * @type {import('@patternfly/pfe-tools/11ty/DocsPage').DocsPage}
+     */
+    const docsPage = this.ctx._;
+    console.log({ tagName });
+    const allProps = docsPage.manifest.getCssCustomProperties(tagName);
+    console.log(allProps);
+    const props = allProps.filter(prop => !prop.name.startsWith(`--${tagName}`));
+
+    console.log(props);
+    return `found ${props.length} tokens`;
+  });
 };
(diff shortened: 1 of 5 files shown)


zeroedin:
Was looking to jump in on this PR and see if I could push it over the finish line, but have questions about what the final output is supposed to be.

On the code tab we have a shortcode `renderCssCustomProperties` however from the screenshot in the main content of the PR this looks to generate something different.  @coreyvickery  or @bennypowers do you have additional context that might help me take this on?

coreyvickery:
@zeroedin Original big idea was Marionne and I wanted to put design token names into `rh-table` for all docs pages. Anything from color, typography, space, etc. We were going to do this all by hand, but rightfully so, Benny corrected us and wanted it automated.

For the sake of time—at least until RHDS 1.0 docs are released—I think we should only featuring non-token spacer values in a hand-coded table (I don't think it's `rh-table`). A sample ux dot page is below.

https://xd.adobe.com/view/11e76e09-10f4-44af-b74b-1f3e562970e5-44b4/screen/751ced98-a149-4a12-b8e9-842450cacdb5

After the initial docs for RHDS 1.0 are pushed live, we can come back to this and push for more automation and more representing design tokens on ux dot.

cc @bennypowers

bennypowers:
thanks @coreyvickery. @marionnegp had discussed offline a compromise position that we could further refine down the line:

1. all elements should document all of their non `--_private` css vars in their custom-elements manifests, including system tokens
2. docs pages should 'just print' all the tokens from the manifest. perhaps they can be collated by category, or split up into sections, i.e. "here are the element-specific vars, here are the system tokens at play, by category"

system tokens in the table could perhaps link back to their token pages, once #717 lands

bennypowers:
is this still relevant?
https://ux.redhat.com/elements/avatar/code/#css-custom-properties

What these tables should do is apply descriptions and defaults based on the tokens.json from `@rhds/tokens`. If it would help, @zeroedin we can expose more js apis in that package to provide token metadata as well as values

zeroedin:
> metadata

That would be helpful.

zeroedin:
@coreyvickery My assumption is we will want to update the styles here to match the look and feel provided by the new docs XD's, current design is based off the old images.

image

coreyvickery:
@zeroedin Yes, let's update the styles of the spacers to resemble the new ones in my XDs. Table looks good too, but I would want each column to have the same width.

bennypowers:
See #1005

bennypowers:
See also #1003

zeroedin:
Adding screenshot of what the full range looks like since the PR is missing this as an example:

image

coreyvickery (approved):
@zeroedin Looks great until we get `Red Hat Mono` added.

PR #953, opened 2023-04-24
docs: sort elementDocs alphabetically by slug

eyevana:
## What I did

1. We're using the `elementDocs` collection when generating the side nav, but it's not created/sorted with the other collections used in the `RHDSAlphabetizeTagsPlugin`. Before returning, I'm sorting `elementDocs` by the slug since it takes a component's alias into consideration. 

## Testing Instructions

1. Run the docs with `npm run start` locally and verify that the elements are listing in alphabetical order in the side navigation under `Elements`.

### DIFF
--- a/docs/_includes/component/header.njk
+++ b/docs/_includes/component/header.njk
@@ -122,7 +122,7 @@
               <li class="site-navigation__sub-menu__item">
                 {{ menuLink('All elements', '/elements/', 'site-navigation__sub-menu__link') }}
               </li>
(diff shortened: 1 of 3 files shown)


brianferry (approved):
Checked this in the deploy preview and locally, looks good

bennypowers (approved):
Sharp, on point, easy to understand: in short, an excellent MR!

bennypowers (approved):

PR #961, opened 2023-04-28
fix: add/improve screensize calculation for footer and navigation secondary

zeroedin:
## What I did

1. Added initial load screen size calculation for `rh-footer`
2. Improved initial load screen size calculation for `rh-navigation-secondary` by removing the screen size calculation from the render method itself and moved it into connectedCallback so it only calculates on initial load.  In the render method it would run on all updates rather then only trigger from updates from the onChange event in the screensize controller.  

## Todo

- [ ] Add changeset

## Testing Instructions

1. View Deploy preview for [Footer](https://deploy-preview-961--red-hat-design-system.netlify.app/elements/footer/demo/) and [Navigation Secondary ](https://deploy-preview-961--red-hat-design-system.netlify.app/elements/navigation-secondary/demo/)

## Notes to Reviewers

Confirm that when loaded in a Mobile sized view or in Desktop sized view that state of each component, footer and secondary nav are properly loaded to their layouts.  Footer having an accordion, and secondary nav having the mobile menu in mobile view, then both without in desktop.

### DIFF
--- a/.changeset/olive-dragons-promise.md
+++ b/.changeset/olive-dragons-promise.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`rh-footer`: fixes mobile layout on initial load


zhawkins (approved):

heyMP (approved):
Very clean pattern.  Looks great 😍

PR #982, opened 2023-05-06
fix: replace "postinstall" to "prepare"

### LINKED ISSUE #981: [bug] wireit: command not found
### Describe the bug

The  devDependency `wireit` had been added into "postinstall"

https://github.com/RedHat-UX/red-hat-design-system/blob/fde62697383da2bb8922d516785dc0975fa292f9/package.json#L63

The projects that depend on `@rhds/elements` should not need install `wireit`

### Which Component?

no component relavent 

### Steps to reproduce

just did a `npm install @rhds/elements` in other projects

### Expected behaviour

install without any error

### Screenshots

_No response_

### Element HTML

_No response_

### Operating System (OS)

macOS

### Browser

Chrome

### More browser information

_No response_

### Additional Info

_No response_

kunyan:
The `prepare` scripts only works for dev

https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts

fix #981

## What I did

1. replace "postinstall" to "prepare"

## Testing Instructions

1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "clean": "git clean -fX -e node_modules",
     "clean:nuke": "git clean -fX",
     "lint": "wireit",
-    "postinstall": "wireit",
+    "prepare": "wireit",
     "patch": "patch-package",
     "husky": "husky install",
     "prepack": "npx pinst --disable",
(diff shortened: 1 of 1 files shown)


bennypowers:
I don't think this will help: `prepare` is definitely run locally when installing the package:

https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts

PR #983, opened 2023-05-08
chore: remove husky and patch-package

### LINKED ISSUE #981: [bug] wireit: command not found

bennypowers:
closes #981 
closes #982 

## What I did

1. removed husky and patch-package
2. removed postinstall script

we'll rely on ci for linting, and if wireit breaks for contributors, either fix our wireit config, or train contribs to clear the wireit cache themselves

## Notes
The alternative is to use `prepare` instead of `postinstall`


bennypowers:
might need a changeset

PR #985, opened 2023-05-09
docs(avatar): fix css data type

bennypowers:
see https://drafts.csswg.org/css-values/#component-multipliers

## What I did

1. fix the css data-type syntax for a list of colors


zeroedin (approved):
LGTM

Issue #987, opened 2023-05-11
[rh-tag] outline variant does not have the correct styles applied to it


brianferry:
### Describe the bug

`rh-tag` is no longer inheriting the styles related to the `outline` variant from `BaseLabel`.  Because of this, no matter what is put in to that attribute the tag always looks like a solid tag.

### Which Component?

rh-tag

### Steps to reproduce

1. Go to https://codepen.io/OhHeyBrian/pen/gOBzNZV
2. Observe that the tag is not outlined even though the second has the following html

```html
<rh-tag variant="outline" color="blue">Outline Blue Tag</rh-tag>
```

### Expected behaviour

The tag should be outlined when the `outline` variant is used.

### Element HTML

```HTML
<rh-tag variant="outline" color="blue">Outline Blue Tag</rh-tag>
```

### Operating System (OS)

macOS, Linux

brianferry:
May 19, 2023 Update - Added the outline variant to the tag in the PR.  Added the dark theme for a tag without a variant.  (White text, transparent background)

brianferry:
PR has been merged to address this

Issue #934, opened 2023-04-14
[docs] Tag docs need review/cleanup


hellogreg:
Please review sections and ensure docs are complete.

Current tabs: Overview, Code. No content in the Overview tab, currently. “Overview” heading in the Code tab refers to it as “tooltip” and has no additional description. No Guidelines tab.

Previous IA Docs update issue: none

Tag @coreyvickery for final design review.

coreyvickery:
XD: https://xd.adobe.com/view/11e76e09-10f4-44af-b74b-1f3e562970e5-44b4/
Content doc: https://docs.google.com/document/d/130NPvOgY00F2RkYwzeHNTgx-t7NSyUS44gbNR6QngIM/edit#
Images: https://www.dropbox.com/sh/5diilvza2kwfcbu/AAB8SyeGmgA__OJ7dE8slr55a?dl=0

brianferry:
May 19, 2023 Update - Tag has the updated docs, images, content.  Code review currently happening

brianferry:
PR has been approved & merged.

coreyvickery:
@brianferry One super fast edit.

### Style page
- [ ] Change `Light Theme` to `Light theme` and `Dark Theme` to `Dark theme`

PR #988, opened 2023-05-11
fix(tag): outline variant

brianferry:
## Related issues
#987 
#934 

## What I did

1.  Added a fix into `rh-tag` to add the color `white` and the `outline` variant styles.
2. Added the content into the docs pages into `rh-tag` from #934

## Testing Instructions

1. Please go to the DP.
2. Check the following tabs under `rh-tag` and ensure the content matches:
  - `Overview`
  - `Style`
  - `Guidelines`
  - `Code`
  - `Accessibility`


bennypowers (changes requested):
needs a minor changeset
and if `color=white` was valid and worked in 1.0 we can't remove it until next major, but we can deprecate the white option

zeroedin:
> > needs a minor changeset and if `color=white` was valid and worked in 1.0 we can't remove it until next major, but we can deprecate the white option
> 
> We're adding color white in with these changes, it doesn't appear that it was valid in the previous versions. I've added the changeset

@brianferry outline variant exists in `pf-label` I believe the original specs that I followed for `rh-tag` did not offer this variant, only the colored versions.  

Here is the XD specs im looking at:
https://xd.adobe.com/spec/3e62e93c-8338-4f31-5040-3b81f0ed5c71-4853/screen/6a1953ee-4519-43d3-b02e-13453611c015/specs/

I don't see a white variant but I see the outline variant that then has a white background.  @coreyvickery is this the most up to date XD for tag/label? (should prolly change the component name in the XD too)

coreyvickery:
@zeroedin The white variant is for applications like this:

https://xd.adobe.com/view/6b678a8a-9d3d-4fb7-894f-035a8b78021d-3e34/screen/848d10eb-386f-416b-a5b0-2c4cfba22ec4/

We can add it at a later date.

cc @brianferry

brianferry:
@coreyvickery - It's currently included in this PR but I can remove it if we'd like.  I've just made it where the white color automatically gets the outline variant since it's only for dark modes but I don't know if we'd also like a filled in version that includes a white background and black text?

bennypowers:
is this a correct summary?
- keep outline variant
- remove `white` color option
- the styles that were applied to `white` color option, apply them when the tag is on a dark context instead

bennypowers (changes requested):
needs a screenshot.png / svg

elements/rh-tag/docs/20-guidelines.md
```diff
@@ -0,0 +1,155 @@
...
+
+## White Tag
+
+The unfilled white tag should be used in the dark theme or on dark backgrounds. 
+It should also be used on its own and not be grouped. The text should not 
+indicate a status, it should be written to be a descriptive caption to elements 
+nearby.
+
+{% example palette="light",
+        width=1000,
```
bennypowers, line 54: this overflows the page

image

brianferry: Looks like this is a problem on many images depending on the sizing of the screen.  [Added some css here](https://github.com/RedHat-UX/red-hat-design-system/blob/be2cb48f41251224995810f2f4df03e14d92dccd/docs/scss/components/_example.scss#L215) to fix this for the images.  Though they end up looking pretty small on smaller screens without zooming in.

elements/rh-tag/rh-tag.ts
```diff
@@ -18,32 +22,83 @@ export type TagColor = (
...
+  /** The variant of the label. */
+  @property() variant?: 'filled' | 'outline' = 'filled';
 
+  /** The color of the label. */
   @property() color?: TagColor;
 
+  @colorContextConsumer() private on?: ColorTheme;
+
+  @colorContextProvider()
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
```
bennypowers, line 100: This doesn't need to be a provider as it usually doesn't have any child elements

bennypowers (changes requested):

bennypowers (approved):

Issue #979, opened 2023-05-05
[docs] example image width should be auto


nikkimk:
### Description

By default an example image's width should be its natural width. Right now we end up having to manually set the `--example-img-max-width` CSS variable on each image as an inline style so that it is the correct width. The problems are:

- There's an extra amount of code just to make the image do what an `img` tag without any styles does automaticallty.
- Do set image default width, you have to edit a variable named `max-width` when this variable should set max width.
- It's extra code on every image for something that should be defaulted to `auto`, making our docs more work to maintain.
- It's applying an inline style to something that doesn't have to.

### Suggested solution

I suggest making `.example img` set to `width: auto` and `max-width` to `var(--example-img-max-width, 100%)`

PR #991, opened 2023-05-16
docs: example image css

nikkimk:
## What I did

1. Resolves [#979](https://github.com/RedHat-UX/red-hat-design-system/issues/979). Adjusted default example image size since the example images are 2X the desired size by using: `width: auto; transform: scale(0.5); margin: -7.5% auto;`

## Testing Instructions

1. Check [DP](https://deploy-preview-991--red-hat-design-system.netlify.app/) for issues with image size.


bennypowers:
can we do this without `transform`?

coreyvickery:
@bennypowers Back when the whole website ran on `.njk` files, for each image, Marionne and I would add widths to a `max-width` property in the `` tag. For example:

`Anatomy of a tooltip`

Doing this always ensured the images were the exact width as designed in our XDs. Is there a way to get back to that point without using the `max-width` property? Am I exporting images from XD incorrectly? We have always exported them at 2x so they render sharply.

bennypowers:
`width` and `height` attrs are preferable because the browser can allocate space early on (especially vertical space with `height`)

As well, automated tools like jampack can use the `width` and `height` attrs to generate optimized image `srcsets` for us at build time

nikkimk:
@bennypowers I took your advice and used Eleventy's responsive image to generate srcsets. ptal.

bennypowers (approved):

PR #995, opened 2023-05-21
feat(code-block): add rh-code-block

bennypowers:
## What I did

1. code-block, toolbar, and panel
2. tabs pattern
3. copy button patterns
4. light/dark

## Testing Instructions

1. Usability testing on the DP

## Notes to Reviewers
Please review the text in the code tab. I wrote this myself, since the design docs don't cover these kinds of details.


zeroedin:
What if we added a prefix slot to tabs?

Then you could style tabs.html with:

```
.rh-tabs {
  --rh-tab-panel-padding: 0;

  max-width: 1000px;
}

rh-tabs::part(wrapper) {
  display: grid;
  grid-template-columns: 1fr max-content;
  grid-template-areas: "tabs actions" "panel panel";
}

rh-tabs::part(prefix) {
  grid-area: actions;
  display: flex;
  flex-direction: row;
  gap: var(--rh-space-sm, 8px);
  border-block-end: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle-on-light, #c7c7c7);
}

rh-tabs::part(container) {
  display: contents;
}

rh-tabs::part(tabs-container) {
  grid-area: tabs;
}

rh-tabs::part(panels) {
  grid-area: panel;
}
```

I'm not sure like the idea of sticking extra stuff in tabs slot that aren't tabs this would help to avoid that.

bennypowers:
This is maybe a case for disconnected tabs (`<rh-tab for="idref">`)

a rough sketch of the idea:

```html
<rh-toolbar>
  <rh-tab for="html">HTML</rh-tab>
  <rh-tab for="react">React</rh-tab>
  <zoopy-button floop="copify">Copy</zoopy-button>
  <rh-toolbar-panel>
    <rh-tab-panel id="html">
      <rh-code-block>Standards!</rh-code-block>
    </rh-tab-panel>
    <rh-tab-panel id="react">
      <rh-code-block>Who needs 'em?</rh-code-block>
    </rh-tab-panel>
  </rh-toolbar-panel>
</rh-toolbar>
```

heyMP:
I really like how composable it is.  I do wonder about the expandable feature.  How would you feel about putting that into as something that you could toggle on with a boolean?  I think it might be frustrating for a user to have to add the action buttons markup depending on the length of the code block content.

Example:
Screenshot 2023-06-23 at 4 45 49 PM

Code block is going to be used almost exclusively within WYSIWYG content blocks so it's also going to be tough to make changes if we make it too un-opinionated.

bennypowers:
I'd rather leave that to the pattern level, tbh

bennypowers:
question about names:

if we go with the pr as it is right now, which implements a widget that contains a toolbar and some content related to that toolbar, then what do we call those things?

as of [f5b9e74](https://github.com/RedHat-UX/red-hat-design-system/pull/995/commits/f5b9e74758d43cb4abd20531587680e4a68d86b0) I called the toolbar element `rh-toolbar`, and the containing elemen `rh-toolbar-panel` (currently prefixed)

i guess this is ok, BUT! compare to rh-tabs, where the containing element is called `rh-tabs`, the toolbar is a shadow part, and the content area is called `rh-tab-panel`

so maybe we should have something like

```html
<rh-thingy><!-- insert better name here -->
  <rh-button slot="toolbar">I'm a tool!</rh-button><!-- don't quote me on that -->
  <rh-tab slot="toolbar" for="a">I'm a tab button!</rh-tab>
  <rh-tab-panel id="a">A</rh-tab-panel>
</rh-thingy>
```

bennypowers:
Blocked by https://github.com/patternfly/patternfly-elements/pull/2523

bennypowers:
no longer blocked on https://github.com/patternfly/patternfly-elements/pull/2523, we'll save that for #1091

bennypowers:
@heyMP, @zeroedin we're kicking the can down the road, so sleep on your above objections for a minute ;)

elements/rh-code-block/docs/40-code.md
```diff
@@ -0,0 +1,72 @@
...
+
+{% renderMethods %}{% endrenderMethods %}
+
+{% renderEvents %}{% endrenderEvents %}
+
+{% renderCssParts for='rh-code-block', level=3 %}{% endrenderCssParts %}
+
+{% renderCssCustomProperties for='rh-code-block', level=3 %}{% endrenderCssCustomProperties %}
+
+{% renderInstallation %}{% endrenderInstallation %}
```
nikkimk, line 68: You have the render installation twice on this pages, once on top and once right here.

elements/rh-code-block/docs/50-accessibility.md
```diff
@@ -0,0 +1,6 @@
+{% include 'accessibility/ariaguide.md' %}
+{% include 'accessibility/wcag.md' %}
+{% include 'accessibility/2.1.1-A.md' %}
+{% include 'accessibility/2.1.3-AAA.md' %}
+{% include 'accessibility/2.4.3-A.md' %}
+{% include 'accessibility/2.5.5-AAA.md' %}
```
nikkimk, line 6: I recommend only including the accessibility sections that are applicable to the element here.

nikkimk (changes requested):

elements/rh-code-block/docs/40-code.md
bennypowers: thanks. fixed

elements/rh-code-block/docs/50-accessibility.md
bennypowers: these are the ones i found in the xd

could you send a commit to fix this page?

nikkimk (approved):
Low garbled talking mumbles

bennypowers:
@heyMP I'm not sure what the problem is :see_no_evil: 

The way i understood it, the code is a `<pre>` and there's a max width, hence the scrolling is intentional. does that make sense to you? did i misinterpret your report?

heyMP:
>

bennypowers:
thanks @heyMP that's clear.

Try putting `max-width: 100vw` on the code block element. I think that's a page issue

elements/rh-code-block/rh-code-block.css
```diff
@@ -0,0 +1,49 @@
...
+  background-color: var(--rh-color-surface-lighter, #f2f2f2);
+  border: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle-on-light, #c7c7c7);
+  border-block-start-width: var(--rh-code-block-border-block-start-width,
+    var(--rh-border-width-sm, 1px));
+  padding: var(--rh-space-xl, 24px);
+  font-family: var(--rh-font-family-code, RedHatMono, "Red Hat Mono", "Courier New", Courier, monospace);
+  color: var(--rh-color-text-primary-on-light, #151515);
+  max-width: 1000px;
+  max-height: 640px;
+  overflow-y: scroll;
```
heyMP, line 17:
```suggestion
  overflow-y: auto;
```

I think we should hide the scrollbar if we don't need it.  Example: https://codepen.io/heyMP/pen/wvQdwYa?editors=1010

bennypowers:
@coreyvickery at your leisure

coreyvickery (approved):
Looks good for now.

There will be a change coming to `border-radius` tokens that will impact `rh-code-block`.

(In the DP link on the `Overview` page, compare the background container border-radius seen in the image to the live element.)

PR #1003, opened 2023-05-28
docs: automate system tokens in manifests

bennypowers:
## What I did

1. automatically add description, default, and syntax to custom elements manifest for each system token (e.g. `--rh-color-red-300`)

## Notes to Reviewers
Check element docs for component token tables


bennypowers:
@zeroedin as you're working on #1005 and #949, PTAL here, too

nikkimk (approved):
Leverage great token metadata

PR #1004, opened 2023-05-30
fix(navigation-secondary): unclosed slot tag

bennypowers:
## What I did

1. close an unclosed slot element (install eslint in your editors!)

## Testing Instructions

1. check nav-sec demos


bennypowers:
@jvprh for each of those URLs, I executed this snippet in the javascript console to replicate the fix:

```js
document.querySelector('rh-navigation-secondary')
  .shadowRoot
  .getElementById('cta')
  .querySelector('slot')
  .children[0]
  .remove()
```

and didn't notice any changes, so i think we're good to go.

Issue #1023, opened 2023-06-01
[feat] example image no border for docs


nikkimk:
### Description

We need a plain image style with no border similar to how example images in the docs are responsive and use `srcset`.

### Suggested solution

Extend example image short code to palette="none" for a plain unbordered image.

### Example API

```HTML
{% example palette="none",
          alt="Examples of a tooltip used on large and small breakpoints",
          src="../tooltip-responsive-design.png" %}
```

### Additional context

Regular markdown images are not responsive and depend on hard-coded image widths. (See [feedback on tooltip](https://github.com/RedHat-UX/red-hat-design-system/pull/1021#issuecomment-1572821759) from #1021.)

PR #1024, opened 2023-06-01
docs(example-image-no-border): added styles for example image

nikkimk:
## What I did

1. Modified example SCSS for docs for an example image with no palette, or "none" as described in #1023 .
2. Added "none" option in example shortcode documentation for palette.

## Testing Instructions

1. Go to the [first image under Sizes heading on Spinner's Guidelines page](https://deploy-preview-1024--red-hat-design-system.netlify.app/elements/spinner/guidelines/#sizes) .
2. Image should have no padding.
3. Image should have no border.
3. Image should have no background.
4. Image should should be responsive.

## Notes to Reviewers
Style should be similar [what it currently looks like on ux.redhat.com](https://ux.redhat.com/elements/spinner/guidelines/#sizes) but possibly less blurry.

### DIFF
--- a/docs/scss/components/_example.scss
+++ b/docs/scss/components/_example.scss
@@ -80,6 +80,10 @@ rh-alert + .example {
   border: none;
   border-radius: none;
 
+  img:only-child {
+    margin: 0;
+  }
+  
   @include breakpoint(tango) {
     padding: 0;
   }
--- a/docs/scss/components/_example.scss
+++ b/docs/scss/components/_example.scss
@@ -75,6 +75,16 @@ rh-alert + .example {
 //   margin-bottom: 0;
 // }
 
+.example--palette-none {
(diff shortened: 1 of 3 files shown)


docs/scss/components/_example.scss
```diff
@@ -75,6 +75,16 @@ rh-alert + .example {
 //   margin-bottom: 0;
 // }
 
+.example--palette-none {
+  padding: 0;
+  border: none;
+  border-radius: none;
+
```
zeroedin, line 82: In the navigation secondary docs the images that would use this look and feel are not centered.  I wonder if this is true all the time or if there should be an additional attribute for inline-start alignment.

```Suggestion

  img:only-child {
    margin: 0;
  }
  
  ```

zeroedin (approved):
LGTM

PR #1044, opened 2023-06-06
docs: adjust example shortcode to skip srcset  for svgs

zeroedin:
## What I did

1. Adjust example shortcode to skip srcset generation for svg files.

## Testing Instructions

1. View deploy preview

## Notes to Reviewers

Ensure all images that use example shortcode and are svgs display at the correct sizes.


eyevana (approved):
Thank you!

PR #1051, opened 2023-06-08
docs: header permalinks

### LINKED ISSUE #119: [docs] Anchor linking for headings
Request to create internal anchor links on pages so we can click the heading and capture a direct link to that section:

Screen Shot 2020-10-30 at 1 35 24 PM

bennypowers:
Fixes #119

## What I did

1. add header-link style permalinks to markdown

## Notes to Reviewers

The approach taken in this PR differs from the original design in #119, for accessibility reasons, so please take a look @coreyvickery 

This initial draft does not style the header links any differently from regular links, but that's up for revision, naturally.

This PR relies on a patch to pfe-tools. If the approach taken here is desirable, please wait on merging this until we upstream the changes (cc @eyevana as this may affect cpelements docs site)


bennypowers:
Greg's best-of-all-worlds demo

https://codepen.io/hellogreg/pen/NWEPppE

patches/@patternfly+pfe-tools+1.3.0.patch
bennypowers, line 1: leaving this comment to make sure we don't merge before pfe tools is updated

bennypowers:
@hellogreg PTAL, I made the copy element wrap the headings.

zeroedin (changes requested):
When you scroll the page and then click the copy-permalink button the alert that pops up is underneath the `back-to-top` button.  We should apply a z-index to these alerts so they pop over that button as they can be dismissed and the `back-to-top` button can not.

image

zeroedin (approved):

PR #1056, opened 2023-06-09
docs: style tables tokenValue shortcodes

nikkimk:
## What I did

1. Created `tokenValue` shortcode that prints the value of a given token.
2. Replaced hex codes and border px with tokens in footer docs style tables.

## Testing Instructions

1. Make sure the [DP footer docs](https://deploy-preview-1056--red-hat-design-system.netlify.app/elements/footer/style/) output same content as [footer docs on ux dot](https://ux.redhat.com/elements/footer/style/).


markcaron:
Stale, and I believe we decided against doing this not too long ago. Closing.

PR #1062, opened 2023-06-13
docs: increase header margin

bennypowers:
## What I did

1. increase header margin on element docs (100px - (paragraph block margin===18px))
2. apply that margin to h3

### DIFF
--- a/docs/scss/_base/_base.scss
+++ b/docs/scss/_base/_base.scss
@@ -65,8 +65,9 @@ h1 {
   font-weight: 300;
 }
 
-body.element-docs h2 {
-  margin-block-start: 80px;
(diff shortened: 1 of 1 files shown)


coreyvickery (approved):
@bennypowers Looks good, thanks!

coreyvickery:
@bennypowers A caveat to this. When an h3 is below an h2, there leaves a large undesirable gap. When an h3 is used below an image or container, it looks fine, but there are several situations where h3s are used directly below h2s.

Per my spacing guidelines, the distance between an h2 and an h3 is 32px/2.0rem.

Screen Shot 2023-06-14 at 8 49 55 PM

What do you think we can do here?

PR #1063, opened 2023-06-13
docs: fix bad changeset package refs

bennypowers:
## What I did

1. fix changeset packagename
2. add a lint rule which will fail a PR if it has the wrong package name

## Notes to Reviewers
we can loop back to implement more [markdownlint custom rules](https://github.com/DavidAnson/markdownlint/blob/main/doc/CustomRules.md) that validate our [style guide](https://github.com/RedHat-UX/red-hat-design-system/wiki/Changesets)


package.json
```diff
@@ -145,6 +146,14 @@
         "lib/**/*.css"
       ]
     },
+    "markdownlint": {
+      "command": "markdownlint -c .markdownlint.cjs '.changesets/*.md'",
```
brianferry, line 150:
```suggestion
      "command": "markdownlint-cli2 -c .markdownlint.cjs '.changesets/*.md'",
```

brianferry (approved):

PR #1074, opened 2023-06-16
feat: contain focus outline to the width of the link text

heyMP:
## What I did

1. The flexbox setting on the parent `ul` container was forcing the link width to be 100% of its container.  I'm using fit-content to reduce it to just take up the width of the text inside the link.

## Testing Instructions

1.  Visit
2. Using `tab` move the focus state to the primary links of the footer.
3. Verify that it only outlines around the width of the link text.


bennypowers (approved):
less girth, trimmed much

PR #1075, opened 2023-06-20
docs: spacer table reduce font size, switch to outline to adjust fit

zeroedin:
## What I did

1. Reduces font size for spacer samples
2. Switched to outline from border to output more exact spacer size

## Testing Instructions

1. View deploy preview


coreyvickery (approved):
Looks good, thank you!

PR #1077, opened 2023-06-20
docs: wrap tokens tables in shortcode

zeroedin:
## What I did

1. Wrapped token tables in `{% tokensTable %}` shortcode

## Testing Instructions

1. View [deploy preview](https://deploy-preview-1077--red-hat-design-system.netlify.app/)

## Notes to Reviewers

All tables that included css hex codes should now be hidden.  Any associated headers with those tables should also be wrapped in the shortcode so they are no longer visible.


docs/_plugins/shortcodes/tokensTable.cjs
```diff
@@ -0,0 +1,11 @@
+module.exports = function(eleventyConfig) {
+  eleventyConfig.addPairedShortcode('tokensTable',
+    /**
+     * Tokens Table
+     * Display a table rows with token usage
+     */
+    function(content) {
```
bennypowers, line 7:
```suggestion
    function(content, options) {
```

include: string
exclude: string
only: 'tokens'|'cssProps'
types: string
element: string

I _think_ that would be enough? but if you think this is jumping the gun, let's defer until we talk it over with design

elements/rh-footer/docs/10-style.md
```diff
@@ -47,6 +47,8 @@ theme.
 
 ### Website-specific footer
```
nikkimk, line 48: Some of these headings have nothing but token tables below them. We might want to move those headings into the shortcode as well.

zeroedin: Example: https://deploy-preview-1077--red-hat-design-system.netlify.app/elements/footer/style/#website-specific-footer

These headers only contained token tables

docs/_plugins/shortcodes/tokensTable.cjs
zeroedin: Let's defer this till we know more what we want this shortcode to do.

coreyvickery (approved):
Looks good, thanks.

PR #1089, opened 2023-06-26
docs: make subnav sticky

### LINKED ISSUE #1060: [docs]: subnav should be sticky
[subnav docs](https://xd.adobe.com/view/258ee636-1778-4e80-9ee3-17a5568e3ef1-70dd/screen/1a3cd010-d9e5-4801-8a08-7d8e741c4375/specs/) (thanks @coreyvickery for putting these together) say that subnav should be sticky

bennypowers:
Closes #1060

## What I did

1. make the subnav's container sticky, calculating the height of the whitespace and header

## Testing Instructions

1. check out an elements page in the dp

### DIFF
--- a/docs/scss/components/_header.scss
+++ b/docs/scss/components/_header.scss
@@ -30,11 +30,20 @@
   // height of the h1 and whitespace which precedes the subnav
   // in the block direction from the top of the main article header
   top: calc(
-      72px - /* height of logo header*/
-      32px - /* l-main__header top padding*/
-      (35px * 1.3) - /* h1 font size x line height*/
-      (35px * 0.67) /* best guess at 0.67em margin-bottom for the h1*/
+      72px /* height of logo header*/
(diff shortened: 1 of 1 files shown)


brianferry:
Looks good on screens < 1400px, though I think we might need a different stylesheet for >= 1400px

image

Looks like when the screen is >= 1400px, the .header class gets a padding left, top, and right.  Also the padding on the h1 element changes because the font-size changes to 48px.

bennypowers:
Thanks @brianferry. PTAL. should fix large screens. mobile still has some issues, but I think we can merge anyways and go back after this release to fix the styles in a more fundamental way (i.e. use tokens)

that being said if you wanna tinker, go right ahead, just commit here and merge when it looks good to you

zeroedin:
> should fix large screens

Nice work, this looks great!

coreyvickery:
@bennypowers Looks good to me.

PR #1097, opened 2023-06-27
docs: add permalink/copy to repo status table header & use SVG for check

### LINKED ISSUE #424: [docs] component status table design tweaks
Not overly picky, but the x mark is not quite the center of each cell.  It's not a deal-breaker. I don't want to hold this up. we can adjust later if need be. Also, I think we can make the date the same color as the row line(#D2D2D2). I don't think it needs to be at the forefront with the other content but that is my opinion.

_Originally posted by @ronniemcbride in https://github.com/RedHat-UX/red-hat-design-system/pull/396#pullrequestreview-1031639741_

Also, instead of `x` we can use an icon from the icon collection:

> @coreyvickery @marionnegp Did we decide to go with X marks instead of Check marks? Also, we might be considering changing the marks used in the future. It is possible we might change it to a circle with a check-in from our icon set. @bennypowers swapping the mark for an RH-icon replacement should be considered. Will that be an issue later?

eyevana:
Closes #424

## What I did

1. Using markdown for repo status table header so that it automatically adds the permalink/copy functionality
2. Removed some markup and the heading element ID that I don't see being used anywhere

## Testing Instructions

1. Verify that the permalink/copy functionality appears and works on the `Repo status` heading.
2. Verify that you now see SVG checks now instead of the HTML entities

### DIFF
--- a/docs/_plugins/shortcodes/repoStatus.cjs
+++ b/docs/_plugins/shortcodes/repoStatus.cjs
@@ -17,9 +17,10 @@ function repoStatus({ heading = 'Repo status', type = 'Pattern' } = {}) {
   } else {
     return /* html*/`
 
-<section class="section section--palette-default container">
-  <a id="Component status"></a>
-  <h2 id="component-status" class="section-title pfe-jump-links-panel__section">${heading}</h2>
+<section class="section section--palette-default container">  
+  
+  ${`## ${heading} {.section-title .pfe-jump-links-panel__section} `}
+  
   <p>Learn more about our various code repos by visiting <a href="https://ux.redhat.com/about/how-we-build/" target="_blank">this page</a>.</p>
   <div class="component-status-table-container">
     <table class="component-status-table">
--- a/docs/_plugins/shortcodes/repoStatus.cjs
+++ b/docs/_plugins/shortcodes/repoStatus.cjs
@@ -21,11 +21,10 @@ function repoStatus({ heading = 'Repo status', type = 'Pattern', level = 2 } = {
   } else {
     return /* html*/`
 
-<section class="section section--palette-default container">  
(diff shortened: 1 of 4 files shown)


docs/_plugins/shortcodes/repoStatus.cjs
```diff
@@ -17,9 +17,10 @@ function repoStatus({ heading = 'Repo status', type = 'Pattern' } = {}) {
   } else {
     return /* html*/`
 
-<section class="section section--palette-default container">
-  <a id="Component status"></a>
-  <h2 id="component-status" class="section-title pfe-jump-links-panel__section">${heading}</h2>
+<section class="section section--palette-default container">  
+  
+  ${`## ${heading} {.section-title .pfe-jump-links-panel__section} `}
```
zeroedin, line 22: Would there ever be need to make this a different header?  `h3` etc

Doing something similar over here: https://github.com/RedHat-UX/red-hat-design-system/pull/1098/commits/bb5de95ddfde44f30eaf6e61cbace9caded58521

eyevana: @zeroedin I don't think so, but I can make the change to allow for it if we ever needed to

eyevana: @zeroedin do you know why L25 doesn't work, but L27 does? https://deploy-preview-1097--red-hat-design-system.netlify.app/elements/accordion/#repo-status

zeroedin: because its markdown i believe it needs an empty line above it

zeroedin (approved):
Looks Great Tables Made

PR #1098, opened 2023-06-27
docs: combine repeated code tab shortcodes into single shortcode

### LINKED ISSUE #1052: [docs] Subcomponent descriptions
Components all have descriptions.  However in the docs we are currently only displaying the descriptions from the parent component.  Subcomponents descriptions do not get displayed.  Should we display these descriptions under the sub component headers on the code tab?

image

### LINKED ISSUE #1073: [docs] shortcode that renders all docs code tab sections for an array of elements
### Description

For any given component, the code tab requires the same series of shortcodes. And for components with multiple custom elements, like accordion, footer, tabs, etc., that series repeats for every element which leave a lot of room for error:

```
## <element-tagname>
{% renderSlots for='element-tagname', level=3 %}{% endrenderSlots %}

{% renderAttributes for='element-tagname', level=3 %}{% endrenderAttributes %}

{% renderMethods for='element-tagname', level=3 %}{% endrenderMethods %}

{% renderEvents for='element-tagname', level=3 %}{% endrenderEvents %}

{% renderCssParts for='element-tagname', level=3 %}{% endrenderCssParts %}

{% renderCssCustomProperties for='element-tagname', level=3 %}{% endrenderCssCustomProperties %}
```

### Suggested solution

We should have a short code that takes a a single component or comma-separated list of components and generates the code about of each component in the list.

### Example API

`{% renderElementsCode for='rh-accordion, rh-accordion-heafer, rh-accoordion-panel', level=3 %}` or 
`{% renderElementsCode for='rh-cta', level=3 %}`

### Additional context

This was a discussion in [#1068](https://github.com/RedHat-UX/red-hat-design-system/pull/1068#pullrequestreview-1482073906).

zeroedin:
## What I did

1. Combined repeated shortcodes on the code tab to a single shortcode `renderCodeDocs`
2. Adds ability to hide and show description under component tag header `hideDescription`

Closes #1052
Closes #1073

## Testing Instructions

1. View deploy preview for each component look for regressions in information not shown or incorrect information shown.

### DIFF
--- a/docs/_plugins/shortcodes.cjs
+++ b/docs/_plugins/shortcodes.cjs
@@ -8,6 +8,7 @@ const Section = require('./shortcodes/section.cjs');
 const Demo = require('./shortcodes/demo.cjs');
 const SpacerTokensTable = require('./shortcodes/spacerTokensTable.cjs');
 const TokensTable = require('./shortcodes/tokensTable.cjs');
+const renderCodeDocs = require('./shortcodes/renderCodeDocs.cjs');
 
 /** @typedef {import('@patternfly/pfe-tools/11ty/DocsPage').DocsPage} DocsPage */
 
@@ -29,4 +30,5 @@ module.exports = function(eleventyConfig) {
   eleventyConfig.addPlugin(Demo);
   eleventyConfig.addPlugin(SpacerTokensTable);
   eleventyConfig.addPlugin(TokensTable);
+  eleventyConfig.addPlugin(renderCodeDocs);
 };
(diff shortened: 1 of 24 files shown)


bennypowers:
will this make it easier to migrate when we remove the templates from pfe tools, or harder?

zeroedin:
> will this make it easier to migrate when we remove the templates from pfe tools, or harder?

Thats a good question, my assumption is templates sort of just move downstream but we create an API in `pfe-tools` for the shared data access.

I tried to keep this PR simple by basically just combining those calls to a single call.  Not sure it will make it any more difficult to migrate later as we'd still have to make a call to the data and apply that to a template locally, maybe this way the short code can remain the same, but internally just calls locals templates and applies the data from the pfe-tools API.

So instead of calling `this.ctx._.renderEvents` and getting back HTML we'd call `this.ctx._.eventsData` and just get back a JSON object.  

Did you have something different in mind?

bennypowers:
Sounds good just wanted to try to keep our eyes in our heads

bennypowers (approved):

PR #1101, opened 2023-06-27
fix(audio-player): added resize observe to reposition menu

nikkimk:
## What I did

1. Added resize observer to check for player layout changes and update menu position accordingly

## Testing Instructions

1. 1Use the [default DP demo](https://deploy-preview-1101--red-hat-design-system.netlify.app/elements/audio-player/demo/).
2. Click the More options menu button (kebab icon) to open the menu.
3.  Adjust window width until the player layout changes.
4.  Verify that the menu position updates so that it stays with the player.

### DIFF
--- a/elements/rh-audio-player/rh-audio-player.ts
+++ b/elements/rh-audio-player/rh-audio-player.ts
@@ -248,6 +248,14 @@ export class RhAudioPlayer extends LitElement {
 
   #dir = new DirController(this);
 
+  #width = this.offsetWidth;
+
+  #resizeObserver = new ResizeObserver(() => {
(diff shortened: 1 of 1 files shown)


elements/rh-audio-player/rh-audio-player.ts
```diff
@@ -248,6 +248,14 @@ export class RhAudioPlayer extends LitElement {
 
   #dir = new DirController(this);
 
+  #width = this.offsetWidth;
```
bennypowers, line 251: this might break in framework apps. do we have a test for instantiating with document.createElement?

bennypowers (approved):

PR #1033, opened 2023-06-03
docs(dialog): fixes

eyevana:
Closes #927

## What I did

1.

## Testing Instructions

1.

PR #1103, opened 2023-06-28
docs(dialog): remove heading from style page

bennypowers:
Closes #1102

PR #1110, opened 2023-06-29
docs: remove tokens table heading

eyevana:
## What I did

1. Address this issue https://github.com/RedHat-UX/red-hat-design-system/pull/1033#issuecomment-1610770455

## Testing Instructions

1. Verify heading in that tokens table is no longer present


bennypowers:
https://github.com/RedHat-UX/red-hat-design-system/pull/1103

PR #1115, opened 2023-07-05
chore: update patternfly deps

bennypowers:
## What I did

1. update patternfly elements dependencies


bennypowers:
@zeroedin we need this before

zeroedin (approved):

PR #1118, opened 2023-07-05
docs: render installation prose

zeroedin:
## What I did

1. Added prose clarifying use of `npm install` and `import maps`
2. Added prose about future state CDN
3. Added lightdom block to all components with `-lightdom.css`

## Testing Instructions

1. View [deploy preview](https://deploy-preview-1118--red-hat-design-system.netlify.app/)

### DIFF
--- a/docs/_plugins/shortcodes/renderInstallation.cjs
+++ b/docs/_plugins/shortcodes/renderInstallation.cjs
@@ -3,7 +3,7 @@
 /**
  * @param {string} content
  */
-function renderInstallation(content) {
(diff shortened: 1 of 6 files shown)


heyMP:
I think we might have to expand on the import map docs but I'm not exactly sure what to point to right now.  The issue being there is some additional steps of creating an "unbundled" build that would support the import map.

But for now, I think we should roll with it.  We'll have the flat file CDN tool done soon as well as the CDN.

heyMP (approved):
LGTM 🚀

PR #1128, opened 2023-07-06
docs: don't bundle rhds before build

### LINKED ISSUE #962: [bug] endless loop in dev-server on `npm run start`
Running `npm run start` will run a endless loop on what might be caused by the minified package?  More investigation needed.

Video of bug https://youtube.com/shorts/OokdnGavq4A?feature=share

bennypowers:
closes #962

... hopefully?

## What I did

1. stop bundling rhds on each 11ty build
2. exclude `.js.map` from 11ty watch


zeroedin (approved):
Works for me, I could no longer trigger a loop in the build.

Llamas Graze Tall Maize

PR #1129, opened 2023-07-06
docs: feedback shortcode

bennypowers:
closes #1117

## What I did

1. make a new `feedback` short code
2. use that new shortcode in the `feedback` include
3. use that new shortcode on the pages which have feedback divs

## Testing Instructions

[/about/how-we-build/](https://deploy-preview-1129--red-hat-design-system.netlify.app/about/how-we-build/) | [uxdot](https://ux.redhat.com/about/how-we-build/)
[/design-code-status/](https://deploy-preview-1129--red-hat-design-system.netlify.app/design-code-status/) | [uxdot](https://ux.redhat.com/design-code-status/)
[/elements/](https://deploy-preview-1129--red-hat-design-system.netlify.app/elements/) | [uxdot](https://ux.redhat.com/elements/)
[/elements/accordion/ (for example)](https://deploy-preview-1129--red-hat-design-system.netlify.app/elements/accordion/ (for example)) | [uxdot](https://ux.redhat.com/elements/accordion/ (for example))
[/foundations/color/](https://deploy-preview-1129--red-hat-design-system.netlify.app/foundations/color/) | [uxdot](https://ux.redhat.com/foundations/color/)
[/foundations/grid/](https://deploy-preview-1129--red-hat-design-system.netlify.app/foundations/grid/) | [uxdot](https://ux.redhat.com/foundations/grid/)
[/foundations/personalization/](https://deploy-preview-1129--red-hat-design-system.netlify.app/foundations/personalization/) | [uxdot](https://ux.redhat.com/foundations/personalization/)
[/foundations/spacing/](https://deploy-preview-1129--red-hat-design-system.netlify.app/foundations/spacing/) | [uxdot](https://ux.redhat.com/foundations/spacing/)
[/foundations/typography/](https://deploy-preview-1129--red-hat-design-system.netlify.app/foundations/typography/) | [uxdot](https://ux.redhat.com/foundations/typography/)
[/get-started/](https://deploy-preview-1129--red-hat-design-system.netlify.app/get-started/) | [uxdot](https://ux.redhat.com/get-started/)
[/get-started/design-system-kit/](https://deploy-preview-1129--red-hat-design-system.netlify.app/get-started/design-system-kit/) | [uxdot](https://ux.redhat.com/get-started/design-system-kit/)
[...]

PR #1135, opened 2023-07-07
docs: update release notes page

markcaron:
## What I did

1. Moved the old release notes to a new `prerelease-notes.njk`
2. Reformatted the Release Notes to markdown (`release-notes.md`)
3. Updated the styling and structure according to Major/Minor versions rather than the monthly format we had before.
4. Added "Highlights" in the table rather than "Major Changes" and "Minor Changes" to make it higher-level for a wider audience.
5. Used GitHub links, both at the top and following each version's section, to take the user to a more detailed view in GitHub (i.e. why reinvent the wheel too much?)
6. Removed the "cross links" / feedback section for now, which will need to be added back once that shortcode PR merges (#1129).

## Testing Instructions

1. Run the deploy preview
2. Go to Release Notes
3. Does it look good?

## Notes to Reviewers

Go easy on me! 😅

I left a lot of CSS in a `<style>` block for now, so we could keep these specific styles scoped to this page, especially the temporary/preliminary `<rh-tile>` custom element which I used for the GitHub link at the top (since it's the perfect element for that type of link, IMO).

Since it's docs, we probably could release ASAP and not wait on Bulbasaur.


hellogreg:
**A few copy updates**

**1.1.0**

For consistency's sake, add a period after the Notes of this section:

Screencap of "Added outline variant for rh-tag" section

**1.0.0**

Add a space between "to" and the "<" symbol in these two Change entries:

Screencap of the two "Renamed" sections

Add periods at the end of these two Notes entries:

Screencap of the two "Renamed" Notes sections, missing periods

coreyvickery (approved):

hellogreg (approved):
lgtm!

Issue #1153, opened 2023-07-20
[bug] rh-avatar includes base 64 png which is bloating file size


zhawkins:
### Describe the bug

When bundling rh-audio-player, I noticed that there were base 64'd pngs in the bundle. With @nikkimk 's help we tracked it down to rh-avatar.

https://github.com/RedHat-UX/red-hat-design-system/blob/main/elements/rh-avatar/rh-avatar.ts#L13

Can we either break those out into separate svg's or external png files? That should reduce the JS file dramatically.

With the base 64'd png:

```
│ rh-audio-player-entry.js               │ 132.46 kB │
```

Without:

```
│ rh-audio-player-entry.js               │ 77.45 kB  │
```

### Which Component?

rh-avatar

### Steps to reproduce

1. Look at https://github.com/RedHat-UX/red-hat-design-system/blob/main/elements/rh-avatar/rh-avatar.ts#L13

### Expected behaviour

Base 64 images aren't included in JavaScript files

### Screenshots

![Screen Shot 2023-07-19 at 11 28 26 AM](https://github.com/RedHat-UX/red-hat-design-system/assets/2671663/a416268b-6080-42fc-b924-42eb51062daa)

### Operating System (OS)

macOS

coreyvickery:
@zhawkins @nikkimk SVGs attached.

![avatar-light-theme](https://github.com/RedHat-UX/red-hat-design-system/assets/45767595/2ed26e71-4f22-4adc-988c-aa02e43fd027)

![avatar-dark-theme](https://github.com/RedHat-UX/red-hat-design-system/assets/45767595/143a55fa-5e9b-490f-af13-6b705f3f9dc5)

PR #1154, opened 2023-07-20
fix(avatar): default image file sizes

nikkimk:
## What I did

1. Replaced default base 64 PNGs with SVGs in `rh-avatar`, resolving #1153 

## Testing Instructions

1. Check [Sample Component in DP](https://deploy-preview-1154--red-hat-design-system.netlify.app/elements/avatar/#sample-component).
2. Verify that [file size](https://github.com/RedHat-UX/red-hat-design-system/pull/1154#issuecomment-1644263472) is significantly smaller.


eyevana (approved):

brianferry (approved):

brianferry (approved):

elements/rh-avatar/rh-avatar.ts
```diff
@@ -10,10 +10,14 @@ import { colorContextConsumer, type ColorTheme } from '../../lib/context/color/c
...
-  saturated: '',
+const DEFAULT_AVATARS = {
+  light: html`<svg xmlns="http://www.w3.org/2000/svg" style="enable-background:new 0 0 36 36" viewBox="0 0 36 36" role="presentation" part="img">
```
bennypowers, line 14: we could perhaps achieve further gains by using the `svg` template tag and interpolating the dark or light colours, or (maybe even better) just by using css on the inline svg

nikkimk: We could. And I had considered it. I've had some issues in some of my own work in the past with colors not updating, but I can't remember if it was resolved by a requestUpdate() or if it was how a particular browser cached the SVG.

PR #1166, opened 2023-07-26
docs: tokens pages updates

bennypowers:
## What I did

1. added feedback section to tokens pages

## Testing Instructions

1. https://deploy-preview-1166--red-hat-design-system.netlify.app/tokens/

### DIFF
--- a/docs/tokens/index.md
+++ b/docs/tokens/index.md
@@ -3,6 +3,13 @@ layout: layout-tokens.njk
 title: Overview
 ---
 
+<style>
+  ul.multi-column--min-400-wide {
+    list-style-type: none;
+    padding-inline: 0;
+  }
+</style>
+
 ## Introduction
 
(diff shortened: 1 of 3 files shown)


docs/_plugins/shortcodes/example.cjs
```diff
@@ -18,34 +18,41 @@ const path = require('path');
...
-      widths: [width1x, width2x],
-      filenameFormat(id, src, width, format) {
-        const extension = path.extname(src);
-        const name = path.basename(src, extension);
-        // rewrite the default 2X image since we don't need two copies
-        return width === width2x ? `${name}.${format}` : `${name}-${width}w.${format}`;
-      },
-    });
-  } catch (error) {
-    return false;
```
bennypowers, line 35: @nikkimk NB: we should not have swallowed these errors, see next commit for the places which crashing here helped to fix

coreyvickery (approved):

PR #1179, opened 2023-07-31
docs: vpn links

### LINKED ISSUE #592: [feat] Standard way to show a link is gated
### Description

Can we standardize the way we handle links when they are going to something that is behind a login? Consider https://ux.redhat.com/components/link/. Should a link have an icon or description associated with it when it goes to a new link? What about when search results go to a gated link? Links in a navigation experience?

### Suggested solution

Icon, label or description associated with a link

### Example API

_No response_

### Additional context

These links could be inline in a page, in navigation, in search results, in learning paths, etc. We'd want to consider all of the context. It would be great to have guidelines at the link level for how we should handle this. 

Consider how customer portal is already doing this on their search results: https://access.redhat.com/search/#/?q=rhel

bennypowers:
Closes #592 

## What I did

1. add styles for `a.vpn`: displaying the brand padlock icon after vpn links
2. add `vpn` class to vpn links, including ctas on fts page

## Testing Instructions

1. check any element's code page installation section
2. check fts page

## Notes to Reviewers
This is a straw PR to get the ball rolling. if design likes it then great we can merge, otherwise we'll iterate until it satisfies.


brianferry:
image

Should the Workfront & CMS Help links also include the .vpn class?

bennypowers:
:shrug:  prolly?

marionnegp:
@bennypowers, the question of which icon to use needs to be taken to the design leads. Brand is also updating UI icons and may need to create a new icon.

@coreyvickery, I added this to the next leads agenda, and I think you'll be out. Let me know if you have anything specific you want me to mention.

marionnegp:
@bennypowers, I think this should be removed from the review queue while we figure out how to display these links. @markcaron, is this something you want to get done during Caterpie?

bennypowers:
I'm fine with that

markcaron (commented):
I wonder if we even need the class `.vpn`?

We could use the `href` attribute selector, like:

```
a[href^="https://red.ht/"],
a[href^="https://url.corp.redhat.com/"] {
  ...
}
```

This is based on the assumption that we're properly utilizing the VPN / URL shortener.

nikkimk (changes requested):
On the FTS page, the icon doesn't align properly with the CTA link.
Screenshot 2024-02-02 at 10 30 17 AM

zeroedin:
> On the FTS page, the icon doesn't align properly with the CTA link. Screenshot 2024-02-02 at 10 30 17 AM

Maybe a `:not(rh-cta a)` style is necessary then maybe we could make `rh-cta` self aware of internal links and have it style itself and make that a common pattern everywhere?

nikkimk (approved):
Lucid groundhog; tiny meteorologist

markcaron:
> Lucid groundhog; tiny meteorologist

Now in my top 5 LGTMs of all time! 🤣

PR #1184, opened 2023-08-02
docs: updated component max character / max word counts

nikkimk:
## What I did

Worked with @coreyvickery to determine which components/parts of components should max character counts in the guidelines pages of docs, and added to or updated docs accordingly:

- [x] Accordion (title was updated, panel doesn't need a max count)
- [x] Alert (already exists)
- [x] Announcement
- [x] Audio player (already exists for title and description, not needed elsewhere)
- [x] Avatar (does not need max count)
- [x] Badge (already exists)
- [x] Blockquote (does not need max count)
- [x] Button (already exists)
- [x] Call to action (already exists)
- [x] Card (does not need max count)
- [x] Code block (does not need max count)
- [x] Dialog (already exists)
- [x] Disclosure
- [x] Footer (does not need max count)
- [x] Jump links
- [x] Navigation, primary
- [x] Navigation, secondary (already exists)
- [x] Link(s)
- [x] Pagination (does not need max count)
- [x] Popover (already exists)
- [x] Progress steps
- [x] Promo
- [x] Spinner (already exists)
- [x] Statistic (already exists)
- [x] Sticky banner
- [x] Sticky card
- [x] Subnav  (already exists)
- [x] Tabs (already exists)
- [x] Tag (already exists)
- [x] Tile
- [x] Tooltip 

## Testing Instructions

1. From the DP, verify that each guidelines page has max character / word counts, if applicable.

## Notes to Reviewers

- This is a work in progress.
- We determined that some components or parts of components did not require a max count.

### DIFF
--- a/elements/rh-table/docs/30-code.md
+++ b/elements/rh-table/docs/30-code.md
@@ -2,6 +2,8 @@
 
 ## Usage
 
+{% alert state="warning", title="Warning" %} Ensure that the table follows the recommendations on the [accessibility tab](../accessibility) so that the table works with assistive technology. {% endalert %}
+
 ### Title
 
 Specify the title of the table using a `caption` element.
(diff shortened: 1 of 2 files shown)


nikkimk:
@bennypowers there should be text preceding all of the tables. Some of it existed before this PR. PTAL

bennypowers (approved):

marionnegp (changes requested):
Instead of "The maximum character count for the elements of a [insert element/pattern] are listed below", can we say, "The recommended maximum character count for the elements of a [insert element/pattern] are listed below and include spaces"?

nikkimk:
@marionnegp ptal

marionnegp (changes requested):
Changes are looking good!

- [ ] Didn't notice this last time, but on Audio Player's page, the first column header is "Tab count." Is that right?
- [ ] Do you think it's valuable to add a sentence about how the character count includes spaces? I was thinking it could be a caption under the tables.

marionnegp (approved):
Local Grandma Tramples Musician

PR #1187, opened 2023-08-05
docs: create example paired shortcode

eyevana:
## What I did

1. Timestamp docs necessitated that the example shortcode not only work images, but live elements as well. I'm converting the example shortcode into a paired shortcode as a result.

## Testing Instructions

1. Verify that existing shortcode usage wasn't impacted by the conversion

### DIFF
--- a/docs/_plugins/shortcodes/example.cjs
+++ b/docs/_plugins/shortcodes/example.cjs
@@ -58,7 +58,7 @@ async function getImageHTML(opts) {
 
 /** @param {import('@11ty/eleventy/src/UserConfig')} eleventyConfig */
 module.exports = function(eleventyConfig) {
-  eleventyConfig.addShortcode('example',
+  eleventyConfig.addPairedShortcode('example',
     /**
      * Example
      * An example image or component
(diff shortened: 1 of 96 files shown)


bennypowers (commented):
Can this be accomplished without modifying the example shortcode? considerable effort was expended to simplify that feature by making it images-only

could the example HTML for timestamp just be wrapped in a div with some classes applied?

Why does timestamp necessitate different treatment, wouldn't it be better for all element examples to have the same design?

eyevana:
@bennypowers Definitely! I can go the route of just adding divs with classes. I just wanted to take advantage of the shortcode since the styling was the same. Corey wanted a live examples of timestamp with the [`lightest`](https://deploy-preview-1149--red-hat-design-system.netlify.app/elements/timestamp/guidelines/#default), [`darkest`](https://deploy-preview-1149--red-hat-design-system.netlify.app/elements/timestamp/style/#dark-theme), and [`wrong`](https://deploy-preview-1149--red-hat-design-system.netlify.app/elements/timestamp/guidelines/#headings) styles), but your suggestion works too.

eyevana:
Reverting changes in timestamp PR and creating div with styles instead.

PR #1188, opened 2023-08-06
docs(card): explain element vs pattern

bennypowers:
## What I did

1. added an explanation on the card pattern page about rh-card element vs card patterns

## Notes to Reviewers
Design should review the approach taken here and comment. It seems to me that this page could use some prose for each pattern usage guidelines etc)


markcaron:
@bennypowers looks like this PR is waiting for copy/content, eh?

TBD on who picks this up.

CC: @hellogreg @coreyvickery @marionnegp

marionnegp:
Yeah, I think we can move the "Variants" section to the card pattern page. We might also want to add a link to the card pattern page in the ["Other components" section](https://deploy-preview-1188--red-hat-design-system.netlify.app/elements/card/guidelines/#other-components) of the card element docs.

bennypowers:
@marionnegp I've elaborated as much as seemed reasonable, but it wasn't clear to me how to implement most of the [patterns listed here](https://deploy-preview-1188--red-hat-design-system.netlify.app/patterns/card/#patterns). i think we should pair on this, or if you have examples of the patterns in question, please link to them

markcaron:
@marionnegp @hellogreg any remaining thoughts on this PR?

We should try to merge this soon, since some other PRs seem to require it. We can tweak after if needed.

docs/patterns/card/patterns/asset.html
```diff
@@ -1,79 +1,113 @@
...
+      <rh-card>
+        <svg slot="header" viewBox="0 0 296 171">
+          <rect fill="rgba(0 102 204 / 0.10)" width="296" height="171"/>
+        </svg>
+        <p>Lorem ipsum dolor sit amet consectetur adipiscing elit.
+          Nullam elifend elit sed est egestast, a solicitudin mauris
+          tincidunt</p>
+        <rh-cta slot="footer">
+          <a href="#">Call to action</a>
+        </rh-cta>
```
bennypowers, line 68:
```suggestion
        <rh-cta slot="footer"
                href="#">Call to action</rh-cta>
```
etc

marionnegp (commented):
Just some early feedback...
- I think the "Usage" section should be moved to a separate "Guidelines" tab, even if there's not a lot of info. Just helps to make the subnavs more consistent.
- Are code blocks getting added for each pattern?
- What do you think about having the code block span the entire accordion/disclosure and having it show more by default?
Screenshot 2024-07-22 at 12 41 18 PM
- The image card demo with the code blocks looks nice!

markcaron:
I believe I've added all the Card patterns except for the Video Card (which I commented out in the code).

@adamjohnson @zeroedin, We will need to discuss how to handle the play button and thumbnail in the pattern along with the described Modal solution. Do we wait for `<rh-video>` to handle the play button/thumbnail pattern?

@marionnegp can you look through the patterns and let me know which ones are missing?
Do we need to tackle the light gray versions of the cards too?

marionnegp:
@markcaron, I don't think you'd need to include the light gray versions. `<rh-card>` should be able to use any surface color, so that would be a lot of demos to include. 😅 

Looks like you've captured all the patterns, except for the one with the video thumbnail, which was discussed during today's office hours. I don't see that card pattern very often, so it might be ok to include in a minor release or the next milestone.

The bar for the title card is too tall. Is this an issue with `<rh-card>`?
Screenshot 2024-07-29 at 9 46 37 AM

zeroedin:
> The bar for the title card is too tall. Is this an issue with `<rh-card>`?

All headers as default have a margin-block-end on them (flow layer of uxdot styles).  

Because the h4 is slotted in the component and is not directly styled to override page styles this style bleeds through to the element.  

The way to fix it would be either: 

1. in the pattern to explicitly pass a `margin: 0;` on `rh-card#example-title-bar-card h4[slot]`
2.  in the component shadow styles slotted header style update `margin: unset` with an `!important` to override page style margins always in `#header ::slotted(:is(h1, h2, h3, h4, h5, h6))`

markcaron:
@marionnegp, I would think the latter of @zeroedin's proposed solutions is the answer to fixing the title bar height.

markcaron:
@zeroedin @marionnegp looks like we'll just need to [make a change here](https://github.com/RedHat-UX/red-hat-design-system/blob/staging/charmander/elements/rh-card/rh-card.css#L90) to fix this in Charmander. Will do in another PR

docs/patterns/card/patterns/explicit-themes.html
```diff
@@ -0,0 +1,21 @@
+<uxdot-pattern color-palette="light">
+  <h3 slot="heading">Lighter themed card</h3>
+  <rh-card color-palette="lighter">
+    <h2 slot="header">Card title</h2>
+    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla vitae elit 
+      libero, a pharetra augue. Nullam id dolor id nibh ultricies vehicula ut id 
+      elit. Donec id elit non mi porta gravida at eget metus.</p>
+    <rh-cta slot="footer"><a href="#">Call to action</a></rh-cta>
```
bennypowers, line 8:
```suggestion
    <rh-cta slot="footer" href="#">Call to action</rh-cta>
```

Etc

markcaron: Ok, I updated all of the `<rh-cta>` to match the updated API as suggested. Since this is a docs update, I don't believe this will be 100% accurate/reflective of the current API until Charmander goes out, eh?

marionnegp (commented):
I'm seeing a missing space in the image card example.
Screenshot 2024-08-06 at 11 21 29 AM

Is it possible to do add a bottom margin for something like `#header ::slotted(img) + #footer`? Maybe this is something to tackle in another PR?

markcaron:
> I'm seeing a missing space in the image card example.
> Is it possible to do add a bottom margin for something like `#header ::slotted(img) + #footer`? Maybe this is something to tackle in another PR?

Yeah. I left a comment in the code to solve later, once we have the image slot and other adjustments being made to card in Charmander, this should resolve itself.

image

The following PRs (in Charmander) will resolve the issues mentioned above:

- [Unset slotted header margins](https://github.com/RedHat-UX/red-hat-design-system/pull/1728) mentioned in the [above comment about Title cards](https://github.com/RedHat-UX/red-hat-design-system/pull/1188#issuecomment-2256012174)
- [Promo card variant](https://github.com/RedHat-UX/red-hat-design-system/pull/1569) creates the image slot and fixes padding/margin gaps between the slots mentioned in the [Image card comment](https://github.com/RedHat-UX/red-hat-design-system/pull/1188#pullrequestreview-2221638224)

bennypowers:
may i suggest this path forward?
1. the conflicts in staging/charmander should be resolved
2. this branch should be rebased to staging/charmander
3. there resulting conflicts should be resolved

marionnegp (approved):

markcaron:
@bennypowers sorry, for some reason I didn't see your last comment before merging.

PR #1190, opened 2023-08-08
docs: fix partial template

eyevana:
## What I did

1. Fixed a typo in the feedback partial

## Testing Instructions

1. Verify that the feedback section for pages under `Patterns` uses the correct markup
2. Verify that the feedback section for pages under `Elements` uses the correct markup


bennypowers:
See https://github.com/RedHat-UX/red-hat-design-system/pull/1129

PR #1194, opened 2023-08-10
fix(cta): context

bennypowers:
## What I did

1. remove context provider from cta
2. make context consumers register after first update

## Testing Instructions

1. check all the context integrations (nav, accordion come to mind)

### DIFF
--- a/.changeset/many-apes-relax.md
+++ b/.changeset/many-apes-relax.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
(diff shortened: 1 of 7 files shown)


zeroedin (approved):
Made one a small fix for the tests, otherwise this good to go.  Works well in accordion and secondary nav context demos.

zeroedin (changes requested):
Needs changeset

zeroedin (approved):
Let's Get This Moving

Issue #1198, opened 2023-08-14
[bug] rh-card public css property to modify background color


zeroedin:
Currently the `rh-card` does not provide a public css property to modify background color `--rh-card-background-color`.  This feature would support to proposed changes in https://github.com/RedHat-UX/red-hat-design-system/pull/1188

bennypowers:
duplicating comments from the PR, for design review:

Before we merge this, it's worth considering weather it would be wiser to advise users to change the surface color tokens for a given card or class of cards, rather than providing an overriding background variable.

My reasoning is as follows: if we expose the background color variable to users, then users who wish to implement for example a dark background will skip the use of our color context system all together. That is likely to lead to maintenance difficulties, feature creep, and design and accessibility violations.

On the other hand, if we advise users who wish to apply a custom background to explicitly change the surface token values for a given card or class of cards, then they will have to explicitly participate in the context system. Because the context surface tokens have the words "on light" and "on dark" in their names, the user will be subtly encouraged to provide values for both theme variants.

This would have the added advantage of inheriting the provided values to the card's context-consuming children e.g. accordions, which in most cases would be desirable, and could be explicitly opted out of with a child selector

brianferry:
Added docs to help developers customize background color through css tokens.  Can be seen at the bottom of the page here https://deploy-preview-1200--red-hat-design-system.netlify.app/patterns/card/

PR #1200, opened 2023-08-15
 docs(card): customizing theme colors

brianferry:
## Issue

#1198

## What I did

1.  Added `--rh-card-background-color` as a css variable to rh-card for #1188


.changeset/wild-apes-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`rh-card`: adding public css variable for card background color. (#1198)
```
bennypowers, line 5:
```suggestion
`<rh-card>`: adding public css variable for card background color
```

bennypowers (changes requested):
can you add a demo, please?

bennypowers:
Before we merge this, it's worth considering weather it would be wiser to advise users to change the surface color tokens for a given card or class of cards, rather than providing an overriding background variable.

My reasoning is as follows: if we expose the background color variable to users, then users who wish to implement for example a dark background will skip the use of our color context system all together. That is likely to lead to maintenance difficulties, feature creep, and design and accessibility violations.

On the other hand, if we advise users who wish to apply a custom background to explicitly change the surface token values for a given card or class of cards, then they will have to explicitly participate in the context system. Because the context surface tokens have the words "on light" and "on dark" in their names, the user will be subtly encouraged to provide values for both theme variants.

This would have the added advantage of inheriting the provided values to the card's context-consuming children e.g. accordions, which in most cases would be desirable, and could be explicitly opted out of with a child selector

brianferry:
> Before we merge this, it's worth considering weather it would be wiser to advise users to change the surface color tokens for a given card or class of cards, rather than providing an overriding background variable.
> 
> My reasoning is as follows: if we expose the background color variable to users, then users who wish to implement for example a dark background will skip the use of our color context system all together. That is likely to lead to maintenance difficulties, feature creep, and design and accessibility violations.
> 
> On the other hand, if we advise users who wish to apply a custom background to explicitly change the surface token values for a given card or class of cards, then they will have to explicitly participate in the context system. Because the context surface tokens have the words "on light" and "on dark" in their names, the user will be subtly encouraged to provide values for both theme variants.
> 
> This would have the added advantage of inheriting the provided values to the card's context-consuming children e.g. accordions, which in most cases would be desirable, and could be explicitly opted out of with a child selector

That's a good point, I was debating whether or not updating the token for the cards was the way to go or if giving them a background variable would be "easier" vs more correct.  If we did go with the token route we'd just want to document it as a pattern or in the element docs letting the user know how they could go about updating it for light, dark, or both contexts.

zeroedin:
> worth considering weather it would be wiser to advise users to change the surface color tokens for a given card or class of cards, rather than providing an overriding background variable

> debating whether or not updating the token for the cards was the way to go or if giving them a background variable would be "easier" vs more correct

Sounds like the example in the patterns section where this issue originated is incorrect instead.  If we do not want people overriding the background due to color context.  https://github.com/RedHat-UX/red-hat-design-system/pull/1188/files#diff-74187edef1126c20210dc56031164bf56de11f133eacd202a89143cca266baf1R133

If i understand correctly @bennypowers you are asking for something like this instead which would be strictly pattern.

```css 
rh-card.alt {
  --rh-color-surface-lightest: rebeccapurple;
  --rh-color-surface-lighter: #bada55;
  --rh-color-surface-light: cornsilk;
  --rh-color-surface-darkest: indigo;
  --rh-color-surface-darker: aliceblue;
  --rh-color-surface-dark: #FF91AF;
}
```
In this example as long as the `.alt` card isn't in a color context implementation would only have to override the default value for surface lightest (IIRC).

bennypowers:
@brianferry @zeroedin I agree, we should update the docs. WDYT if we pivot this PR by reverting the css changes and then updating the docs?

brianferry:
@marionnegp @coreyvickery would either of you be able to review the content under "Custom Theming" https://deploy-preview-1200--red-hat-design-system.netlify.app/patterns/card/#custom-theming .  I added two examples in there of customizing the background, border color, and the font color.

docs/patterns/card/index.md
```diff
@@ -160,5 +199,17 @@ A card can be used in light and dark themes.
     padding-inline: 0;
     padding-block: 0;
   }
+
+  rh-card.custom-light-theme {
+    --rh-color-border-subtle-on-light: #EF6461;
+    --rh-color-surface-lightest: #0081AF;
```
marionnegp, line 205: I suggest making the surface color lighter. Maybe use `#feeded`? It's a custom color used in the homepage hero.

docs/patterns/card/index.md
```diff
@@ -160,5 +199,17 @@ A card can be used in light and dark themes.
...
+    --rh-color-text-primary-on-light: #30292F;
+  }
+
+  rh-card.custom-dark-theme {
+    --rh-color-border-subtle-on-dark: #988B8E;
+    --rh-color-surface-darkest: #210B2C;
+    --rh-color-text-primary-on-dark: #F5FFC6;
```
marionnegp, line 212: Looking at another existing custom color scheme (the Command Line Heroes site), this can use:
border: `#5e40be`
surface: `#261a4c`
text: `#e8e4f5`

marionnegp (commented):
I made some suggestions to update the examples' colors.

The `--_context-background-color` is showing up next to the examples under the Theme and Custom Theme sections. Can those be removed?
Screenshot 2023-08-30 at 9 43 42 AM

docs/patterns/card/index.md
brianferry: Those look a lot better.  Thank you!

docs/patterns/card/index.md
brianferry: Awesome, thanks!

brianferry:
@marionnegp 

I added a `width: fit-content` for the `rh-context-provider` and that looks like it has fixed the issue with the background taking the full width of the docs.

marionnegp:
@coreyvickery, should people be able to add a headline to the header, like in [the slotted title card pattern](https://deploy-preview-1200--red-hat-design-system.netlify.app/patterns/card/#sample---slotted-title)? I think it makes the spacing look off, but I'm not sure if other teams have asked for it.

coreyvickery:
> @coreyvickery, should people be able to add a headline to the header, like in [the slotted title card pattern](https://deploy-preview-1200--red-hat-design-system.netlify.app/patterns/card/#sample---slotted-title)? I think it makes the spacing look off, but I'm not sure if other teams have asked for it.

This should probably be discussed in another issue, yes? @brianferry What do you think?

bennypowers:
> > @coreyvickery, should people be able to add a headline to the header, like in [the slotted title card pattern](https://deploy-preview-1200--red-hat-design-system.netlify.app/patterns/card/#sample---slotted-title)? I think it makes the spacing look off, but I'm not sure if other teams have asked for it.
> 
> This should probably be discussed in another issue, yes? @brianferry What do you think?

yes, please open a new issue for this

docs/patterns/card/index.md
```diff
@@ -115,6 +115,45 @@ A card can be used in light and dark themes.
...
+
+Examples include:
+
+<ul>
+    <li>--rh-color-surface-lightest</li>
+    <li>--rh-color-border-subtle-on-light</li>
+    <li>--rh-color-text-primary-on-light</li>
+</ul>
+
+For more information, please see the <a href="/elements/card/code/#css-custom-properties">card css custom properties</a>.
```
bennypowers, line 130:
```suggestion
For more information, please see the [card css custom properties](/elements/card/code/#css-custom-properties").
```

docs/patterns/card/index.md
```diff
@@ -115,6 +115,45 @@ A card can be used in light and dark themes.
...
+
+To customize a card the design tokens must be altered.  These design tokens are different depending on the context for the card (light or dark theme).
+
+Examples include:
+
+<ul>
+    <li>--rh-color-surface-lightest</li>
+    <li>--rh-color-border-subtle-on-light</li>
+    <li>--rh-color-text-primary-on-light</li>
+</ul>
```
bennypowers, line 128:
```suggestion
- `--rh-color-surface-lightest`
- `--rh-color-border-subtle-on-light`
- `--rh-color-text-primary-on-light`
```

These should also link to their token doc page hashes

bennypowers (changes requested):
The root problem here is that we don't have a docs page that explains what the colour context system is, how it works, and how to use (or abuse) it.

bennypowers (approved):

bennypowers:
leaving this open for @marionnegp to approve before merging

brianferry:
@marionnegp - Created a bug for the issue you raised https://github.com/RedHat-UX/red-hat-design-system/issues/1209

Issue #1205, opened 2023-08-28
[bug] headings on full width image card aren't working


diwanshi:
### Describe the bug

We should be able to provide a heading below the full width image on rh-card, but it seems it doesn't work

### Which Component?

rh-card

### Steps to reproduce

1. Go to [this page](https://ux.redhat.com/elements/card/demo/variants/)  and check the variant with full width image (last card)
2. Add any of the heading tags before `<p>` and provide a slot="header"
3. Check the heading, it doesn't show up
4. Try providing h2 as part of the body
5. The styles for heading doesn't take any effect

The only way we could make this work was to wrap the whole body inside a div and applying styles to the div but thats not the recommended approach.

### Expected behaviour

The heading is shown below the full width image card

### Screenshots

![Screenshot 2023-08-28 at 12 26 36 PM](https://github.com/RedHat-UX/red-hat-design-system/assets/5659299/8ce72bda-b629-41fd-9115-e5015f8cf21a)

![Screenshot 2023-08-28 at 12 26 11 PM](https://github.com/RedHat-UX/red-hat-design-system/assets/5659299/7bc7c971-b04d-484b-bb34-b1aaed817f68)

### Element HTML

```HTML
<rh-card class="full">
      <img slot="header" src="../kitten-400x250.jpeg">
      <h2 slot="header">Heading h2</h2>
      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.
         Nullam eleifend elit sed est egestas, a sollicitudin mauris
         tincidunt. Pellentesque vel dapibus risus. Nullam aliquam
         felis orci, eget cursus mi lacinia quis. Vivamus at felis sem.</p>
      <rh-cta priority="primary" slot="footer">
        <a href="#">Call to action</a>
      </rh-cta>
</rh-card>
```

### Operating System (OS)

macOS

### Browser

Chrome, Safari, Firefox

brianferry:
@diwanshi 

![image](https://github.com/RedHat-UX/red-hat-design-system/assets/9708152/6ae8ea1e-f86b-4175-86de-221518addb6d)


Is this something like what you would expect?

& with the following html?

```html
  <rh-card class="full">
    <img slot="thumbnail" src="kitten-400x250.jpeg"></img>
    <h2 slot="header">Headline, sm</h2>
    <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.
       Nullam eleifend elit sed est egestas, a sollicitudin mauris
[...]

PR #1207, opened 2023-08-31
fix(card): adding thumbnail slot, header styling for image / header combo

brianferry:
## What I did

1.  Added a thumbnail slot to the rh-card component so that users can have both the thumbnail image and an heading tag insider of the `<header>` block in the shadowdom. 
2. Styled the header block so that the image and the header (if they're both present) stack vertically.


brianferry:
@coreyvickery @marionnegp 

Need some feedback here with regards to the header sizing.  Not currently limiting the header image size here to give room for the image and the header tag.  Not sure if there should be a hard limit on that image size.

@nikkimk @hellogreg 

Both the thumbnail image and the heading tag are in the <header> block which seems okay but wasn't sure if the placement of these would cause any problems? (Image tag above heading tag)

hellogreg:
> Both the thumbnail image and the heading tag are in the block which seems okay but wasn't sure if the placement of these would cause any problems? (Image tag above heading tag)

If the image is non-decorative (i.e., needs `alt` text), I think you'd ideally want it to be after its associated heading in the source, even if it's first visually. That way users of assistive tech navigating by headings will experience it as part of that section. (In other words, if the image comes before the heading, a screen reader user may never know it's there.)

This is a case where the logical reading order doesn't necessarily match the visual reading order.

[W3C WAI tip example](https://www.w3.org/WAI/tips/developing/#reflect-the-reading-order-in-the-code-order)

Where it gets tricky is if both the image and the heading are links (as I know has been done when implementing cards in the past, and maybe present). Then visual focus order could be moving in an unexpected direction: tab focus moving up from the heading to the image. But I think we really shouldn't be having two or three links (image, heading, and CTA) to the same destination in cards anyway.

marionnegp:
@coreyvickery @brianferry, I'm not sure if we need a hard limit on the image size. Might be better to add guidelines to documentation if we need to make a recommendation.

bennypowers:
is it possible to do this without adding or removing a new slot name?

brianferry:
> is it possible to do this without adding or removing a new slot name?

We can instruct the user to include both their image and heading element in the header slot and update the css to show all the items in the header.  Then update our .full pattern css to include the styles needed.  

Pretty minor changes with that method

bennypowers (approved):
Nice, minimal change that gets the job done :)

bennypowers:
@diwanshi does this address your issue in #1205 ?

diwanshi:
@bennypowers @brianferry yes that works for us.

bennypowers:
@brianferry can you add a patch changeset? Then we're good to merge

brianferry:
@bennypowers added

PR #1210, opened 2023-09-05
docs(tokens): fix copy link url

brianferry:
https://github.com/RedHat-UX/red-hat-design-system/issues/1211

## What I did

1. Updating the copy link generated on the tokens doc to include the category as currently the links do not copy over correctly. 

## Testing Instructions

1.


docs/_plugins/tokens.cjs
```diff
@@ -75,7 +75,7 @@ function copyCell(token, variable) {
         <code slot="content">${variable}</code>
       </rh-tooltip>
       <rh-tooltip position="left-start">
-        <button class="copy-button" data-copy="https://ux.redhat.com/tokens/#${token.name}">
+        <button class="copy-button" data-copy="https://ux.redhat.com/tokens/${token.attributes?.category}/#${token.name}">
```
bennypowers, line 78: do we store the slug on the token object? attributes.category:
a) might not be available
b) might not be accurate - we document the tokens differently to how we make them

brianferry: From what I was seeing it looks like we have that slug in the attributes.category object.  The only places where it's missing is the headers like 'dimension' and in the $extensions objects.  But if we'd rather have that attribute surfaced elsewhere / on the parent we can do that?

bennypowers: we'll have to move the data processing in the `category` function to the `tokens` data object.
i'll look into this tomorrow

bennypowers: ok  you're right looks like the category attribute lines up

bennypowers (approved):

PR #1223, opened 2023-09-15
fix(navigation-secondary): correct support for current page indicator

### LINKED ISSUE #1191: [docs] rh-navigation-secondary current page indicator
From design system chat: 

> Wes Carr,
> secondary nav question
> hey hey. i'm not seeing the current page indicator working on live pages, but it's here in the docs. is there a timeline for it being updated or is there a manual way to toggle it on? https://ux.redhat.com/elements/navigation-secondary/guidelines/#current-page-indicator

Guidelines on https://ux.redhat.com/elements/navigation-secondary/guidelines/#current-page-indicator state we should have a current page indicator.  

### Discovery needed:
Should this responsibility fall on the component or implementation?

A `current` attribute or class could be added to the dropdown `<a>` to manipulate lightdom styles.  Is it possible that each implementation maybe have a different way of marking that current page? Such as using [attributes] vs .classes vs naming conventions such as `current` vs `current-page`

lightdom css example:

```css
:is(rh-secondary-nav-dropdown, rh-navigation-secondary-dropdown) a[current],
:is(rh-secondary-nav-dropdown, rh-navigation-secondary-dropdown) a.current {
    border-block-start-color: var(--rh-color-brand-red-on-light, #ee0000);
}

:is(rh-secondary-nav-dropdown, rh-navigation-secondary-dropdown)[color-palette="dark"] a[current],
:is(rh-secondary-nav-dropdown, rh-navigation-secondary-dropdown)[color-palette="dark"] a.current {
    border-block-start-color: var(--rh-color-brand-red-on-dark, #ff442b);
}
```

Implementors can currently use the example above to accomplish this requirement of the current page indicator

Dropdown links are "possible" index pages that never get navigated to unless the navigation fails to upgrade and the dropdown takes over from the default link function, so they themselves don't always represent a navigable page that will be "current".  However any number of sub pages in their dropdowns maybe current, so the CMS or implementation would have to determine when to make the dropdown show this indication.  Also its possible that a link exists under more then one dropdown in a navigation menu.

zeroedin:
## What I did

1. Corrected and updated support for current page indicator.

Closes #1191

### TODO:

- [x] Add tests for current page indicator.
- [x] Fix focus ring overlap
- [x] Patch or Minor change?

## Testing Instructions

1. View Deploy Preview


bennypowers (approved):
LGTM. @nikkimk?

nikkimk (approved):
Losing gold to maurauders

Issue #1174, opened 2023-07-27
[docs] Hide repo status section on Elements and Patterns pages


markcaron:
Per discussion: https://github.com/orgs/RedHat-UX/discussions/1173

We need to (at least temporarily) hide the repo status section on Elements and Patterns pages until we know how we're going to handle this type of info.

PR #1226, opened 2023-09-19
docs: don't render repo status table

eyevana:
## What I did

1. Returning an empty string instead of the repo status table until we find a better [solution](https://github.com/orgs/RedHat-UX/discussions/1173) for maintaining the data. 

## Testing Instructions

1. Go to all the pattern and element overviews and verify that the repo status table is hidden.

## Notes to Reviewers

I tried commenting them out in the MD files, but the comment was prematurely closed from the HTML comments in the shortcode code. 

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/1174


brianferry (approved):
Launches Gnu Temp Maintaining

PR #1229, opened 2023-09-20
docs(rh-timestamp): update changeset

eyevana:
## What I did

1. Updated changeset to match previously released feature (code block)


eyevana:
@bennypowers maybe the indentation I had in my changeset cause it to become malformed? In any case, I just copied the formatting from `rh-code-block`. I'm hoping that'll do the trick.

bennypowers (approved):

PR #1236, opened 2023-09-22
docs: decouple from patternfly tools

brianferry:
## What I did

1.  Updated the render install function to change the name so there is not a conflict with the pf tools render installation function

## Testing Instructions

1.


heyMP (commented):
LGTM 👍

heyMP (approved):

PR #1245, opened 2023-10-02
fix(cta): revert removal of colorPalette, deprecate instead

zeroedin:
## What I did

1. Revert removal of colorPalette, deprecate instead

### TODO:

- [x] Figure out language for how we want to mark up the changeset for this reversal.

## Testing Instructions

1.

## Notes to Reviewers

Please modify language in changeset as you see fit.


bennypowers (changes requested):
this is fine. I suggest putting all new lines in one place (even though it would not follow ordering guidelines) so that you can wrap those lines in `// START DEPRECATION WARNING` and `// END DEPRECATION WARNING` comments

This will also need an `@attr color-palette {@deprecated reason}` jsdoc on the class, and the pr (i.e. commit) title should be updated

zeroedin:
> This will also need an `@attr color-palette {@deprecated reason}` jsdoc on the class, and the pr (i.e. commit) title should be updated

Doesn't look like `cem` provides @deprecated on @attr like it does on @slot (see rh-footer). 

Maybe just a depreciated note in the description will suffice?

bennypowers:
> > This will also need an `@attr color-palette {@deprecated reason}` jsdoc on the class, and the pr (i.e. commit) title should be updated
> 
> Doesn't look like `cem` provides @deprecated on @attr like it does on @slot (see rh-footer).
> 
> Maybe just a deprecated note in the description will suffice?

we can manually update the cem in a post-build script then. keep the jsdoc tho

elements/rh-cta/rh-cta.ts
```diff
@@ -136,15 +145,44 @@ export class RhCta extends LitElement {
   /** Is the element in an RTL context? */
   #dir = new DirController(this);
 
+
```
bennypowers, line 148:
```suggestion
```

elements/rh-cta/rh-cta.ts
```diff
@@ -136,15 +145,44 @@ export class RhCta extends LitElement {
...
   #logger = new Logger(this);
 
   get #isDefault(): boolean {
     return !this.hasAttribute('variant');
   }
 
+  // START DEPRECATION WARNING
+  /* note: remove ColorPalette type, and property decorator import above */
+  /**
+   * @deprecated
```
bennypowers, line 158:
```suggestion
   * @deprecated do not use the color-palette attribute, which was added by mistake. use context-providing containers (e.g. rh-card) instead
```

bennypowers (approved):

PR #1246, opened 2023-10-03
fix(audio-player): add French translation 

simonlabonne:
Added French translation to the player

## What I did

1.

## Testing Instructions

1.


bennypowers (approved):
Cela me semble correct

Please add a patch changeset

PR #1249, opened 2023-10-04
feat: add react wrappers

zeroedin:
## What I did

1. Adds react wrappers build and export

### TODO

- [x] Add changeset
- [x] Fix upstream hardcoded values so wrappers generate with correct project namespace  https://github.com/patternfly/patternfly-elements/issues/2649

## Testing Instructions

1.


scripts/generate-react-wrappers.ts
```diff
@@ -0,0 +1,6 @@
+import { generateReactWrappers } from '@patternfly/pfe-tools/react/generate-wrappers.js';
```
bennypowers, line 1: `@lit-labs/react` got graduated to `@lit/react` with lit 3.0.0. we'll do that update separately i think

zeroedin:
This is waiting on upstream release

PR #1254, opened 2023-10-09
docs: playgrounds performance improvements

bennypowers:
## What I did

1. improved docs demo playground loading performance by bundling the playground-elements dependency

## Testing Instructions

1. check out a element code page and click "load demo", compare with ux.

## Notes to Reviewers
potential [security issue with sandbox urls](https://github.com/google/playground-elements#sandbox-security), but we don't have any code on the back end so :shrug:


docs/_plugins/shortcodes/rh-playground.js
```diff
@@ -0,0 +1,135 @@
...
+import { classMap } from 'lit/directives/class-map.js';
+
+import 'playground-elements';
+import '@rhds/elements/rh-button/rh-button.js';
+import '@rhds/elements/rh-spinner/rh-spinner.js';
+
+class RhPlayground extends LitElement {
+  static styles = css`
+    :host {
+      max-height: 785px;
```
zeroedin, line 11: Let's remove this max-height rule while we are at it.  This stops the element from being able to be resized vertically. 

```Suggestion
```

bennypowers: @marionnegp this was to prevent long demos from taking over the page

maybe max height + resizable?

zeroedin: fwiw I did the change locally, each subcomponent of playground has a height, I don't think the outer host can scale beyond the sum of those parts. (tab bar, editor, preview) Doesn't really look like this rule is doing anything in its current state.

zeroedin: You might be thinking of the next rule down for the `div` which also sets the same max-height which is the pre-load demo view.  That would allow long demos to take over.

bennypowers: ok whichever you see fit :) feel free to commit directly to the branch

zeroedin (commented):
Might be out of scope for this PR, but noticed that code was getting "mixed" in the playground demos as well.  Screen shot below shows context-picker somehow getting injected with the demo stylesheet and a head tag.

image

bennypowers:
> Might be out of scope for this PR, but noticed that code was getting "mixed" in the playground demos as well. Screen shot below shows context-picker somehow getting injected with the demo stylesheet and a head tag.
> 
> image

try clearing site data including cache and service workers

brianferry:
> > try clearing site data including cache and service workers
> 
> no go, happens in all browsers, firefox, safari and chrome.

FWIW it looks like this is the same behavior that exists currently on ux.redhat.com so not an issue with this PR but an issue we'll want to add

bennypowers:
can yyou open a new issue?

zeroedin (approved):
🚀

PR #1297, opened 2023-10-24
perf: speed up docs builds

bennypowers:
## What I did

1. Cache 11ty import map between builds

### DIFF
--- a/docs/_plugins/importMap.cjs
+++ b/docs/_plugins/importMap.cjs
@@ -2,33 +2,47 @@
 const { join } = require('node:path');
 const { pathToFileURL } = require('node:url');
 const { glob } = require('glob');
+const { AssetCache } = require('@11ty/eleventy-fetch');
 
-module.exports = function(eleventyConfig, {
-  inputMap = undefined,
-  defaultProvider = undefined,
-  localPackages = [],
-} = {}) {
-  const cwd = process.cwd();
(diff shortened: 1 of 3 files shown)


docs/_plugins/importMap.cjs
```diff
@@ -2,33 +2,47 @@
...
+  } else if (TOTAL.duration > 1000) {
+    console.log(
+      `🐢 Import map generator done in ${chalk.yellow(TOTAL.duration)}ms\n`,
+      `  Resolving local packages took ${chalk.yellow(RESOLVE.duration)}ms\n`,
+      `  Tracing RHDS sources took ${chalk.yellow(TRACE.duration)}ms`,
+    );
+  } else {
+    console.log(
+      `⚡ Import map generator done in ${chalk.blue(TOTAL.duration)}ms\n`,
+    );
```
zeroedin, line 29: Would be nice in the future on ci build in github if we could ship these somewhere to store and then pull back to generate a graph on performance over time as well.

bennypowers: It should be straightforward (hah!) To cache the outputs against a hash of the dependencies (currently this file and the eleventy config) and restore the flat cache on each run

zeroedin (approved):
Lines Generated Time Measured

PR #1302, opened 2023-10-25
feat(tile): form control

bennypowers:
## What I did

1. make tile a FACE

## Testing Instructions

1. check the unit tests, is anything missing?
2. manual testing in all the tile demos


nikkimk:
@bennypowers there's a visual regression.
Screenshot 2023-10-25 at 10 49 34 AM
Screenshot 2023-10-25 at 10 49 47 AM

nikkimk (changes requested):
There's a visual regression. Also RTI is now broken on tile groups.

bennypowers:
the visreg is probably because i replaced things like `<div id="foo"><slot name="foo">` with `<slot id="foo" name="foo">`. probably i missed one of them in the shadow css file, or maybe they need `display: inline` instead of `display: block`, or some such. 

if you have time Nikki after demos PR would you mind taking a look? otherwise i'll give it a peek tomorrow

nikkimk:
> the visreg is probably because i replaced things like `<div id="foo"><slot name="foo">` with `<slot id="foo" name="foo">`. probably i missed one of them in the shadow css file, or maybe they need `display: inline` instead of `display: block`, or some such.
> 
> if you have time Nikki after demos PR would you mind taking a look? otherwise i'll give it a peek tomorrow

@bennypowers Yeah. Those divs were necessary to get the layout needed. I tried to do it in fewer when I first set it up. I can do it but it will most likely entail putting them back in.

bennypowers:
Slot has display: contents by default which means that we far as CSS is concerned, is as if the shed content is directly there in the div

So if you remove the div and apply all it's properties to the slot (including default ones) it should just work

The exception would be if you're already setting a display property on the slot

elements/rh-tile/rh-tile.ts
```diff
@@ -141,6 +141,7 @@ export class RhTile extends LitElement {
   override async willUpdate(changed: PropertyValues<this>) {
     this.#internals.role = this.radioGroup ? 'radio' : this.checkable ? 'checkbox' : null;
     this.#internals.ariaChecked = !this.#isCheckable ? null : String(!!this.checked);
+    this.#internals.ariaLabel = !this.#isCheckable ? null : 'headline';
```
bennypowers, line 144: see test output. this line causes the tile to announce as 'headline', whereas previously it announced as 'Tile 1'. I believe the test case below is the correct behaviour, meaning that the element's text content should be the label. I will revert this line but preserve the rest of the fixes from this commit, and please review: am i seeing this clearly?

image

bennypowers: @nikkimk see https://github.com/RedHat-UX/red-hat-design-system/pull/1302/commits/6e68ff0de350510a25b8b3f7cdc0f15f234854eb

nikkimk: @bennypowers according to our Guidelines in docs, the heading slot is what should be used: https://ux.redhat.com/elements/tile/guidelines/#heading

bennypowers: Ok, we can adjust the code which computes the label to only consider elements slotted into the headline slot. 

but what if there's no headline text, we still need the accessible-label escape hatch, right?

nikkimk: Let's confirm with @marionnegp, but if I'm not mistaken selectable tiles should always have a headline. 

https://xd.adobe.com/view/925ed24c-3ddf-4111-bacf-2a42e5249f7a-c70d/screen/16207897-40cc-4e49-b064-2940e769b11b

marionnegp: Yes, the headline should be the text for the option you're selecting.

bennypowers: It's it possible to have a selectable tile without a headline?

marionnegp: Selectable tiles aren't supposed to have images or icons, so they should always have a text headline.

bennypowers:
What's left to do here? I lost track...

markcaron:
@nikkimk can you advise on where we're at?

elements/rh-tile/test/rh-tile-group.spec.ts
```diff
@@ -72,84 +33,181 @@ describe('<rh-tile-group>', function() {
     });
   });
 
-  describe('radio tile', function() {
+  describe('as a radio group', async function() {
     beforeEach(async function() {
-      element = await createFixture<RhTileGroup>(tileGroup);
+      element = await createFixture<RhTileGroup>(html`
+        <rh-tile-group radio>
+          <rh-tile>Tile 1</rh-tile>
```
nikkimk, line 40: Tiles should always have a headline.

```
<rh-tile><h1 slot="headline"><a href="#">Tile 1</a></h1></rh-tile>
<rh-tile><h1 slot="headline"><a href="#">Tile 2</a></h1></rh-tile>
<rh-tile><h1 slot="headline"><a href="#">Tile 3</a></h1></rh-tile>
```

nikkimk (changes requested):

elements/rh-tile/test/rh-tile-group.spec.ts
nikkimk: @bennypowers I see you made changes. lmk when you're ready for me to tal

nikkimk (approved):
Livestock grazing tomorrow’s manure

PR #1314, opened 2023-10-27
docs: fix back-to-top blocking right aligned buttons and links

zeroedin:
## What I did

1. Fixed back-to-top link blocking right aligned buttons and links

## Testing Instructions

1.


nikkimk (approved):
Leftover gum; table messy

PR #1315, opened 2023-10-30
fix(card): vertical alignment of footer

bennypowers:
## What I did

1. make the card's footer area always align to the bottom of the card 
2. add a demo which links to the relevant guidelines

## Testing Instructions

1. check in various conditions (with proxy)

### DIFF
--- a/.changeset/card-footer-layout.md
+++ b/.changeset/card-footer-layout.md
@@ -1,0 +1,4 @@
+---
+"@rhds/elements": patch
+---
+`<rh-card>`: ensure footer is always on the bottom of the card
--- a/elements/rh-card/demo/grid.html
+++ b/elements/rh-card/demo/grid.html
@@ -1,0 +1,47 @@
+<div id="card-grid">
+  <rh-card>
+    <h2 slot="header">Grid Card</h2>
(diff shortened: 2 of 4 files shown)


marionnegp (changes requested):
The footer does look like it's anchoring to the bottom. This might be a dumb question, but how would I test it in different contexts "with proxy"?

The spacing between each section is off. The header's padding is correct, but the body and footer's `padding-block` should be changed based on the image below.
Screenshot 2023-11-01 at 11 32 03 AM

bennypowers:
image
you can see the cta is 32px from the bottom of the footer (33 in the screenie because there's 1px between the edge of the cta and the `a`, and visbug had trouble picking up on the cta)

@marionnegp do you mean that the footer content should appear further up the card? if so then let's close this PR, which is about aligning the footer content to the bottom of the card in a grid, regardless of the specific content height of each card

marionnegp:
Added a new issue to fix `rh-card` padding: #1324

bennypowers:
that one demo file (grid.html) is the only one you need to check here, i think

So if this is good, let's approve-and-merge

marionnegp (approved):

PR #1317, opened 2023-10-30
fix(lib): picker scope

bennypowers:
## What I did

1. fix context picker so that it's change event stops at its target.
2. add `npm run serve` command to spin up 11ty dev serve only (without wds)
## Testing Instructions

1. check this out on the context branch in the cards demo on that page. changing the nested context shouldnt' change the parent context.


zeroedin (commented):
Not specific to the changes in this PR but recognized during the review and noted differences in appearance for picker in general across browsers.  Firefox having the best support and I believe the intended look and feel.  Safari being unusable. Chrome in between.  Will roll this out into a separate issue. 
image

zeroedin (approved):
Nice to see being able to launch to the documentation again separately from dev server.

Lemmings Guaranteed To Migrate

PR #1320, opened 2023-10-31
fix(accordion): Remove BaseAccordion dependency

bennypowers:
## What I did

1. remove `heading-tag` and `heading-text` from accordion header
2. only generate shadow header if one is not slotted in
3. use heading level controller for that
4. remove BaseAccordionHeader
5. make AccordionHeaderChangeEvent a cancellable Event

## Testing Instructions

1. the usual

### DIFF
--- a/.changeset/heading-levels-opts.md
+++ b/.changeset/heading-levels-opts.md
@@ -1,0 +1,4 @@
+---
+"@rhds/elements": minor
+---
+`HeadingLevelController`: add `forceWrap` option to the wrap function, narrow the type of the `level` field.
--- a/elements/rh-accordion/rh-accordion-header.css
+++ b/elements/rh-accordion/rh-accordion-header.css
@@ -1,3 +1,41 @@
+button,
+a {
+  cursor: pointer;
+}
+
+.toggle,
+.toggle:before,
+.toggle:after {
+  padding: 0;
+  margin: 0;
+}
+
+.toggle {
+  position: relative;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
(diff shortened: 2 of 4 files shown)


bennypowers:
@zeroedin some preliminary work ahead of the controller stuff from pfe

bennypowers:
@zeroedin what are the chances on this one? would we rather reimplemlent for now, then circle back if-and-when there's a controller available?

zeroedin:
> @zeroedin what are the chances on this one? would we rather reimplemlent for now, then circle back if-and-when there's a controller available?

I think the reimplementation is fine for now, we dont have a controller that is abstract enough.  I'll give this a once over as is.

markcaron:
@bennypowers let's do this in Charmander

markcaron:
@bennypowers isn't this getting resolved in Charmander? Do we need to close this PR?

zeroedin:
> @bennypowers isn't this getting resolved in Charmander? Do we need to close this PR?
> 
> CC: @zeroedin

This PR needs to be resolved/merged before PF4.0 is brought in as a dependency as it removes the BaseAccordion class.  I'll take a look at it tomorrow and see if I can't get it over the hump.

zeroedin:
Closing this PR as stale and outdated.  `<rh-accordion>` already doesn't require `BaseAccordion` from pfe.

Example `RhAccordion extends LitElement` on staging/charmander
https://github.com/RedHat-UX/red-hat-design-system/blob/68b860257dcbd3b483b894092c1a2a17e6365bd9/elements/rh-accordion/rh-accordion.ts#L61

PR #1298, opened 2023-10-24
docs: theming

bennypowers:
## What I did

1. add color context page, with @zeroedin 

## Notes to reviewers
I arbitrarily chose the url /foundations/color/context for this page. if that changes in review for this pr, make sure we also update the reference in rh-spinner.ts (see #1280 )

this PR relies on #1291 

----

## UPDATE Nov 9 2023:

Talking this over with Mr Dr @coreyvickery

    New top-level nav item between getting starter and tokens called Theming
    sub items:
        overview (short prose, graphics, nav links)
        Palettes and Themes
        Customizing
        (maybe) advanced techniques

The palettes and themes page would be the one in this PR as of commit [aef61ce](https://github.com/RedHat-UX/red-hat-design-system/pull/1298/commits/aef61ce52c18d646a678dbe5cd6902acc738b5dd) which describes the color context system

"Color Context" would no longer be a 1st-class subject. Instead, we talk to our users about themes and palettes, and the word "context" is only used in passing in the body text to describe the technical vehicle or 'delivery system' for carrying out theming. We can then link to the wiki or later a separate ux dot page with more tech specs on the context system

WHY? we've identified that the word "color context" is confusing and that even after months of advocacy and docs efforst, it's still not clear to people what this thing is or how to use it. By contrast, most designers and developers know what a theme is, or at least have a working definition, and we can piggy-back on that to clarify and sharpen the concept of what theming means in RHDS.

PR #1321, opened 2023-11-03
docs: redo Color page

marionnegp:
## What I did

1.

## Testing Instructions

1. Check [DP](https://deploy-preview-1321--red-hat-design-system.netlify.app/foundations/color/)

## Notes to Reviewers

@coreyvickery, some changes while you were out:
- On the "Theming" subpage, I changed "Custom colors" to "Custom theme" because Ryan said Brand will reject requests for new colors. I figured custom theme is what you meant.
- I updated the content under "Usage" > "Surface". The background of a button isn't really considered a surface, so the paragraph had to be reworded a bit.
- I also replaced "component" with "element and pattern" wherever it appear, except for the "Graphical objects and UI components" section.


marionnegp:
@coreyvickery, do you know what "contact the Brand team" should link to?

@zeroedin or @bennypowers, can `<rh-tag>` be added to the template? I tried adding it to `layout-foundations.njk` frontmatter and using `{% tag %}`, but that caused an error. Also, does it matter whether I use HTML or markdown to create tables? Both are used on UX dot.

coreyvickery:
@marionnegp Not sure, Ryan or Libby would know.

coreyvickery:
@marionnegp Where did the information in the second bullet come from?

coreyvickery:
@marionnegp Looks great, some small edits.

### Overview
- [ ] Under the `Primitive and semantic colors` heading, bold the word `primitives`

### Theming
- [ ] The light theme image looks a bit blurry when compared to the dark theme image. If there's no fix, don't worry about it

### Usage
- [ ] Under the `Best practices` heading under `Custom themes`, only the words `Design system` should be linked

### Palette
- [ ] Ensure all columns in all tables are the same width

bennypowers (approved):
I updated the markup and some of the content (e.g. phrasing content between headers)

I think the color palette tables should be *removed* and replaced with token tables, or better yet links to the token pages. The palette page should contain directions on how to combine token colours in design projects.

bennypowers:
@nikkimk if you have time can you go over the a11y markup here? can we replace the refs with shared reference links from an include? am i being too precious about that?

bennypowers:
I think the theming content should move to #1298

bennypowers:
@marionnegp if you like i can merge #1298 into this PR and we'll continue here

coreyvickery:
> @marionnegp if you like i can merge #1298 into this PR and we'll continue here

@bennypowers I would prefer the theme stuff stay in #1298 and the PRs stay separate. I guess we can remove any mention of theming from the Color pages until the new Theming section is done.

bennypowers:
thats fine, but the theming content here is *good*, and #1298 has lots of stub pages without content, sooo....

docs/foundations/color/theming.md
```diff
@@ -0,0 +1,136 @@
...
+            state="warning" %}
+Do not apply the Red Hat red color to text in dark environments unless
+it meets [WCAG 2.1 AA][wcag-2-1-aa] requirements.
+{% endalert %}
+
+{% example palette="darkest",
+          alt="examples of several elements against a dark gray surface",
+          src="/assets/color/dark-theme.png" %}
+
+### Desaturated theme {% tag color="purple" %}Beta{%endtag%}
```
bennypowers, line 52: the word `desaturated` is new and doesn't match with the code, except in the case of tile.

is desaturated a color palette? is it a variant on any of the existing color palettes?

markcaron:
I'd like to see a few changes in the way we present images in docs, especially when it's a `do this` / `don't do this` image.

1. Separate out the `do this` / `don't do this` examples into separate images, rather than combining them into a single image. When combined, it's difficult to write `alt` text as well as provide caption/supporting text.
2. Make the green check and red alert icons a separate `tag` and pair it with HTML text, like we do on the Brand site. See image below.
3. Provide caption and supporting text with every image. This helps provide localized context for each example. If we need to, we can use `rh-card` or a simple column/grid structure. See image below.

image

Second, I think we could use a lot less vertical space and create more interesting rhythm through these docs if we create the occasional column structures. One place in particular where we present the color palette tokens. Using a 100% width table creates a lot of white space between the swatch, the token and the hex code—which is a bit of an accessibility issue since the columns can seem disassociated (due to the white space). I think we can present the color tables as 50% width columns (on medium / large screens), so we get a closer pairing of text and swatch as well as a pleasant visual of color gradients side-by-side.

For example:
image

coreyvickery (approved):

PR #1323, opened 2023-11-08
fix(tooltip): fixing content display issues before upgraded

brianferry:
Issue: On [Red Hat Product Trials](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux/server/trial) and on **mobile** when the tooltip is on the page but not yet upgraded by @floating-ui (ie not yet interacted with to get the coordinates) the content ends up causing the screen to scroll too far to the right and creates a layout error like below:

image

To fix this,

## What I did

1.  Added an `initialized` private property to tooltip. 
2. Updating `initialized` to true when `show` function is called.
3. CSS is set to `display: none` on the tooltip content until the show function is called, then it is set to show permanently.

### DIFF
--- a/.changeset/healthy-onions-work.md
+++ b/.changeset/healthy-onions-work.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+rh-tooltip: fixing issue where content would take up unwanted space before floating-ui initialization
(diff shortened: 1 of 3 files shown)


.changeset/healthy-onions-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+rh-tooltip: fixing issue where content would take up unwanted space before floating-ui initialization
```
eyevana, line 5:
```suggestion
`rh-tooltip`: fixes issue where content would take up unwanted space before floating-ui initialization
```

eyevana (approved):
Minor suggestion on changeset, otherwise LGTM! :)

PR #1326, opened 2023-11-09
fix(card): collapse margins

### LINKED ISSUE #1324: [bug] <rh-card> padding
### Describe the bug

The block padding for card's body and footer needs to be adjusted. Right now the space between the body and footer is 64px because of padding in both sections. The space between the body and footer should be 24px (`--rh-space-xl`).

Additionally, can the padding-inline of header be changed to `--rh-space-2xl`? It'll help the content in the header align with the rest of the content.

### Which Component?

rh-card

### Steps to reproduce

1. Pull up a demo of card.
2. Check to see if padding matches screenshots below.

### Expected behaviour

The space between the body and footer should be 24px (--rh-space-2xl). Inline padding of header should change to 32px (--rh-space-2xl).

### Screenshots

**Current padding/spacing:**

Screenshot 2023-11-08 at 9 26 29 AM

Screenshot 2023-11-08 at 9 26 35 AM

**Expected spacing:**
 
Screenshot 2023-11-01 at 11 32 03 AM

### Element HTML

_No response_

### Operating System (OS)

macOS

### Browser

_No response_

### More browser information

_No response_

### Additional Info

_No response_

bennypowers:
Closes #1324

## What I did

1. replace padding with margin in card
2. collapse margins for body, header, footer
3. update grid demo

## Testing Instructions

1. confirm that the grid demo cases are comprehensive
2. confirm that header and body *inline* margins are correct
3. confirm that header, body, and footer *block* margins correctly collapse
4. repeat all the above on mobile


marionnegp (changes requested):
In the grid demo, @brianferry noticed that there's a 2px padding on the footer, but I don't see the same thing on the single card demo. Any idea why that 2px is getting added?

**Grid demo**
Screenshot 2023-11-14 at 12 49 01 PM

**Card demo**
Screenshot 2023-11-14 at 12 49 20 PM

Spacing looks correct otherwise.

bennypowers:
that's the margin-block-start auto. it's intentional in order to bring that particular footer in alignment with it's row. the `2px` value isn't specific, it was calculated by the browser engine. remove content from the body of that card to see the 2px disappear

bennypowers:
i.e. this is a feature not a bug
image

marionnegp:
Yeah, I see that if I remove the content or make a different card have more text, the tallest card takes on that 2px margin. Sorry if this is a dumb question, but why do the cards in this page's "Recommended products" section not use `margin-block-start`? Is `margin-block-start` the only way to keep the footer at the bottom? I know that calculated margin isn't a static value and is pretty negligible, but I still want to reduce any confusion between what's expected based on the design.

bennypowers:
@marionnegp i didn't get the question, which cards specifically did you mean?

marionnegp:
Oh man, I knew that I'd forget to link it. I was talking about the three cards at the top of the [All Red Hat Products page](https://www.redhat.com/en/technologies/all-products).

bennypowers:
because they're not rh-cards, they're custom

marionnegp (approved):
Since the additional margin is pretty negligible...Lentils Get Too Mushy.

We may want to consider adding a note to `<rh-card>`'s Code subpage though.

PR #1330, opened 2023-11-09
fix(cta): move componenent level css props to private vars

zeroedin:
## What I did

1. Fixed component level CSS vars `--rh-cta` with private prefixed`--_` variables allowing `--rh-cta` scope to be used to set values in the component.  This matches our wiki best practice on [defining CSS variables](https://github.com/RedHat-UX/red-hat-design-system/wiki/CSS-Styles#defining-css-variables)

## Testing Instructions

1.

### DIFF
--- a/elements/rh-cta/rh-cta.css
+++ b/elements/rh-cta/rh-cta.css
@@ -16,7 +16,7 @@
   font-size: inherit !important;
   font-weight: inherit !important;
   line-height: inherit !important;
-  text-decoration: var(--rh-cta-text-decoration) !important;
+  text-decoration: var(--_text-decoration) !important;
   z-index: 2 !important;
 }
 
@@ -40,17 +40,17 @@
 #container {
   position: relative;
   white-space: nowrap;
-  color: var(--rh-cta-color);
+  color: var(--_color);
   font-family: var(--rh-font-family-heading, RedHatDisplay, "Red Hat Display", "Noto Sans Arabic", "Noto Sans Hebrew", "Noto Sans JP", "Noto Sans KR", "Noto Sans Malayalam", "Noto Sans SC", "Noto Sans TC", "Noto Sans Thai", Helvetica, Arial, sans-serif);
   font-size: var(--rh-font-size-text-xl, 1.125rem);
   font-weight: var(--rh-font-weight-bold, 600);
   line-height: var(--rh-line-height-lg, 1.5);
-  background-color: var(--rh-cta-background-color);
-  border-color: var(--rh-cta-border-color, transparent);
+  background-color: var(--_background-color);
+  border-color: var(--_border-color, transparent);
   border-radius: var(--rh-border-radius-default, 3px);
   border-width: var(--rh-border-width-sm, 1px);
 
-  --_context-background-color: var(--rh-cta-background-color) !important;
+  --_context-background-color: var(--_background-color) !important;
   --_arrow-size: 13px;
   --_arrow-plus-padding: calc(var(--rh-space-md, 8px) + var(--_arrow-size));
   --_pf-icon-plus-padding: calc(5px + var(--pf-icon--size));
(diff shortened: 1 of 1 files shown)


bennypowers:
this is breaking, requires a semver major

elements/rh-cta/rh-cta.css
```diff
@@ -209,43 +212,43 @@ svg {
...
-  --rh-cta-hover-background-color: transparent;
-  --rh-cta-hover-border-color: transparent;
-  --rh-cta-hover-inner-border-color: transparent;
-  --rh-cta-hover-color: var(--rh-color-interactive-blue-darkest, #004080);
-  --rh-cta-hover-text-decoration: none;
-  --rh-cta-focus-background-color: transparent;
+  --_background-color: var(--rh-cta-background-color,transparent);
+  --_border-color: var(--rh-cta-border-color, transparent);
+  --_color: var(--rh-cta-color, var(--rh-color-interactive-blue, #0066cc));
+  --_hover-background-color: var(--rh-cta-hover-background-colo, transparent);
```
nikkimk, line 218: --_hover-background-color: var(--rh-cta-hover-background-color, transparent);

elements/rh-cta/rh-cta.css
```diff
@@ -115,33 +115,33 @@ svg {
...
 :host(:hover) #container {
-  color: var(--rh-cta-hover-color);
-  border-color: var(--rh-cta-hover-border-color);
-  background-color: var(--rh-cta-hover-background-color);
+  color: var(--_hover-color);
+  border-color: var(--_hover-border-color);
+  background-color: var(--_hover-background-color);
 
-  --rh-cta-text-decoration: var(--rh-cta-hover-text-decoration);
+  --_text-decoration: var(--rh-cta-hover-text-decoration var(--_hover-text-decoration));
```
nikkimk, line 144: --_text-decoration: var(--rh-cta-hover-text-decoration, var(--_hover-text-decoration));

nikkimk (dismissed):

elements/rh-cta/rh-cta.css
zeroedin: all that copy and paste was bound to be a typo ;)

elements/rh-cta/rh-cta.css
zeroedin: nice catch

bennypowers (changes requested):
since this is a breaking change (it would require us to release `@rhds/elements@2.0.0`), we have to consider well before merging.

markcaron:
This is going out in Chansey, which will be a major version bump (to 2.0), so let's get this done.

bennypowers:
Let's move on this for Charmander

bennypowers:
@zeroedin would you mind copying in https://codepen.io/bennyp/pen/abrqXVG and see to what degree this pr helps with theming? what would it be like to theme a cta on a variety of color-palettes, using only css vars?

zeroedin:
> @zeroedin would you mind copying in https://codepen.io/bennyp/pen/abrqXVG and see to what degree this pr helps with theming? what would it be like to theme a cta on a variety of color-palettes, using only css vars?

Good call, I have a feeling this will open that up to easier customization.

zeroedin:
@bennypowers PTAL desaturated demo: https://deploy-preview-1330--red-hat-design-system.netlify.app/elements/call-to-action/demo/desaturated/

elements/rh-cta/demo/desaturated.html
bennypowers, line 1: i want the css for this file to look like this:

```css
.theme-mauve {
  --rh-whatever-on-light: foo;
  --rh-whatever-on-dark: bar;
  /* etc */
}
```
`[color-palette]` attribute selectors are a smell and shouldn't be required for any theming operation

zeroedin: if we avoid `[color-palette]` from the lightdom custom theme CSS, we need many more CSS props in the element with light|dark context awareness.

```
[data-theme="awesome-purple"] [color-palette="darkest"] rh-cta[variant="primary'] {
  --rh-cta-background-color: lilac;
  --rh-cta-color: var(--rh-color-text-primary-on-light, #151515);
  ...
}
```
vs
```
[data-theme="awesome-purple"]  rh-cta[variant="primary'] {
  --rh-cta-background-color-on-light: rebeccapurple;
  --rh-cta-background-color-on-dark: lilac;
  --rh-cta-color-on-light:  var(--rh-color-text-primary-on-light, #151515);
  --rh-cta-color-on-dark:  var(--rh-color-text-primary-on-dark, #ffffff);
  ...
}
```

If that is the direction we want  I'll update the CSS to do just that.

zeroedin: desaturated demo remove will tackle with themeing

zeroedin:
This PR does not remove CSS properties, it only makes them accessible as listed in the code tab.  This PR is much like https://github.com/RedHat-UX/red-hat-design-system/pull/1685 for tile fixed order token.

markcaron (approved):
Laying Groundwork; Tailor-Made

PR #1348, opened 2023-11-16
style: lint docs styles

bennypowers:
## What I did

1. tweak stylelint config
2. lint all the things

## Testing Instructions

1. have a stroll through the dp, make sure nothing's broken


marionnegp (changes requested):
If an image uses the `example--palette-none`, it sometimes gets centered, but I'm unsure why it changed in this DP. Here's an example from the [`<rh-accordion`>'s Guidelines](https://deploy-preview-1348--red-hat-design-system.netlify.app/elements/accordion/guidelines/):
Screenshot 2023-12-05 at 12 11 43 PM

Sticky card's sample pattern image is doing the same thing.
Screenshot 2023-12-05 at 12 13 20 PM

Those are the only differences I've noticed.

PR #1350, opened 2023-11-16
docs: samples

bennypowers:
## What I did

1. add `{% sample %}` paired shortcode
2. use it for:
    - card
    - timestamp
3. tinkered with codeblock's css so that it works nice in grids. did i break stuff? you tell me :laughing:


bennypowers:
I don't think this needs to be comprehensive. the only blockers on this should be things like:

- the design needs improvements
- something is broken

but, like .. "we should also do this for xyz element" can be a follow up pr

docs/_plugins/shortcodes/sample.cjs
```diff
@@ -0,0 +1,111 @@
...
+     * @param {string}    [options.code]              vertical layout
+     * @this {EleventyContext}
+     */
+    function example(content, {
+      style = '',
+      colorPalette = 'lightest',
+      stacked = false,
+      picker = false,
+      columns = 1,
+      code = 'show',
```
brianferry, line 36: Should this be a boolean?

elements/rh-accordion/docs/10-style.md
```diff
@@ -1,3 +1,5 @@
+<script type="module">import '/assets/elements/uxdot-code-sample.js';</script>
```
brianferry, line 1: Would it be possible to move this into the sample script?

brianferry: Fair point, I had thought about #1 but like you said the browser should in theory be able to handle that but it doesn't really impact anything other than needing to remember to put that line on a page when the code sample import is being used.

brianferry (approved):
Went through the demo files and the docs and everything looks good to me

zeroedin (changes requested):
These blocks should have a red border and icon like the top with the image

image


image

zeroedin (commented):
General comment non blocker to this PR.  Is the intent to replace all "sample" images with real working components?  If so should we get those components not in this PR added to the backlog to follow up on. /cc @markcaron

zeroedin (changes requested):
https://deploy-preview-1350--red-hat-design-system.netlify.app/elements/accordion/style/#theme

Color Palette isn't clickable here in Chrome 120.0.6099.199:

image

Doesn't display at all in Safari  17.2.1 :

image

And is weirdly cropped in Firefox 115.6.0esr (64-bit)

Screenshot 2024-01-08 at 1 47 10 PM

markcaron:
> General comment non blocker to this PR. Is the intent to replace all "sample" images with real working components? If so should we get those components not in this PR added to the backlog to follow up on. /cc @markcaron

@zeroedin not necessarily _all sample images_, but at least the ones that make the most sense.

bennypowers:
the anatomy sample component is at the MVP level. if that's the only blocker, slap a `hidden` attr on it and let's merge this puppy

markcaron:
@bennypowers honestly, I feel like the Sample Component has gone beyond the original intention.

Don't get me wrong, it's looking pretty cool with all the features, but I think it changes the context of the design sections, like the Style tab. We don't want to replace ALL images, since images still have their use in these sections.

For example, if you're wanting to see all the different Sizes of the Audio Player, it looks "fine" (aside from the last one breaking out of the container):

image

But, on Mobile, you lose the visual since ALL of the Audio Player samples become "mobile friendly":

image

We want the image to show ALL the sizes/options, and not have the visual representation / Sample Component react live to the browser or container width.

_Note: the little purple tooltips are also wonky on mobile, but that's a separate topic_

I feel like the true use of the Sample Component lives on the Code and Overview tabs, where we'd want to show a live example, and not on design sections where we'd want to highlight specific features (that retain their context across devices).

For example, on mobile, the Configuration section is using an image still, and it doesn't lose it's context—you can still see all the config options visually:

image

Hopefully that makes sense.

This isn't a blocker anyway. Maybe we take this out of the Caterpie release, and fast follow with it since it's just docs.

bennypowers:
yup, makes sense to me

zeroedin:
Do we want to hold on `uxdot-anatomy-sample` until we have time to flesh it out more?  I think it would be nice to get the `uxdot-code-sample` component in without that anatomy holding this PR up.

bennypowers:
Superceeded mostly by #1465

Issue #1361, opened 2023-11-27
[bug] "Installation" section's tab panel width


marionnegp:
### Describe the bug

The "Installation" section of any element's "Code" subpage is too wide and is adding horizontal scroll. The tab font size also looks larger than it should be. 

Should this be an update to `<rh-tabs>`?

### Which Component?

rh-tabs

### Steps to reproduce

An example can be found on the [audio player's page](https://ux.redhat.com/elements/audio-player/code/).

### Expected behaviour

The tabs + tab panel should have the same width as the rest of the page's content.

### Screenshots

<img width="1437" alt="Screenshot 2023-11-27 at 11 09 23 AM" src="https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/cbb7db5c-2eaf-4798-b6ff-a978b93bb793">

### Operating System (OS)

macOS

PR #1362, opened 2023-11-27
fix(tabs): tab button font-size default value

zeroedin:
## What I did

1. Added `--rh-tabs-link-font-size` to with default font-size of 1rem

Reference #1361

### TODO:
- [x] Determine if breaking change or not, I can see it either way.    

## Testing Instructions

1.

### DIFF
--- a/.changeset/green-teachers-add.md
+++ b/.changeset/green-teachers-add.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tabs>`: adds default value for tab font-size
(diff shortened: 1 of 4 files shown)


marionnegp (approved):
Levitating Gnomes Travel Merrily

brianferry (approved):
LGTM Generators Trivialize Merges

bennypowers (changes requested):
instead just set rh-tab's font size on rh-tabs, no opostrophe. if slotted specificity is a thing, then fix it in ux-dot

bennypowers (approved):

PR #1373, opened 2023-12-07
feat(table): color context support

### LINKED ISSUE #1371: [feat] RH-Table supports dark theme
### Description

Be able to add an rh-table to a dark background 

### Suggested solution

Have text inside rh-table show white or light gray on a dark background

### Example API

_No response_

### Additional context

Tables added in the WYSIWYG reference rh-table and right now it only works on white/light-gray backgrounds 

image

To test

1. add a section on a RHDC page with a black background and set theme to dark
2. bring in simple text component 
3. inside WYSIWYG add a table 
4. Save the block
5. Text is now black on black

zeroedin:
## What I did

1. Adds support for color context 
2. Adds color context demo

Closes #1371 

### TODO
- [x] Determine table hover colors for the various color palettes.
- [x] Update docs on this PR or spin off to another issue/PR: https://github.com/RedHat-UX/red-hat-design-system/issues/1389
- [x] Changeset

## Testing Instructions

1. View the [color-context demo](https://deploy-preview-1373--red-hat-design-system.netlify.app/elements/table/demo/color-context/)

## Notes to Reviewers

Double check that the color combinations paas WCAG accessibility guidelines

### DIFF
--- a/.changeset/calm-fishes-sleep.md
+++ b/.changeset/calm-fishes-sleep.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-table>`: added support for color context when used as a child of a context provider
--- a/.changeset/calm-fishes-sleep.md
+++ b/.changeset/calm-fishes-sleep.md
@@ -2,4 +2,4 @@
 "@rhds/elements": patch
 ---
 
-`<rh-table>`: added support for color context when used as a child of a context provider
+`<rh-table>`: added color palette and theme support
--- a/elements/rh-table/rh-table-lightdom.css
+++ b/elements/rh-table/rh-table-lightdom.css
@@ -1,10 +1,5 @@
-rh-table table {
-  min-width: 100%;
-  margin: 0 auto;
-  table-layout: fixed;
-  border: 0;
-  border-collapse: collapse;
-
+rh-table,
+rh-table[color-palette="lightest"] {
   --rh-table-row-border:
     var(--rh-border-width-sm, 1px)
     solid
@@ -27,7 +22,22 @@ rh-table table {
       );
 }
 
-rh-table[color-palette^="dark"] table {
+[color-palette="light"] > rh-table,
+rh-table[color-palette="light"],
+[color-palette="lighter"] > rh-table,
+rh-table[color-palette="lighter"] {
+  --rh-table-row-background-color:
+    rgb(
+        var(
+          --_rh-color-white-rgb, /* making private as currently doesn't exist */
+          255 255 255
+        )
+        / var(--rh-opacity-50, 50%)
+      );
+}
+
+[color-palette^="dark"] > rh-table,
+rh-table[color-palette^="dark"] {
   --rh-table-row-background-color:
     rgb(
         var(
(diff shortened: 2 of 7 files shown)


zeroedin:
@coreyvickery need design review sign off on the hover colors used on the [color-context demo](https://deploy-preview-1373--red-hat-design-system.netlify.app/elements/table/demo/color-context/)

hellogreg (approved):
Lamp Genie's Tricky Magic

(Contrast looks good!)

zeroedin:
> Lamp Genie's Tricky Magic

Elite level LGTM right here 😮

.changeset/calm-fishes-sleep.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-table>`: added support for color context when used as a child of a context provider
```
bennypowers, line 5:
```suggestion
`<rh-table>`: added color palette and theme support
```

elements/rh-table/rh-table.ts
```diff
@@ -1,11 +1,16 @@
...
 import { customElement } from 'lit/decorators/custom-element.js';
+import { property } from 'lit/decorators/property.js';
+import { classMap } from 'lit/directives/class-map.js';
 
 import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
 
-import styles from './rh-table.css';
 import { RequestSortEvent, RhSortButton } from './rh-sort-button.js';
 
+import { colorContextConsumer, type ColorTheme } from '../../lib/context/color/consumer.js';
```
bennypowers, line 10: demo and lightdom css (and my suggestion for editing the changelog) have color palettes as well as themes on table, but i'm only seeing theme here

zeroedin: Are you suggesting making table itself a provider too add direct support of `color-palette` attribute? 

@marionnegp and @coreyvickery and I had a discussion where we thought it would be better if table wasn't itself a provider, but just a consumer.  There would need to be a parent element that was providing that context.    Can you elaborate on why you think it should be a provider as well?

zeroedin: after poking around a bit more I think I understand the comment now.  the lightdom css actually wouldn't have worked as intended given a nested structure.  Only way to handle that is to have the component react with something styleable  on the host, hence the need for the color palette.

elements/rh-table/rh-table-lightdom.css
```diff
@@ -12,46 +11,88 @@
...
+}
+
+rh-table[color-palette^="dark"] table {
+  --rh-table-row-background-color:
+    rgb(
+        var(
+          --_rh-color-white-rgb, /* making private as currently doesn't exist */
+          255 255 255
+        )
+        / var(--rh-opacity-10, 10%)
```
marionnegp, line 37: Background color and opacity look correct here, but the Figma and the demo colors look different.

**Dark contexts**
Demo:
Screenshot 2024-01-03 at 4 38 59 PM

Figma:
Screenshot 2024-01-03 at 4 39 11 PM

**Light and lighter contexts**
Demo:
Screenshot 2024-01-03 at 4 56 28 PM

Figma:
Screenshot 2024-01-03 at 4 54 45 PM

Is it possible that the opacity value for the light and lighter contexts is being used for the dark contexts and vice versa? This is what it looks like if the dark theme row hover has a 50% opacity in Figma:
Screenshot 2024-01-03 at 4 58 07 PM

marionnegp (changes requested):

elements/rh-table/rh-table-lightdom.css
zeroedin: This is a bug with the last update I made, thank you for catching.  Will fix

zeroedin: Pushed a fix, unfortunately it has a compromise of the context provider having to be a direct parent OR the table itself has to provide the specific color palette to match when nested.  Going to write up a issues/POC about this problem for discussion.

zeroedin: POC: https://github.com/RedHat-UX/red-hat-design-system/pull/1395

coreyvickery (approved):
Looks great.

marionnegp (approved):

PR #1374, opened 2023-12-08
fix(tabs): incorrect application of inline margin on vertical box tabs

### LINKED ISSUE #1299: [bug]: `<rh-tabs>` margin-line when vertical
tabs should never have a margin-inline (inset) when also vertical.

Fix:
```css
:host(:is([centered]:not([vertical]), [box="inset"]:not([vertical]))) [part="tabs"] {
  margin-inline: var(--rh-tabs-inset, var(--_inset-inline-margin, auto));
}
```

this was exposed by change up to demos that allows these properties to be easily added: https://deploy-preview-1291--red-hat-design-system.netlify.app/elements/tabs/demo/color-context/

zeroedin:
## What I did

1. Removed inset margin from being applied to vertical tabs that have box set to inset.

Closes #1299 

## Testing Instructions

1.

## Notes to Reviewers

To see the difference view the [vertical demo](https://deploy-preview-1374--red-hat-design-system.netlify.app/elements/tabs/demo/vertical/) and add `box="inset"` in the codeview to `rh-tabs`.  Then remove inset leaving only `box` as an attribute, the style should be identical.  Do the same thing on [ux.redhat.com vertical demo](https://ux.redhat.com/elements/tabs/demo/vertical/) and you'll see the bug that is fixed with this PR.

### DIFF
--- a/.changeset/cold-islands-compare.md
+++ b/.changeset/cold-islands-compare.md
@@ -2,4 +2,4 @@
 "@rhds/elements": patch
 ---
 
-`<rh-tabs>`: fixed incorrect application of inline margin on vertical'
+`<rh-tabs>`: fixed layout of non-vertical box variant tabs
--- a/.changeset/cold-islands-compare.md
+++ b/.changeset/cold-islands-compare.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tabs>`: fixed incorrect application of inline margin on vertical'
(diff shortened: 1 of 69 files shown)


.changeset/cold-islands-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tabs>`: fixed incorrect application of inline margin on vertical'
```
bennypowers, line 5:
```suggestion
`<rh-tabs>`: fixed layout of non-vertical box variant tabs
```

bennypowers (approved):

Issue #1353, opened 2023-11-16
[bug] `<rh-tab-panel>` has color provider decorator however not implemented


zeroedin:
`<rh-tab-panel>` should only be a consumer.  There is no current design spec that calls for a panel to provide its own context.  

That said there is precedent in `<rh-accordion>` where a panel provides it's own context:
<img width="707" alt="image" src="https://github.com/RedHat-UX/red-hat-design-system/assets/13666/4a8f86fb-2694-451c-8f7e-dc627da5b4ef">

However maybe doing the same to a tab panel might cause the panel to feel detached from the tab that activated it when the panels color context is different?  Accordion is more self contained so the relationship still exists visually.

I believe the addition of the decorator was a mistake and we should remove the color context provider decorator.  As it stands there is no current style implementation for panel to switch even if `color-palette` is set and I don't believe doing so would be a breaking change.

@coreyvickery @marionnegp @bennypowers thoughts?

marionnegp:
@zeroedin, my gut says to keep `<rh-tab-panel>` a consumer only, at least until someone comes to us with a use case that requires it to be a provider.

I think we could get away with `<rh-tab-panel>` providing its own context when using open tabs. With box tabs, we'd want the active tab color to match the tab panel. Here's a screenshot of some (very shoddy) mockups just to see what it would look like if someone added a pink background to the panel.

<img width="941" alt="Screenshot 2023-11-16 at 4 24 11 PM" src="https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/d587d50d-fec7-433e-ba7e-6771f90c76bc">

zeroedin:
> 
> ```
> <rh-surface color-palette="darkest">
> 
> <rh-tabs>
>   <rh-tab slot="tab">Consumer</rh-tab>
>   <rh-tab-panel>... normal ...</rh-tab-panel>
> 
>   <rh-tab slot="tab">Override</rh-tab>
>   <rh-tab-panel id="override">
>     <rh-surface color-palette="darkest"> ... overridden ... </rh-surface>
>   <rh-tab-panel>
> </rh-tabs>
> 
> </rh-surface>
> ```
> 
> ```css
> #override {
>   padding: 0;
>   & > rh-surface {
>     padding: var(--rh-space-2xl, 32px);
>   }
> }
> ```


Just for clarity, did you mean to override with a new value for `color-palette` in the above example?  

[...]

PR #1376, opened 2023-12-08
fix(tabs): remove color provider from `rh-tab-panel`

zeroedin:
## What I did

1. removed `color-palette` attribute

## Testing Instructions

1.

## Notes to Reviewers

`color-palette` in the case of panels was never wired up.  Using it would not provide any change to the component.  My assumption is is a minor but maybe we can get away with a patch change?  Discussion on this is in the issue [#1353 ](https://github.com/RedHat-UX/red-hat-design-system/issues/1353) In the future we'd suggest using `<rh-surface>` inside of the panel.

### DIFF
--- a/.changeset/itchy-dogs-matter.md
+++ b/.changeset/itchy-dogs-matter.md
@@ -1,5 +1,5 @@
 ---
-"@rhds/elements": minor
+"@rhds/elements": patch
 ---
 
-`<rh-tab-panel>`: removed color-palette attribute`
+`<rh-tab-panel>`: removed unused color-palette attribute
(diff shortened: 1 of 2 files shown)


.changeset/itchy-dogs-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-tab-panel>`: removed color-palette attribute`
```
bennypowers, line 5:
```suggestion
`<rh-tab-panel>`: removed unused color-palette attribute
```

.changeset/itchy-dogs-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": minor
```
bennypowers, line 2:
```suggestion
"@rhds/elements": patch
```

bennypowers (approved):

PR #1380, opened 2023-12-13
fix(card): layout for first slotted p

bennypowers:
## What I did

1. fix various issues with rh-card (see changeset)
2. cherry-picked those fixes from the #1350 for ease of review

### DIFF
--- a/.changeset/card-first-p.md
+++ b/.changeset/card-first-p.md
@@ -1,0 +1,4 @@
+---
+"@rhds/elements": patch
+---
+`<rh-card>`: improved styles for first body paragraph when there is other content displayed first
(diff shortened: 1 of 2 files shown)


zeroedin (approved):
There are 2 issues I noticed while reviewing this PR and confirmed in a discussion with @marionnegp.  

This change works when the body (anonymous slot) is the :first-child.  However when `[slot="header]` is present you get a 48px gap between the header and the first `<p>`.   Per the design spec we should only have 32px here.  However we have 32px margin on #body + 16px margin on the `<p>`.

For the sake of getting this PR out the door, we are going to move that to a new issue as it is additional scope to what this is trying to fix which is the version where header slot is empty.

The second issue that is also going to be spun off to its own issue, is that when a header slot is `.empty` the class is overwritten by `#header { display: flex }` and `display:none` from the class does not get applied, needs an `!important`  added.

PR #1381, opened 2023-12-13
fix(code-block): allow slotted prism-highlighted code

bennypowers:
## What I did

1. fix various issues with rh-code-block (see changeset)
2. cherry-picked those fixes from the #1350 for ease of review


brianferry (approved):
Ran tests locally, looks good

PR #1386, opened 2023-12-15
docs: variants demo not loading images in live section of docs page

brianferry:
## What I did

1.  Replaced the images in the `rh-card` and `rh-avatar` with static images until we can figure out why they're not being loaded in the service worker.


elements/rh-card/demo/variants.html
```diff
@@ -77,8 +77,8 @@ <h2 slot="header">End aligned content, footer</h2>
 
 <rh-card class="full">
   <img slot="header"
-       alt="a placeholder header image"
-       src="https://fakeimg.pl/400x250"
+       alt="a kitten's cute little face"
+       src="https://ux.redhat.com/elements/card/demo/kitten-400x250.jpeg"
```
zeroedin, line 81: What if we got rid of the image all together and used an SVG placeholder (idea and svg code stolen from @bennypowers sample PR). I know I no more cute kittens 😿 

replace the `...` with the following svg:

```html
<svg slot="header" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 120">
  <title>A placeholder image in a card header slot</title>
  <rect x="0" y="0" width="320" height="120" fill="#e0e0e0"/>
  <line x1="0" y1="0" x2="320" y2="120" stroke="#d0d0d0"/>
  <line x1="320" y1="0" x2="0" y2="120" stroke="#d0d0d0"/>
</svg> 
```
Which looks like this:
image

brianferry: @zeroedin thanks!  Went ahead and updated now

zeroedin (approved):
Launch To Grand Majesty

PR #1392, opened 2024-01-04
fix(dialog): remove pfe dependency

bennypowers:
## What I did

1. remove `PfModal`

### DIFF
--- a/elements/rh-dialog/rh-dialog.css
+++ b/elements/rh-dialog/rh-dialog.css
@@ -1,3 +1,190 @@
+:host {
+  display: block;
+  position: relative;
+  --_spacer-align-top: var(--pf-c-modal-box--m-align-top--spacer,
+    var(--pf-global--spacer--sm, 0.5rem));
+  --_height-offset: min(var(--_spacer-align-top), var(--pf-global--spacer--2xl, 3rem));
+}
+
+[hidden] {
+  display: none !important;
+}
+
+section {
+  display: flex;
+  position: fixed;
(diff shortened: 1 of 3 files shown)


elements/rh-dialog/rh-dialog.css
```diff
@@ -1,36 +1,151 @@
...
+
+header ::slotted(:is(h1,h2,h3,h4,h5,h6)[slot="header"]) {
+  font-size: var(--rh-font-size-heading-sm, 1.5rem);
+  font-weight: var(--rh-font-weight-body-text-regular, 400);
+  font-family: "Red Hat Display", RedHatDisplay, Overpass, Helvetica, sans-serif;
+}
+
 [part="close-button"] {
-  color: var(--rh-color-icon-subtle, #707070);
+  background-color: transparent;
```
marionnegp, line 101: Changing the icon color from `--rh-color-icon-subtle` to `transparent` removes the hover effect for the close button and makes it black instead of dark gray. What's the reason for changing it? Should the icon color token be added somewhere else?

marionnegp (changes requested):

brianferry (approved):
Limitation Gzip Throws Making

elements/rh-dialog/rh-dialog.css
bennypowers: it's a mistake on my part. thanks for catching it

marionnegp (approved):

Issue #1319, opened 2023-10-31
`<rh-accordion>`: remove BaseAccordion Dependency


bennypowers:

markcaron:
https://github.com/RedHat-UX/red-hat-design-system/pull/1394

PR #1394, opened 2024-01-04
feat(accordion): remove BaseAccordion import from PFE

brianferry:
## What I did

1.  Moved all the code in for `BaseAccordion` (that I could), `BaseAccordionHeader`, and `BaseAccordionPanel`.  

I was able to remove all the need for the code for Header and Panels base classes, however the Header needs to be able to see if what called it was an Accordion component (and if so, send an event) and if I moved that code out of a BaseClass we get a circular dependency.  Went back and forth on it but currently that's the only function still remaining in the BaseAccordion and it's so the header can check if it's a `BaseAccordion` class instead of an `RhAccordion` class.  

## Testing Instructions

1.  Open up the demos for `RhAccordion` and verify that there are no regressions in the code.  Everything should work normally.


nikkimk (approved):
Lopsided geometry. Terrible mathematician.

markcaron:
Closes #1319

PR #1395, opened 2024-01-04
[POC]: color-palettes as themes

zeroedin:
# DO NOT MERGE - Proof of concept

## What I did

1.   Explored what if all `color-palette` values ~~were available as~~ **replaced the idea of** `ColorTheme` in a consumer?
2.  Removed the concept of ColorTheme

Background:  I ran into an issue with `rh-table` where I needed to access the parent set context.  However I needed the explicit value not just `light | dark` which is accessible from the consumer.  I was curious on how much of a change it would be if we actual told the consumer what color palette they were on not just the representation of `light and dark` as currently is accessible in `ColorTheme`.

I exposed this issue when I noticed that I would have trouble in LightDOM CSS to get the correct transparency hover effects for table, and ended up adding a compromise where the LightDOM CSS now expects that `rh-table` is either a DIRECT child of the parent provider OR that you'd have to add the `[color-palette]` attribute to `rh-table` AND  it would also needs to be a provider not just a consumer only.  I didn't find this solution satisfactory.  So here we are.

I will split this background off to a discussion thread tomorrow with a more thoughtful lede, but wanted to push this code up so others can take a look.  

### TODO: 

- [x]  Fix tests where necessary 

## Testing Instructions

1.


bennypowers:
interesting idea :nerd_face: . it would certainly simplify things to only have one type (i.e. either `theme` or `color-palette`, but not both). if we went this way would it be better to remove one of the unions rather than holding two identical lists? Under this approach can we still distinguish between themes (default white and red, pink, whatsoever) and palettes (`(light|dark)(er|est)?`) versions of each of those. 

a concrete example: say we wanted to do a pink-themed landing page a la the keep your options open campaign, and on that page we wanted to default to the light palette, but there should also be a band with a darkest palette (provider) which provides a dark theme. How would that scenario play out in this PR?

zeroedin:
> pink-themed landing page
> light palette
> also be a band with a darkest palette (provider) which provides a dark theme

I'm not sure currently there is any difference with my changes from what we would do right now.   The only difference as I understand is that I'm allowing the component to know the full set so I can key off the specific value.

That said.  A thought I had was also extending this maybe one step further.  What if `ColorTheme` aka `on` actually returned both "palette"  `darkest -> lightest + saturated | desaturated` and "theme": `light | dark` as a key => value pair. ie:
```
{
  'darkest': 'dark',
  'darker': 'dark',
  'dark': 'dark',
  'light': 'light',
  'lighter': 'light',
  'lightest': 'light',
  'saturated': 'dark',
  'desaturated': 'light'
 }
 ```
 
With themes it is my understand we are saying for `dark` we need `light` variants of text color, borders etc.  However in the case of `rh-table` for `lightest` color-palette parent context I actually needed needed something more then just the `light` theme class.  I needed to know specifically it was `lightest` as the hover effect needed slightly modified versus `light` and `lighter`.

File: lib/context/color/provider.ts [hunk 1]
```diff
@@ -43,18 +40,18 @@ export class ColorContextProvider<
...
-    darker: 'dark' as const,
+    darkest: 'darkest' as const,
+    darker: 'darker' as const,
     dark: 'dark' as const,
     light: 'light' as const,
-    lighter: 'light' as const,
-    lightest: 'light' as const,
+    lighter: 'lighter' as const,
+    lightest: 'lightest' as const,
   }));
```
zeroedin, line 49: TODO: rewrite static contexts to use `ColorPalette`  to create a new Map??

zeroedin: Resolved by using the direct `color-palette` values

markcaron:
@bennypowers @zeroedin @adamjohnson I wanna do this. When can we make this a thing?

zeroedin:
Given latest changes to theme tokens and what context consumer/provider provides now I believe we can now close out this POC.

PR #1399, opened 2024-01-05
fix: token migration errors

marionnegp:
## What I did

I checked CSS of elements that use crayon color tokens to make sure that the stylelint migration rules updated them correctly. I found that the linter changed the fallback hex value for gray tokens, instead of the token names. Checked what the colors were in RHDS 1.2.0 to correct the token names and values.

## Testing Instructions

1. Check gray tokens in the demos for footer, pagination, spinner, table, and tabs 

## Notes to Reviewers

No need for all three of you to review. Just tagged you all to see who's able to do it first. :)

### DIFF
--- a/elements/rh-footer/rh-footer.css
+++ b/elements/rh-footer/rh-footer.css
@@ -11,8 +11,8 @@
 
 footer,
 .global-base {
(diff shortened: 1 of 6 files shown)


bennypowers (approved):
Lonely gamers training meticulously

PR #1414, opened 2024-01-11
fix(table): allow mobile thead th to have data-label content

### LINKED ISSUE #1288: [bug] Table's mobile breakpoints exclude `<th>` inside `<tbody>`
The CSS for the mobile breakpoints is excludes the potential for a `<th>` to exist inside a `<tbody>`. For example:

```
:is(rh-table) thead ~ tbody tr td:before {
    font-weight: var(--rh-font-weight-heading-bold, 700);
    text-align: left;
    content: attr(data-label);
    display: inline-block;
  }
```

We should expand the CSS rules/selectors to include a `<th>` inside `<tbody>` for those edge cases where there might be one along with its column header:

```
:is(rh-table) thead ~ tbody tr td::before,
:is(rh-table) thead ~ tbody tr th::before {
    font-weight: var(--rh-font-weight-heading-bold, 700);
    text-align: left;
    content: attr(data-label);
    display: inline-block;
  }
```

Note: also adding the extra `:` to the `::before` pseudo-element.

Can be tested on the following demos:

- https://ux.redhat.com/elements/table/demo/row-and-column-headers/
- https://ux.redhat.com/elements/table/demo/column-headers/

bennypowers:
Closes #1288 

## What I did

1. let thead th have data-label content


markcaron (approved):
Last Guy To Merge

PR #1415, opened 2024-01-11
fix(table): mobile th:before styles

zeroedin:
## What I did

1. Improved mobile styles when `th` has a `data-label` attr in the `tbody`

Improves upon #1288

## Testing Instructions

1.


markcaron (approved):
Let's _(not)_ Get Too Mesmerized

PR #1421, opened 2024-01-15
fix: surface color for dark theme rows on Token pages

marionnegp:
Change copy button selector for dark theme rows, fix code font weight and copy button hover state

## What I did

Copy buttons for dark theme tokens now use a dark theme surface color

## Testing Instructions

Check that dark theme token rows look like this now in DP:

Screenshot 2024-01-15 at 1 33 40 PM


zeroedin (approved):
Lavender Gouache Transforms Murals

Issue #1425, opened 2024-01-17
[epic][docs] Docs Overhaul


markcaron:
Epic for the docs overhaul work in CY24:

- [ ] **Grid & layout**
  - [ ] Discovery on grid, layout and spacing
    - [ ] Create grid, layout and spacing styles for design system
  - [ ] Update pages with better content design (e.g. column structure, etc.)
- [ ] **Typography**
  - [ ] Create typography and base styles for design system
- [ ] **CSS standardization**
  - [ ] Gut/clean up CSS and create a CSS subsystem for the docs site
- [ ] **Template files standardization**
  - [ ] Switch NJK to WEBC for layout
  - [ ] Clean up template files
- [ ] **Wayfinding improvements**
  - [ ] Table of Contents on pages (list of links at top for easy win)
    - [ ] Discovery on Jump Links (outside layout) vs. TOC (at top of layout)
  - [ ] Side Nav element (new)
  - [ ] Categorization for components (e.g. containers, actions, visualization, etc.)
  - [ ] "Also Known As" (i.e. keywords) for helping users find elements (e.g. "tag" vs "label")
  - [ ] Next/Previous page links
  - [ ] GitHub links
    - [ ] Repo link in nav
    - [ ] Edit this page link on pages
    - [ ] Component source link
- [ ] **Page visual improvements**
  - [ ] Overview page
  - [ ] Make Props/Attr/Etc as a table
  - [ ] Best practices (i.e. this/not this) improvements
- [ ] **Demo improvements**
  - [ ] Build better Playground
    - [ ] View full demo link
  - [ ] Finish Code Block extensions
      - [ ] Design updates
      - [ ] Add show more/code disclosure
  - [ ] Relocate Demo as own "tab" (i.e. playground w/ knobs)
  - [ ] Sample component as replacement on Overview playground (i.e. live component embedded in page)
    - [ ] View demo tab link
- [ ] **Pattern pages improvements**
  - [ ] Add pattern pages for certain elements
  - [ ] Pattern help component pages (link to pattern pages)
- [ ] **Customization help**
  - [ ] Discovery on how to provide clearer info on tokens, CSS vars and shadow parts
- [ ] **Updating/adding sections**
  - [ ] About
  - [ ] Get started
  - [ ] Accessibility
  - [ ] Contributing
  - [ ] Team Support / Help page
- [ ] **Component statuses**
[...]

PR #1420, opened 2024-01-15
chore: adjust dev server config for pfe-tools 2

bennypowers:
## What I did

1. update dev server config to suit https://github.com/patternfly/patternfly-elements/pull/2655

## Testing Instructions

1. check out the dev server, 11ty demos, tests

PR #1465, opened 2024-01-31
docs: layout css rewrite, and adding uxdot-components 

zeroedin:
## What I did

1. Went down a deep rabbit hole. 🐇 🕳️ 
2. Improved layout using modern techniques, container queries and grid techniques.
3. Removed old and crufty css (created: styles/styles.css, styles/typography.css, styles/reset.css
4. Created several new `uxdox-*` components `uxdot-masthead`, `uxdot-sidenav`, `uxdot-feedback`, `uxdot-hero`
5. Separated page specific vs site /global css (would like to improve this by implementing nunjucks blocks instead of inserting a `<style>` tag inline
6. MOAR space.  "Content region" now has a maximum of 1300px wide of usable area and is container query responsive (we can debate what the size should be, this was an arbitrary decision while reworking the overall layout).  TOC area now is also responsive and is flipped to "top" of the content when in a smaller viewport.  Currently this is unused but ping me for my ideas on that on how we can utilize that space given a jumplinks/toc addition.
7.  `uxdot-sidenav`  is now generated from data (sorta) see `eleventy.config.cjs`, not thrilled with his, but it works for now, needs improved.  Intent was to reduce overhead of html in that template.  Ideally the data would be normalized prior to hitting the interface of the component.

### TODO:

- [ ] All the things.  No seriously, lots of pages are left to inspect and rebuild to take advantage of the new layout, drop their cruft and normalize.  I did the homepage `/` and `about/*` pages.
- [ ] Have discussions on what i've done.  I realize I just hacked and slashed and dropped my opinions in some areas now its feedback time.
- [ ] Reference #1425 I found myself breaking one thing and not wanting to just stop at a certain task.  This PR hits on a bunch of the issues listed in the epic, partially, or completely.  Figure out which ones I might have completed here and make sure to note those.
- [ ] Actually link up `uxdot-sidenav` to the trigger button and make it function on mobile with mobile styles.  I wanted to stop here as I know @marionnegp and @coreyvickery are currently doing research on this.  Think of what I've done as a placeholder for now.
- [ ] Add more todo's that I can't think of right now.

## Testing Instructions

1. View Deploy Preview


bennypowers:
How bad would the conflicts be if you tried to merge or rebase off #1420?

coreyvickery:
@zeroedin Feedback below.

- Switch order of icon and text link
- Increase padding in between icon and text to `16px`

zeroedin:
> @zeroedin Feedback below.
> 
> * Switch order of icon and text link
> * Increase padding in between icon and text to `16px`

@coreyvickery just a heads up @markcaron had an additional idea to include "current version" in this area too during office hours.  Eventually making that a dropdown similar to what patternfly is doing. Given that does this change anything for you?

coreyvickery:
@zeroedin Yeah, let's only do `16px` of padding and not change the order of anything.

zeroedin:
> @zeroedin Yeah, let's only do `16px` of padding and not change the order of anything.

@coreyvickery  https://github.com/RedHat-UX/red-hat-design-system/issues/1485 here is the issue @markcaron created.  Let me know if you want to mock some things up, or if Mark's example works for you.  

Right now the DP is showing your prior requested reversed icon/text order version: https://deploy-preview-1465--red-hat-design-system.netlify.app/

image

I can easily swap this back and put in the version.  Let me know what you think.

docs/_includes/partials/component/masthead.njk
```diff
@@ -0,0 +1,11 @@
+<uxdot-masthead role="header">
```
nikkimk, line 1: This should be `<uxdot-masthead role="banner">`

docs/patterns/index.md
```diff
@@ -13,42 +14,72 @@ summaries:
...
+</style>
+
 {# NOTE: all images in this view need to be 340 by 200 px in order to maintain same ratio. #}
 
 ## Overview
 Patterns compose elements and tokens with content and validation rules to 
 create uniform, accessible experiences.
 
-<div class="multi-column--min-400-wide margin-top--10">
+<nav class="grid">
```
nikkimk, line 54: Since there can be more that one <nav> on a page, the need unique `aria-label` elements. I suggest `<nav class="grid" aria-label="Patterns">`

docs/get-started/index.md
```diff
@@ -0,0 +1,54 @@
...
+    }
+  }
+
+</style>
+
+## Get started
+
+The Red Hat Design System for digital experiences gives designers and developers the tools to create accessible and on-brand user interfaces quickly. The role-specific pages linked below will guide you through using the design system.
+
+<nav id="get-started-nav">
```
nikkimk, line 34: Since there can be more that one <nav> on a page, the need unique `aria-label` elements. I suggest `<nav id="get-started-nav" aria-label="Get started">`

docs/elements/index.njk
```diff
@@ -0,0 +1,83 @@
...
+<section>
+  <h2>Overview</h2>
+  <p>Red Hat design system's elements are custom HTML elements - interactive 
+  building blocks of our design system. Each element was created to meet a 
+  specific UI and accessibility need. Elements should be used harmoniously 
+  together in the same space to create more intuitive user interfaces and 
+  experiences.</p>
+</section>
+
+<nav>
```
nikkimk, line 56: Since there can be more that one <nav> on a page, the need unique `aria-label` elements. I suggest `<nav class="grid" aria-label="Elements">`

docs/assets/javascript/elements/uxdot-sidenav.js
```diff
@@ -0,0 +1,273 @@
...
+        <div id="close-button-container">
+          <button id="close-button"
+              part="close-button"
+              aria-label="Close dialog"
+              @keydown=${this.onKeydownCloseButton}
+              @click=${this.toggle}>
+            <pf-icon set="patternfly" icon="close" size="lg"></pf-icon>
+          </button>
+        </div>
+        <nav part="nav">
```
nikkimk, line 121: Since there can be more that one <nav> on a page, the need unique `aria-label` elements. I suggest `<nav class="grid" aria-label="Main menu">`

nikkimk (dismissed):
I'd be interested in what @hellogreg thinks, but as I was testing side nav with a keyboard, I kept wanting to use the arrow keys to navigate which ended up scrolling the page behind the modal.

It may be good to use the RTIC so that users can quickly navigate the menu items with arrow keys and prevent the content behind the modal from scrolling.

CONTRIBUTING_DESIGN.md
```diff
@@ -73,7 +73,7 @@ tags:
 ### Getting pages into the navigation
 
 The navigation is an included template:
-`pages/component/header.njk`
+`pages/partials/component/sidenav.njk`
```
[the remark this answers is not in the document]
hellogreg: I like the idea of allowing arrow navigation, like in the [WAI disclosure nav example](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/). Their example still allows for tab nav, too.

zeroedin:
> I'd be interested in what @hellogreg thinks, but as I was testing side nav with a keyboard, I kept wanting to use the arrow keys to navigate which ended up scrolling the page behind the modal.
> 
> It may be good to use the RTIC so that users can quickly navigate the menu items with arrow keys and prevent the content behind the modal from scrolling.

I thought the same thing when keyboard navigating it. Will add!

CONTRIBUTING_DESIGN.md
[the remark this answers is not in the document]
zeroedin: @hellogreg heads up I'm going to copy your comment out of this thread as a response to mine about the same topic, this particular thread will be hidden when i resolve the specific code issue it is attached to.

zeroedin:
> > I'd be interested in what @hellogreg thinks, but as I was testing side nav with a keyboard, I kept wanting to use the arrow keys to navigate which ended up scrolling the page behind the modal.
> > It may be good to use the RTIC so that users can quickly navigate the menu items with arrow keys and prevent the content behind the modal from scrolling.
> 
> I thought the same thing when keyboard navigating it. Will add!
> 
> Also we talked about how sidenav when overlaying should maybe act like a dialog and focus trap, will have to figure out that between viewport states.

Comment from Greg:  (copied out of a code request change comment)

I like the idea of allowing arrow navigation, like in the [WAI disclosure nav example](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/examples/disclosure-navigation/). Their example still allows for tab nav, too.

marionnegp (changes requested):
I found some styling fixes. 

- [x] Is there a way to make these grid column widths equal? This example is from the Tokens: Overview page.

Screenshot 2024-03-01 at 4 05 40 PM

- [x] Body font weight in tile should be regular. This weight seems to be coming from the element though.  (SS edit: corrected font paths which improve look, but rh-tile does implement wrong font display vs text for body @marionnegp is going to open new issue)
- [ ] The margin between a heading and the section content should be `--rh-space-lg`
- [ ] Margin between two headings should be `--rh-space-2xl`
- [x] Until we get DPO type rules, can there be an 8px (`--rh-space-md`) space between each list item?
- [ ] Can the links in the TOC have no underline?
- [x] Can the hash marks on the token pages be hidden? 
- [x] I think the grid gap for tiles on Overview pages should be 32px (`--rh-space-2xl`). 48px looks a bit big.
- [ ] Until we get `<rh-side-nav>`, can the `uxdot-sidenav-dropdown-menu-item` elements have a font size of `--rh-font-size-body-text-md`? I believe that's what the current side nav is using.

zeroedin:
> * Is there a way to make these grid column widths equal? This example is from the Tokens: Overview page.

This is due to the `code-block` not wrapping / overflowing with the longer token name at certain viewport widths.  Might have to tweak styles there.  I'll see what I can do.

marionnegp:
@zeroedin, that's fine with me!

docs/assets/javascript/elements/uxdot-best-practice.js
bennypowers, line 1: there are quite a few of these, so in a future PR we might look at writing them in TS

zeroedin: Yeah I would definitely like to improve these in the future.

bennypowers (changes requested):
can we have dsd?

zeroedin:
> can we have dsd?

Should we do it as a follow up to this PR which is already bigger then original intent.  If we want to release both all at once, maybe we can create another branch to base this off of instead of main,  along with another PR pointing to that new branch that handles the dsd part as well, then merge those into main together?  Could knock off the other items on my fast follow list maybe too.

bennypowers:
well we get a fouc here, so if we can avoid that with an eleventy transformer, we should

coreyvickery:
@marionnegp @zeroedin Not sure if this is even helpful anymore, but here is a Figma with spacing and typography stuff.

https://www.figma.com/proto/cU0GFIoCz6EU5v3B3byw0U/Docs-overhaul---template-spacing-and-typography?page-id=1%3A2&type=design&node-id=1-1454&viewport=102%2C129%2C1&t=NETho8DWjiwj1qpp-1&scaling=scale-down&mode=design

zeroedin:
@coreyvickery definitely still helpful.  Thank you!

zeroedin:
Ref #1509 

@hellogreg  Usersnap was removed in this commit https://github.com/RedHat-UX/red-hat-design-system/pull/1465/commits/e74fd302cdacd3cf809aa51226571e80cb156792 on this PR.

marionnegp:
@zeroedin, it looks like there's only 1 slot in the example shortcode. Is it possible to have that container allow up to 3 slots/images, each with their own alt code?

bennypowers:
> @zeroedin, it looks like there's only 1 slot in the example shortcode. Is it possible to have that container allow up to 3 slots/images, each with their own alt text?

shortcodes have two kinds of input: *arguments* and *content*. All shortcodes can take zero-or-more arguments; only *paired* shortcodes can take content.

Arguments can be any object in scope of the shortcode call site. i.e. data from the document's frontmatter, or the 11ty data cascade. so in an element 'code' page for example, you could pass the custom element manifest for a given element as an argument to a shortcode, since that cem object is in scope on those pages.

content is any nunjucks content, so you can "slot" markdown content in, so-to-speak. But don't confuse shortcode content and HTML slots, they're not the same thing. There can only be one content string passed to the shortcode.

Here's an example of how you might define a content string as a nunjucks variable in scope, then pass that as an argument to a hypothetical shortcode `myShortcode` which takes a hypothetical arg called `myContentArg`:

```njk
{% set contentAsArg %}
Here's some **markdown** content _passed_ as an arg to a shortcode
{% endset %}

{% myShortcode myContentArg=contentAsArg %}
And Here's the `content` arg for this shortcode call
{% endmyShortcode %}
```

The javascript that implements that shortcode would look something like this:
```js
eleventyConfig.addPairedShortcode('myShortcode', function(content, { myContentArg = '' } = {}) {
  return /* html */`
<div class="my-shortcode">${myContentArg ? `
  <div class="my-arg-content">${myContentArg}</div>` : ``}
  ${content}
</div>`;
});
```

-----

in terms of code aesthetics, this 'feels' a little clunky to me. If the pattern in question has a small number of well-defined variants, perhaps it's better to take a variant arg instead

```njk
{% myShortCode variant="3-column" %}
  <div>1st col</div>
  <div>2nd col</div>
  <div>3rd col</div>
{% endmyShortCode %}
```

Or, if the "second slot"'s content is simple, like heading text, we could use an arg for that, like we do in the alert shortcode

zeroedin:
> > @zeroedin, it looks like there's only 1 slot in the example shortcode. Is it possible to have that container allow up to 3 slots/images, each with their own alt text?
> 
> shortcodes have two kinds of input: _arguments_ and _content_. All shortcodes can take zero-or-more arguments; only _paired_ shortcodes can take content.
> ... [truncated to save space]

@bennypowers There is no example shortcode anymore , `<uxdot-example>` replaced it.

The way we would do this now is:

```
<uxdot-example>
  <img alt>
  <img alt>
  <img alt>
 </uxdot-example>
 ```
 
For @marionnegp  use case (we discussed this offline), I mentioned the issue may end up being different widths per image as the example component would attempt to maximize the image to full scale unless a width-adjustment attribute was used.  The solution to that I suggested would be slotting a grid and using any layout you want there.
 
 ```
 <uxdot-example>
   <div class="grid">
       <img alt>
       <img alt>
       <img alt>
   </div>
 </uxdot-example>
 ```
 
`uxdot-example` in itself is just wrapper styles for border alignment padding and color-palettte (context).
 
The piece I'd was hoping to do was an auto generated `srcset` / `<picture>` implementation on SSR which it currently does not handle, that can be a follow up or we can rework this completely.

File: docs/assets/javascript/elements/uxdot-sidenav.js [hunk 7]
```diff
@@ -140,8 +140,24 @@ class UxdotSideNav extends LitElement {
...
-  #navItems = [];
+  get _navItems() {
+    const slot = this.shadowRoot?.querySelector('slot').assignedElements({ flatten: true }) ?? [];
+    const items = slot?.flatMap(slotted => {
+      return Array.from(slotted.querySelectorAll(`uxdot-sidenav-dropdown > details > summary, uxdot-sidenav-item > a`)) ?? [];
+    });
+    return items;
+  }
+
+  // TODO: figure out why getItems()s does appear to function js/ssr vs ts.
```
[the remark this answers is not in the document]
zeroedin: Understood but `#tabindex` shouldn't instantiate until connectedCallback() or @slotchange called method which both wrap the `isServer`.  

What I mean/meant by the comment is trying to decipher if it the issue was somehow related to the difference in the js vs the ts transpiled output.  

If I understand correctly `this.#tabindex.updateItems` which calls `this.#tabindex.getItems()` via the callback should still return items when called in connectedCallback or in the #rtiUpdate which gets called on @slotchange.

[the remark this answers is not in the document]
zeroedin: was able to resolve this issue, was in actuality a missing `return` statement 🫠 .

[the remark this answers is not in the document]
bennypowers: nice. glad you found an easy fix

bennypowers (approved):

marionnegp (approved):
Found some small design adjustments we'd want to make. They're not blockers, and I'll create new issues or add notes to an existing one.

PR #1474, opened 2024-02-05
fix(dialog): fixing issue with open, close, cancel events not firing

brianferry:
In implementing `rh-dialog` in my trials we noticed that we weren't getting back open, close, or cancel events from the dialog.  From looking at the issue it appears to be because the observer function changed and needs to call the previous function if the dialog is not a video.  

Here's a codepen where the dialog does not send the events as expected:  https://codepen.io/OhHeyBrian/pen/ExMRwVv

## What I did

1.  Added a check in the open observer that will send it to the event firing function if the modal type is not a video. 

## Testing Instructions

1.  There is a new demo at `/events` for `rh-dialog` that will output the events in the console.


bennypowers (changes requested):

bennypowers (approved):
Loquacious gentleman triggers memories

Issue #1309, opened 2023-10-26
[bug] `<rh-accordion-header>` is missing its `accents` slot


bennypowers:
### Describe the bug

[the rh-accordion-header docs](https://ux.redhat.com/elements/accordion/code/#rh-accordion-header) say it has an `accents` slot/part pair. it doesn't. it should

### Which Component?

rh-accordion-header

### Steps to reproduce

1. use accents slot on rh-accordion-header
2. nothing happens

### Expected behaviour

slot works, looks amazing. wins awards, etc.

### Demo

[Codepen](https://codepen.io/zeroedin/pen/eYxpdrj?editors=1100)

### Element HTML

```HTML
<rh-accordion id="tests">
  <rh-accordion-header>
    <h4>Test 1</h4>
    <rh-badge slot="accents" color="green">Pass</rh-badge>
  </rh-accordion-header>
  <rh-accordion-panel>...logs or whatever</rh-accordion-panel>
  <rh-accordion-header>
    <h4>Test 2</h4>
    <rh-badge slot="accents" color="red">Fail</rh-badge>
  </rh-accordion-header>
  <rh-accordion-panel>...logs or whatever</rh-accordion-panel>
</rh-accordion>
```

### Operating System (OS)

Linux

### Additional Info

h/t @aabughosh who originally raised the issue

cc @zeroedin who suggested doing the same for rh-tabs

bennypowers:
not directly related, but consider also adding content attributes, so...

```html
<rh-accordion id="tests">
  <rh-accordion-header heading="Test 1"></rh-accordion-header>
  <rh-accordion-panel>...logs or whatever</rh-accordion-panel>
  <rh-accordion-header heading="Test 2"></rh-accordion-header>
  <rh-accordion-panel>...logs or whatever</rh-accordion-panel>
</rh-accordion>
```

markcaron:
Closed by #1505

PR #1505, opened 2024-03-22
fix: adds accents slot to rh-accordion-header

diwanshi:
## What I did

1. Added an accents slot to rh-accordion-header which was missing from the code
2. Wrapped text and accents slot into a container span
3. Created a accents.html file with rh-tags in accemts slot
4. Added styles for mobile

## Testing Instructions

1.

## Notes to Reviewers

1. We still need to revisit the styles and mappings for the accents as we didn't find any in the figma

### DIFF
--- a/elements/rh-accordion/rh-accordion-header.css
+++ b/elements/rh-accordion/rh-accordion-header.css
@@ -107,6 +107,12 @@ span {
   text-align: start;
 }
 
+[part="container"] {
+  display: flex;
+  flex-direction: column;
+  gap: var(--rh-space-xs, 4px);
+}
+
 #button[aria-expanded="true"] #icon {
   rotate: calc(var(--_isRTL, -1) * 180deg);
 }
@@ -144,4 +150,10 @@ span {
   position: absolute;
   inset-block: 0;
   inset-inline-start: 0;
+}
+
+@media screen and (min-width: 576px) {
+  [part="container"] {
+    flex-direction: row;
+  }
 }
--- a/elements/rh-accordion/demo/accents.html
+++ b/elements/rh-accordion/demo/accents.html
@@ -1,0 +1,31 @@
+<rh-accordion>
+    <rh-accordion-header>
+      <h4>Item One</h4>
+      <rh-tag slot="accents" color="green" variant="outline" icon="circle-check">Green</rh-tag>
+      <rh-tag slot="accents" color="red" variant="filled" icon="info">Red</rh-tag>
+      <rh-tag slot="accents" color="orange" variant="outline" icon="bell">Orange</rh-tag>
+      <rh-tag slot="accents" color="purple" variant="filled" icon="ban">Purple</rh-tag>
+    </rh-accordion-header>
+    <rh-accordion-panel>
+      <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
+    </rh-accordion-panel>
+    <rh-accordion-header>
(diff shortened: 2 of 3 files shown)


bennypowers:
cc @marionnegp I seem to recall we removed this slot for a reason.
 Does that ring a bell to you?

diwanshi:
@bennypowers i just updated the description, this PR is regarding this bug https://github.com/RedHat-UX/red-hat-design-system/issues/1309

marionnegp:
@bennypowers, I don't remember wanting to remove this slot, and I saw @brianferry noted in chat that it should be there.

marionnegp:
I mocked up a [couple placement options](https://www.figma.com/file/4pHhr5IaEyWMWzqwC5z6ds/rh-accordion-accent-slot?type=design&node-id=1%3A20&mode=dev&t=lGQq00vOoGJwNEdw-1) (below the title or next to it), and I also mocked up an accordion with both an icon and accent slot. Not sure if we want to include both placement options, since some teams may need to keep the titles and tags on the same line, at least at larger breakpoints.

bennypowers:
set this to draft so that the new design specs can be implemented

diwanshi:
@marionnegp hey i'm planning to pick this up again to implement the new designs. Have we finalized which option we're going with or do we want to give user an option to select the placement?

marionnegp:
@diwanshi, I was thinking we could give the users the option to select placement, and we can add guidelines for using each option.

marionnegp (changes requested):
Is there a way to allow the tags to wrap to new lines on small screens? I'm seeing this at a 320px screen size:
Screenshot 2024-04-19 at 2 30 51 PM

Everything else looks good!

diwanshi:
@marionnegp in this case, you want each tag to be wrapped to new lines or only the ones that are getting cut?

marionnegp:
@diwanshi, only the ones that are getting cut off. It would look something like this:
Screenshot 2024-04-19 at 2 49 25 PM

elements/rh-accordion/rh-accordion-header.css
```diff
@@ -145,3 +154,10 @@ span {
   inset-block: 0;
   inset-inline-start: 0;
 }
+
+@media only screen and (max-width: 576px) {
```
zeroedin, line 158: Maybe a `@container` query here instead of `@media`

elements/rh-accordion/rh-accordion.ts
```diff
@@ -101,6 +117,13 @@ export class RhAccordion extends LitElement {
     }
   }
 
+  get #ctx(): RhAccordionContext {
+    const accents = this.accents === null
+                    || this.accents === '' as 'inline' ? 'inline' : this.accents;
```
zeroedin, line 122: could this just be:

```
const accents = this.accents ?? 'inline';
```

wouldn't we want to always return the state of either inline or bottom?

zeroedin, line 122:
```Suggestion
    const accents = this.accents ? this.accents : 'inline';
```

zeroedin (changes requested):
Please add a changeset, should be a minor change with the feature add.

elements/rh-accordion/rh-accordion-header.css
diwanshi: @zeroedin i tried adding @container but it doesn't work for me. this is the first time i'm using container query.

diwanshi:
@zeroedin added changeset.

nikkimk (approved):
Liquid gaffe torments mixologist

PR #1502, opened 2024-03-19
docs: reimplement design/code repo statuses

diwanshi:
## What I did

1. This PR is reimplimenting https://github.com/RedHat-UX/red-hat-design-system/issues/1268
2. Replaced repoStatus csv file with repoStatus.yaml file
3. Added 3 shortcodes for the following:
   - compact status list on each component
   - status checklist on each component
   - main Design/Code status table
4. Updated styles for repoStatus list and tables based on redesign
5. Added references to new shortcodes in the elements docs
6. Removed repoStatus references from all the patterns

## Testing Instructions

1. Compare current styles with this updated [figma](https://www.figma.com/proto/TaPXWNuUzEupDIH9w3L3hC/Design%2Fcode-status-redesign?page-id=0%3A1&type=design&node-id=3-3073&viewport=125%2C144%2C2&t=M9xgjCdbvXxTzqqQ-1&scaling=scale-down&mode=design)
2. Check if I've covered all the docs and none of the patterns are referencing the repoStatus
3. Check if the status is accurate for each component and on the main design/table

## Notes to Reviewers
1. The data might not be accurate as of now, I'm already working on collecting the updated data
2. If you've suggestions for better approach on yaml file, pls feel free to comment on that

PR #1507, opened 2024-03-28
feat(site-status): add `<rh-site-status>`

### LINKED ISSUE #356: [feat] Create `<rh-site-status>` element
### Description

New "status indicator" component that would be injected into the new unified footer for sites looking to report out on site outage.

The current implementation of the indicator shows the operational status across all of status.redhat.com and is not site specific. Notes from early conversations around this as it relates to the unified footer: https://docs.google.com/document/d/1vZ2taPSsG2MnpujdwpXxZvhH7ssnQgfAPVrzTPfaH4A/edit#bookmark=id.vq7rdtmnvsdg

Redhat.com is not currently included on status.redhat.com for reasons:
Per Danielle Smith "However, the complexity around determining what should be displayed on this page is not an XS LoE for our dev and infra teams.  Based on the comments in this thread, and what we learned this morning, it does not make sense to build out the comms and support processes to add it to the status-dot page at this time." 

### Acceptance Criteria

- [X] Design done
- [ ] Development done

### Image

_No response_

### Link to design doc

https://www.figma.com/file/IoKYfBlrcx2kXIGdC22Gjq/Footer---status-indicators?type=design&node-id=0%3A1&mode=dev

### Other resources

some thoughts on development:
- if we make the various status texts configurable via slots (with english fallback), then 18n becomes the consumer's responsibility (e.g. the cms)
- this being a red hat element, not a pfelement, we can rely on convention and have something like

`<rh-site-status site-ids="613,770">
  <span slot="ok">All systems operational</span>
</rh-site-status>`

zeroedin:
## What I did

1. Added `<rh-site-status>`

### TODO:

- [x] Add Tests
- [x] Add Docs (Style and Guidelines)
- [x] Rebase to `staging/chansey`
- [x] Answer: Is status API data ever translated? Assuming each translated page could potentially hit a different endpoint for status data.  Therefore should Loading text be translatable using a slot? <rh-site-status>Loading</rh-site-status>. Maybe not for MVP? (Answer: not current translated)
- [x] Review styles compared to figma
- [x] Update README.md

Closes #356

## Testing Instructions

1. View [Deploy Preview](https://deploy-preview-1507--red-hat-design-system.netlify.app/elements/site-status/demo/)


coreyvickery:
@marionnegp Documentation is here.

https://www.figma.com/file/JNc07T5kx5iNlprQB0MwFV/Element---Website-status?type=design&node-id=0%3A1&mode=design&t=6nx0sfCn7SxMQgTR-1

zeroedin:
will rebase this when https://github.com/RedHat-UX/red-hat-design-system/pull/1502 merges

elements/rh-site-status/docs/20-guidelines.md
```diff
@@ -0,0 +1,154 @@
...
+  .grid .example + .example {
+    margin-top: unset;
+    margin-block-start: unset;
+  }
+
+  .grid.icons .example {
+    margin-block-end: var(--rh-space-xl, 24px);
+  }
+
+  .do {
```
bennypowers, line 21: Iirc we already have helpers for do and don't

elements/rh-site-status/rh-site-status.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  #logger = new Logger(this);
+
+  #text = 'Loading';
+
+  #icon: TemplateResult = html``;
+
+  #isLoading = true;
+
+  #isApiStatus = (data: ApiStatus): data is ApiStatus => {
```
bennypowers, line 80: This can be static or in module scope

elements/rh-site-status/rh-site-status.ts
```diff
@@ -0,0 +1,149 @@
...
+    super.connectedCallback();
+    await this.#getStatus();
+  }
+
+  render() {
+    const { on = '' } = this;
+    return html`
+      <a href="https://status.redhat.com/" part="container" class="${classMap({ [on]: !!on })}">
+        ${this.#isLoading ? html`<rh-spinner size="sm"></rh-spinner>` : html`
+          <svg xmlns="http://www.w3.org/2000/svg" width="16" height="17" viewBox="0 0 16 17" fill="none">
```
bennypowers, line 101: Prefer viewbox

elements/rh-site-status/rh-site-status.ts
```diff
@@ -0,0 +1,149 @@
...
+      .then(response => {
+        if (response.ok) {
+          return response.json();
+        } else {
+          throw new Error(`${response.status} ${response.statusText}`);
+        }
+      })
+      .then((data: ApiStatus) => {
+        if (!this.#isApiStatus(data)) {
+          throw new Error('Invalid status data');
```
bennypowers, line 127: Elements shouldn't throw just by connecting

So handle this error somewhere

elements/rh-site-status/rh-site-status.ts
```diff
@@ -0,0 +1,149 @@
...
+        }
+        const statusText = data.status.description;
+        this.#text = textMap[statusText] || statusText.charAt(0).toUpperCase() + statusText.substring(1).toLowerCase();
```
bennypowers, line 130: Can we css this?

elements/rh-site-status/test/rh-site-status.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+        .and
+        .to.be.an.instanceOf(RhSiteStatus);
+    });
+  });
+
+  describe('when endpoint returns an All systems operational', function() {
+    let element: RhSiteStatus;
+
+    beforeEach(async function() {
+      overrideFetch(true, 200, 'All systems operational', () => Promise.resolve({ status: { indicator: 'none', description: 'All systems operational' } }));
```
bennypowers, line 40: Use Sinon for these and make sure to restore the stubs after each test

elements/rh-site-status/test/rh-site-status.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    beforeEach(async function() {
+      overrideFetch(false, 404, 'Not Found');
+
+      element = await createFixture<RhSiteStatus>(html`<rh-site-status></rh-site-status>`);
+      await element.updateComplete;
+    });
+
+    it('should show an error message', async function() {
+      // delaying 100ms to allow the fetch to complete
```
bennypowers, line 65: Use aTimeout

bennypowers (changes requested):

elements/rh-site-status/docs/20-guidelines.md
zeroedin: There is a different style of this with the red border via `{% example palette="wrong" %}`.  However as I understand it we are moving away from that by way of this this new design that was in the figma design for this page, and this code that as it stands is acting as a temporary setup between this and the larger docs rewrite PR which includes a `uxdot-` component for `do | dont`.

I had mentioned to @coreyvickery while I was working on this that the way I wrote it in this PR will be fairly easy to migrate  the component style when this page ultimately gets rebased into that PR.

elements/rh-site-status/rh-site-status.ts
zeroedin: After trying to grok this comment, I reverted the change I started to make.  

Can you please give me a bit more detail on what you are expecting here?  

The throw is there in order to catch the 404 or other network errors that returns if the API URL is incorrect/ service is down etc.  That is handled by the catch in the promise chain.  

I feel like doing this elsewhere would imho create extra state in the component which just feels extraneous.  Also I was thinking you might have meant to remove it from the connectedCallback lifecycle, but moving somewhere else would then cause a double render as well correct?

If I'm off base or missing something let me know.

elements/rh-site-status/test/rh-site-status.spec.ts
zeroedin: I gave this an attempt, but I couldn't wrap my head around trying to stub a private method.  I'd like to pair on this an see the approach you might take.

bennypowers: sounds good

sommething like this

```js
import { stub, type Stub } from 'sinon';

describe('site status', function() {
  let fetchStub: Stub;
  beforeEach(function() {
    fetchStub = stub(window, 'fetch');
  });
  afterEach(function() {
    fetchStub?.restore();
  });
  it('fetches', function() {
    expect(fetchStub).to.have.been.calledWith(URL, options);
  });
});
```

elements/rh-site-status/rh-site-status.css
```diff
@@ -0,0 +1,21 @@
+:host {
+  display: contents;
```
bennypowers, line 2: is it possible to avoid this? i feel like users are likely to complain

elements/rh-site-status/rh-site-status.ts
```diff
@@ -0,0 +1,149 @@
...
+      'description' in data.status &&
+      'indicator' in data.status
+    );
+  };
+
+  /**
+   * Sets color context for child components, overrides parent context
+   */
+  @colorContextProvider()
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette = 'dark';
```
bennypowers, line 76: does this need to be a provider? who are the child elements?

elements/rh-site-status/rh-site-status.ts
```diff
@@ -0,0 +1,149 @@
...
+        this.requestUpdate();
+      })
+      .catch( error => {
+        this.#logger.warn('Error loading site status:', error);
+        this.#text = 'Error loading status';
+        this.#icon = statusIconsMap['danger'];
+        this.#isLoading = false;
+        this.requestUpdate();
+      });
+  }
```
bennypowers, line 142:
```suggestion
    try {
      const data: ApiStatus = await fetch('https://status.redhat.com/index.json', {
        mode: 'cors',
        cache: 'no-cache',
        headers: {
          Accept: 'application/json',
        }
      })
        .then(response => {
          if (response.ok) {
            return response.json();
          } else {
            throw new Error(`${response.statusText}`);
          }
        });
      if (!RhSiteStatus.isApiStatus(data)) {
        throw new Error('Invalid status data');
      }
      const statusText = data.status.description;
      this.#text = textMap[statusText] || statusText;
      this.#icon = statusIconsMap[statusMap[data.status.indicator]];
      this.#isLoading = false;
      this.requestUpdate();
    } catch (error) {
        this.#logger.warn('Error loading site status:', error);
        this.#text = 'Error loading status';
        this.#icon = statusIconsMap['danger'];
        this.#isLoading = false;
        this.requestUpdate();
    }
  }
```
not picking on you (i once wrote an entire blog post about why try/catch with async await is overhyped) but in this case i feel like it'll be easier to read

bennypowers: we're likely to run into l10n issues with the hard coded error strings here

elements/rh-site-status/test/rh-site-status.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+            indicator: 'none',
+            description: 'All systems operational'
+          }
+        })
+      });
+      element = await createFixture<RhSiteStatus>(html`<rh-site-status></rh-site-status>`);
+      await element.updateComplete;
+    });
+
+    afterEach(function() {
```
bennypowers, line 29: you can put this in the same describe block as the declaration, then you don't need to remember to put it everywhere else

just use optional chaining to avoid errors for blocks that don't stub fetch
```js
    afterEach(function() {
      fetchStub?.restore?.();
    });

File: elements/rh-site-status/rh-site-status.ts (hunk 10 above)
Comment (zeroedin) [reply elements/rh-site-status/rh-site-status.ts:142]: > not picking on you 

All good I looked at various opinions around this.  I don't disagree with readability.  Will update.

> we're likely to run into l10n issues with the hard coded error strings her

See the TODO as well.  I pinged @kylebuch8 he doesn't know of any instance where the API right now is translated.

File: elements/rh-site-status/rh-site-status.ts (hunk 9 above)
Comment (zeroedin) [reply elements/rh-site-status/rh-site-status.ts:76]: Given the design specs, there is no change for any color context, so this being removed entirely.

File: elements/rh-site-status/rh-site-status.ts [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+
+  #isLoading = true;
+
+  async connectedCallback() {
+    super.connectedCallback();
+    await this.#getStatus();
+  }
+
+  render() {
+    return html`
```
nikkimk, line 81: So @adamjohnson and I recommend the following to communicate to screen reader users that the site status is loading.

```
<a href="https://status.redhat.com/" aria-busy="${this.#isLoading ? 'true' : 'false'}" aria-live="polite">
        ${this.#isLoading ? html`<rh-spinner size="sm"></rh-spinner>` : html`
            <rh-spinner size="sm">
                 <span class="visually-hidden"><slot="loading-text">Loading</slot></span>
            </rh-spinner>`
```

and then add a visually hidden class.

elements/rh-site-status/docs/40-accessibility.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Screen reader guidelines
+
+Website status should communicate the following to users.
+
+- What is the severity of the status icon
+- What the text says
+- That the text is also a link
+- Where the link will take them if selected
+
```
nikkimk, line 18: - If the status is still loading

nikkimk (changes requested):

elements/rh-site-status/rh-site-status.css
```diff
@@ -0,0 +1,21 @@
+:host {
+  display: block;
+}
+
+a {
```
adamjohnson, line 5: Looks like this component is currently missing focus and hover styles. 

The [DP docs](https://deploy-preview-1507--red-hat-design-system.netlify.app/elements/site-status/style/#interaction-states) show interaction states (hover = underline, focus = underline + outline).

adamjohnson (changes requested):

elements/rh-site-status/rh-site-status.ts
zeroedin: Ok finally remembered why this was in here.  `rh-spinner` is a child element and it needs context to know its on a dark background.  However I can't just set this context without setting a provider.  AND unfortunately it has to be set to reflect which adds the `color-palette` attribute to the host, even if it's not really configurable.

zeroedin: Also weirdly enough if the `rh-spinner` is in the lightdom then it works, but being in the shadow it does not

zeroedin: could just use a `<rh-surface>` internally but figured we were trying to avoid that.

elements/rh-site-status/demo/fetch-override.js
```diff
@@ -1,15 +1,13 @@
...
-          status,
-          statusText,
-          json
-        });
-      }
-      return target.apply(thisArg, args);
-    }
-  });
+  window.fetch = async () => {
+    return Promise.resolve({
```
bennypowers, line 6: why this change? the check was good

bennypowers (approved):
looking towards green messages

PR #1508, opened 2024-04-02
chore: update to @patternfly/pfe-core version 3.0

zeroedin:
## What I did

1. Updated and moved `@patternfly/elements` to devDependencies at version `^3.0.0` 
2. Added `@patternfly/pfe-core` at `^3.0.0` to dependencies
3. Updated `lit` to `^3.1.2`
4. Updated `tslib` to `^2.6.2`
5. Updated `typescript` to  `^5.4.3` 
6. Switch `ttypescript` to `ts-patch`, replacing `ttsc` with `tspc`
7. Updated typescript config `tsconfig` to match expected output.
8. Update api's for InternalController and RTI where necessary.

## Testing Instructions

1.


package.json
```diff
@@ -219,11 +219,11 @@
     }
   },
   "dependencies": {
-    "@patternfly/elements": "^2.4.0",
+    "@patternfly/elements": "^3.0.0",
```
bennypowers, line 222: I want this to be a dev dependency, or to be removed entirely

We should only need core and tools

nikkimk (approved):
Littered gymnasium: total mayhem

nikkimk (approved):
Litigious gentleman takes money

elements/rh-audio-player/rh-audio-player.ts
```diff
@@ -25,7 +27,9 @@ import { RhAudioPlayerScrollingTextOverflow } from './rh-audio-player-scrolling-
 import buttonStyles from './rh-audio-player-button.css';
 import rangeStyles from './rh-audio-player-range-styles.css';
 import styles from './rh-audio-player.css';
-import { RhTooltip } from '../rh-tooltip/rh-tooltip.js';
+
+import '../rh-surface/rh-surface.js';
+import '../rh-tooltip/rh-tooltip.js';
```
nikkimk, line 32: We importing `rh-tooltip` in line 32 if we already did in line 1?

bennypowers: b/c of type imports vs runtime

zeroedin:
> @zeroedin @bennypowers the playground demo is broken for button

This looks to be due to `@lit/context` not being available in the importmap.  Added it however looks like there is another issue that follows.

package.json
bennypowers: on second thought, i have to revise my previous comment. we still rely on `<pf-icon>` e.g. in footer, so for now we'll keep the dependency, but the only imports should be pf-icon, really

bennypowers:
If I'm not mistaken (and I hope I am) this is blocked on https://github.com/lit/lit/issues/4609

If I'm right, we can unblock ourselves with a workaround that adopts Lit's particular interpretation of the context protocol. That's arguably a breaking change though

zeroedin:
I reverted the prior change to `rh-tabs` which pulls in `@lit/context`.  I think that would at least unblock us to upgrade deps.  We can come back around to upgrading tabs when we have the fix for @lit/context

zeroedin:
> I reverted the prior change to `rh-tabs` which pulls in `@lit/context`. I think that would at least unblock us to upgrade deps. We can come back around to upgrading tabs when we have the fix for @lit/context

FYI: Will be reverting this again with a force push.  We can resolve with the latest released version of [`@lit/context`](https://github.com/lit/lit/releases/tag/%40lit%2Fcontext%401.1.1)

elements/rh-menu/rh-menu.ts
```diff
@@ -28,17 +29,28 @@ export class RhMenu extends LitElement {
...
   @colorContextConsumer() private on?: ColorTheme;
 
-  #tabindex = new RovingTabindexController(this);
+  #tabindex = new RovingTabindexController<HTMLElement>(this, {
+    getItems: () => this.#menuItems ?? [],
+  });
+
+  get #menuItems() {
+    const items = this._menuItems.filter((x): x is HTMLElement => x instanceof HTMLElement);
+    items.forEach(item => {
```
bennypowers, line 38: No side effects in getters, please

PR #1521, opened 2024-04-11
fix: fix <rh-alert> body copy and button color

marionnegp:
## What I did

Ensure that the copy in `<rh-alert>` doesn't change in different color contexts. All alert variants should look the same in light and dark theme.

## Testing Instructions

1.

### DIFF
--- a/elements/rh-alert/rh-alert.css
+++ b/elements/rh-alert/rh-alert.css
@@ -93,6 +93,7 @@ header ::slotted(:is(h1,h2,h3,h4,h5,h6)) {
 
 #description {
   font-size: var(--rh-font-size-body-text-sm, 0.875rem);
+  color: var(--rh-color-text-primary-on-light, #151515);
 }
 
 #description > ::slotted(*) {


elements/rh-alert/rh-alert.ts
```diff
@@ -117,10 +117,11 @@ export class RhAlert extends LitElement {
     const hasActions = this.#slots.hasSlotted('actions');
     const hasBody = this.#slots.hasSlotted(SlotController.anonymous as unknown as string);
     return html`
-      <div id="container"
+      <rh-surface id="container"
```
zeroedin, line 120: Instead of using a `rh-surface` element in the shadow root I think we'd want to make `rh-alert` itself a provider/consumer.  By doing so it essentially becomes a surface and can pass context down to its children like the buttons that are slotted.   

```
import { colorContextProvider, type ColorPalette } from '../../lib/context/color/provider.js';
import { colorContextConsumer, type ColorTheme } from '../../lib/context/color/consumer.js';
...

@colorContextProvider()
@property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette = 'lightest';

```

The text can managed by just adding:

```
#container {
  color: var(--rh-color-text-primary-on-light, #151515) !important;
}
```

That said I was trying to implement with hope it would work without reflection of the `color-palette` attribute, essentially making that context invisible to the user.  That said as soon as `reflect: false` is set the slotted buttons no longer receive the correct context.  /cc @bennypowers thoughts?

bennypowers: let's do this some other time

PR #1523, opened 2024-04-17
chore: eslint 9

bennypowers:
## What I did

1. eslint 9
2. pfe eslint 3
3. lint all the things


zeroedin:
Added optional dependency in package.json to get around the following error:

```
[esbuild] Failed to find package "@esbuild/darwin-arm64" on the file system
```

PR #1545, opened 2024-05-01
chore: correcting pfe to devDependencies to dependencies

zeroedin:
## What I did

1.  Moved `@patternfly/elements` to a dependency of the project so a user downloading the project will have access to `pf-icon` which RHDS still depends on.

## Testing Instructions

1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,16 @@
 {
   "name": "@rhds/elements",
-  "version": "1.3.2",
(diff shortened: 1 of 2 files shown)


bennypowers (changes requested):
Needs a patch changeset

bennypowers (approved):
Needs a patch changeset, otherwise lgtm

PR #1547, opened 2024-05-02
fix(site-status): automatically fetch status for domain

### LINKED ISSUE #1543: [feat] <rh-site-status>: endpoint attribute to hook the component up to the status.redhat.com API
### Description

`<rh-site-status>` looks great. But to get it to work, I have to include additional JS to make the API call to then set the status of the component. 

### Suggested solution

I'd like to be able to use an attribute like `endpoint` to be able to point to the status.redhat.com API so I don't have to write the JS glue code to get the component to display the correct status.

### Example API

```HTML
<rh-site-status endpoint="https://dn6mqn7xvzz3.statuspage.io/api/v2/summary.json"></rh-site-status>
```

### Additional context

This makes the component very simple to drop onto any Red Hat web property.

bennypowers:
Closes #1543

## What I did

1. configure site status to automatically fetch and display data for the current domain, falling back to all systems.

## Testing Instructions

1. check out the domains demo, which uses a super-secret private setter to set the domain

## Notes to Reviewers

Should probably be tested in production before release


elements/rh-site-status/rh-site-status.ts
```diff
@@ -31,108 +81,167 @@ const statusIconsMap: Record<string, TemplateResult> = {
...
+});
 
 // map statuspage.io's statuses to icon map keys
-const statusMap: Record<string, string> = {
-  'maintenance': 'warn',
-  'critical': 'danger',
+const STATUS_MAP = Object.freeze({
+  'none': 'ok',
+  'operational': 'ok',
+  // where does the 'maintenance' member come from? accd to https://metastatuspage.com/api,
```
zeroedin, line 96: Likely can be removed, the status map was taken from a prior implementation.  I never compared directly to the docs.  So my fault for the assumption the prior implementation was correct.

elements/rh-site-status/rh-site-status.ts
```diff
@@ -31,108 +81,167 @@ const statusIconsMap: Record<string, TemplateResult> = {
...
+          .then(getSummaryOrThrow);
+      const [component] = data.components
+          .filter(x => x.name === this.#domain)
+          .sort(byISO8601Property('updated_at'));
+      this.#component = component ?? null;
+      this.#status = data.status;
     } catch (error) {
       this.#logger.warn('Error loading site status:', error);
-      this.#text = 'Error loading status';
-      this.#icon = statusIconsMap['danger'];
```
zeroedin, line 134: Removing these makes the status demos no longer conform to the design spec https://www.figma.com/file/j7Rjo2W6PLloxOSXQOpll0/Red-Hat-Design-System-library?type=design&node-id=2447%3A129&mode=design&t=RL9K2HTQV8BGJwgL-1

kylebuch8 (approved):
Lacking Gumption; Timid Minion

Issue #1540, opened 2024-04-30
[feat] `<rh-health-index>` element


markcaron:
Create Health Index element based on original [PFE Health Index component](https://patternflyelements.org/v1/components/health-index/) for displaying the [health of containers in Catalog](https://catalog.redhat.com/software/containers/openshift3/ose-pod/57ea8d029c624c035f96f424?architecture=amd64&image=630e82fc80cc9b32912f31e7) and potentially other packages.

### New design
[View Figma file](https://www.figma.com/proto/0ozRanwnkD4f22XKKb7RMO/rh-health-index?type=design&node-id=7-3292&t=iIWkT5MRYTXHlB3s-0&scaling=min-zoom&page-id=7%3A3291)

#### Light theme
<img width="1474" alt="image" src="https://github.com/RedHat-UX/red-hat-design-system/assets/1890950/17143863-a4e8-46e5-bcc7-f1c930fbb26a">

#### Dark theme
<img width="1443" alt="image" src="https://github.com/RedHat-UX/red-hat-design-system/assets/1890950/396e13a9-4a99-4511-a65a-9b6751e82af9">

### Example on Catalog
<img width="1225" alt="image" src="https://github.com/RedHat-UX/red-hat-design-system/assets/1890950/1a747afc-bb8b-4ded-bd0e-3eceae6f276e">

<img width="220" alt="image" src="https://github.com/RedHat-UX/red-hat-design-system/assets/1890950/b2373ada-7ca9-4b08-af08-17855e3001bb">

PR #1549, opened 2024-05-02
feat(health-index): Adding `rh-health-index` element

markcaron:
## What I did

1. Added `<rh-health-index>` element

## Testing Instructions

1. **Design:** use [Figma mockup](https://www.figma.com/proto/0ozRanwnkD4f22XKKb7RMO/rh-health-index?type=design&node-id=7-3292&t=iIWkT5MRYTXHlB3s-0&scaling=min-zoom&page-id=7%3A3291)

## Notes to Reviewers

Closes: #1540


markcaron:
@coreyvickery this PR will need some screenshots and design docs when you get a chance.

adamjohnson:
Needs a [changeset](https://github.com/changesets/changesets). Run `npx changeset` in your terminal. Select a "minor" change when it asks. For the summary, I usually enter `Test` and then open the file in the code editor in the `.changeset` directory to add the real summary in the following format:

Emoji, summary, short description, example

```
✨ Added <rh-NAME-OF-ELEMENT>

A description of this component goes here.

<rh-NAME-OF-ELEMENT></rh-NAME-OF-ELEMENT>
```

See the [changelog](https://github.com/RedHat-UX/red-hat-design-system/blob/main/CHANGELOG.md) for more examples.

elements/rh-health-index/test/rh-health-index.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+    });
+
+    it('should upgrade', async function() {
+      element = await createFixture<RhHealthIndex>(html`<rh-health-index></rh-health-index>`);
+      const klass = customElements.get('rh-health-index');
+      expect(element)
+          .to.be.an.instanceOf(klass)
+          .and
+          .to.be.an.instanceOf(RhHealthIndex);
+    });
```
adamjohnson, line 19: Once we can denote which grade is "active" for the LG and XL variants, it'd probably be good to test for that here.

adamjohnson (changes requested):
Few accessibility things to check out. 👏

coreyvickery:
@markcaron @adamjohnson Small edits to active grade square.

### Default size (light)
- A: Fill: green-60 / Stroke: green-70
- B: Fill: green-40 / Stroke: green-60
- C: Fill: yellow-30 / Stroke: yellow-50
- D: Fill: orange-40 / Stroke: orange-60
- E: Fill: red-orange-60 / Stroke: red-orange-70
- B: Fill: red-70 / No stroke

### Default size (dark)
- A: Fill: green-60 / Stroke: green-40
- B: Fill: green-30 / Stroke: green-10
- C: Fill: yellow-30 / Stroke: yellow-10
- D: Fill: orange-40 / Stroke: orange-20
- E: Fill: red-orange-50 / Stroke: red-orange-30
- F: Fill: red-60 / red-40

elements/rh-health-index/README.md
bennypowers, line 1: please add some prose here to replace the sample text

elements/rh-health-index/demo/color-context.html
```diff
@@ -0,0 +1,74 @@
...
+
+.demo li {
+    margin: 0;
+    padding: 0;
+}
+
+.demo rh-surface,
+.demo div {
+    margin-block-end: 1rem;
+}
```
bennypowers, line 73:
```suggestion
.demo {
    margin: 2rem;
    padding: 0;
    list-style: none;
    width: 800px;
    display: grid;
    grid-template-columns: 1fr 1fr 1fr 1fr;
    gap: 2rem;
    & h2 {
      font-size: var(--rh-font-size-body-text-lg, 1.125rem);
    }
    
    & li {
        margin: 0;
        padding: 0;
    }
    
    & rh-surface,
    & div {
        margin-block-end: 1rem;
    }
}
```

elements/rh-health-index/rh-health-index.ts
```diff
@@ -0,0 +1,81 @@
...
+   */
+  @colorContextConsumer() private on?: ColorTheme;
+
+  render() {
+    const classes = {
+      [this.grade.toLowerCase()]: !!this.grade,
+    };
+    const boxes = ['A', 'B', 'C', 'D', 'E', 'F'];
+
+    if (this.size === 'sm') {
```
bennypowers, line 48: this should either be one long if / else chain or one big nested ternary. i'd go tern, personally

```js
const { on, size, active, box } = this;
return html`
  <div id="container"
       class="${classMap({ [size ?? '']: !!size,
                           [on ?? '']: !!on,
                           active, 
                           box })>${this.size === 'sm' ? html`
    ...
  ` : this.size === 'md' ? html`
    ...
  ` : html`
    ...
  `}
  </div>
`;
```

That would also make it easier to use shadow classes `active`, `box`, `sm`, and `md`, etc instead of `:host([size="sm"])` selectors

elements/rh-health-index/test/rh-health-index.spec.ts
bennypowers, line 1: we should add some ax tree assertions here for size variants, which assert the tree shape that @adamjohnson described in earlier comments.

bennypowers (changes requested):
a few changes, mostly nitpicks

hellogreg:
Just adding some quick notes based on a chat @markcaron and I had:

- We should add a `role="img"` to the element.
- The element can then (hopefully) auto-populate an `aria-label="Health: grade {#} out of A-F"` or some such.
- There may be translation considerations to... consider.

elements/rh-health-index/rh-health-index.ts
```diff
@@ -0,0 +1,81 @@
...
+      return html`${
+        (this.size === 'md') ? html`<div id="grade">${this.grade}</div>` : html``
+      }
+      <div id="container" class="${this.on}">
+        ${boxes.map(box => {
+          const active = box === this.grade;
+          const boxLowercase = box.toLowerCase();
+          const boxClasses = {
+            'box': true,
+            'active': active,
```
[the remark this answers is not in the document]
markcaron: Going to go with the `aria-label` as [Greg mentioned below](https://github.com/RedHat-UX/red-hat-design-system/pull/1549#issuecomment-2142843124).

elements/rh-health-index/README.md
markcaron: Adding in this next commit

elements/rh-health-index/test/rh-health-index.spec.ts
```diff
@@ -0,0 +1,21 @@
...
+      expect(document.createElement('rh-health-index')).to.be.an.instanceof(RhHealthIndex);
+    });
+
+    it('should upgrade', async function() {
+      element = await createFixture<RhHealthIndex>(html`<rh-health-index></rh-health-index>`);
+      const klass = customElements.get('rh-health-index');
+      expect(element)
+          .to.be.an.instanceOf(klass)
+          .and
+          .to.be.an.instanceOf(RhHealthIndex);
```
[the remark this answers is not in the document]
markcaron: Adding in this next commit

elements/rh-health-index/rh-health-index.ts
markcaron: Fixed in next commit

markcaron:
Still to do:

- Make @coreyvickery's [style tweaks](https://github.com/RedHat-UX/red-hat-design-system/pull/1549#issuecomment-2091916721)
- Refactor CSS
- Write size variant and a11y tests

elements/rh-health-index/rh-health-index.ts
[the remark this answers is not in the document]
adamjohnson: VoiceOver in Safari currently reads:

> Health: grade B out of AF

VoiceOver isn't doing a great job at denoting the `-` in between `A-F`. It literally speaks "AF". What do you think about the following small change to the `aria-label` text:

> Health: grade B out of **_A through F_**

?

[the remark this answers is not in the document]
coreyvickery: Makes sense to me.

elements/rh-health-index/test/rh-health-index.spec.ts
markcaron: May be moot with the `aria-label` now

elements/rh-health-index/rh-health-index.ts
bennypowers: It was nice pairing with you on this

elements/rh-health-index/test/rh-health-index.spec.ts
markcaron: Because of the `aria-label` on `#container`, the size variants have no real impact on the a11y Tree.

bennypowers (approved):

elements/rh-health-index/README.md
bennypowers: @adamjohnson resolve this comment to merge, should you see fit

coreyvickery:
@markcaron Great work, couple of tiny edits.

## `Small`, `Large`, and `Extra large` sizes

### F grade

Essentially just change `red-orange` to `red`.

- [ ] Change fill to `red-10` (demo has `red-orange-20`)
- [ ] Change stroke and thick bar thing to `red-70` (demo has `red-orange-70`)

That's it!

bennypowers:
I did some work on this:
- changed the colours per @coreyvickery 's comment (please check my work)
- reduced the size of the shadow DOM by using pseudo elements instead of HTML for presentation
- reduced the CSS by cutting the number of variables and rules

I'm worried about the aria-label in the shadow root, it's not translatable.
and should the shadow root be an ordered list?

markcaron:
@bennypowers, we had a discussion in Slack about the translation idea. For this MVP we're going to go with just English. But, in a future release we can add in a fixed set of translations that key off the `lang` attribute. We don't want users translating the text, since it'll be very specific (and will also improve the dev experience of not needing to translate).

The ordered list idea is a good one. Probably could be one with `type="A"` (though styling the numerals might be awkward). Keep in mind, though, that this would purely be semantic and, I believe, have little value to a screen reader since we've got it reading the `aria-label` instead.

bennypowers:
I made some changes to the internal semantics:
- remove image role
- remove all semantics from all elements in the shadow root
- apply meter role and semantics to the host via element internals

image

- Chrome gets all properties
- Firefox only gets role, valuetext, and label
- WebKit only gets role and label
- playwright snapshot is not picking up anything
- axe test passes

I tried setting aria attributes on the shadow container element instead of using internals, it did not affect the above, so let's stick with internals.

.changeset/calm-fireants-kneel.md
```diff
@@ -0,0 +1,9 @@
+---
+"@rhds/elements": minor
+---
+
+✨ Added <rh-health-index>
+
+Health index grades the health or security level of something.
+
+<rh-health-index grade="A">A</rh-health-index>
```
bennypowers, line 9:
~~~suggestion
```html
<rh-health-index grade="A">A</rh-health-index>
```
~~~

markcaron:
Seems to be a slight visual regression in the left and right borders of the `lg` and `xl` sizes. There's a slight gray border still.

image

bennypowers:
@markcaron good catch! Worst car, we can remove the start and end inline borders in the case where first or last elements are active. We can do that with classMap

adamjohnson:
> apply meter role and semantics to the host via element internals

Heads up: even though the a11y tree shows the semantic information we want, if you test this via VoiceOver, it says:

> B, level indicator

or 

> B, circular level indicator

☝️ For the first `sm` size.

whereas it should say something along the lines of:

> Health: grade B out of A through F

FWIW, I tested this with VoiceOver in Safari, Edge and Firefox. All produce similar results (mostly letters only, no other context). None produced the result we want.

I made a little [recording of VO + Edge](https://github.com/RedHat-UX/red-hat-design-system/assets/575865/53935825-e46b-4b3a-968a-0da761f3bc88).

bennypowers:
I did a little experimenting. Firefox behaves as expected with voiceover. Chrome behaves as you described. Setting the ariaValuetext to the label, chrome reads the label only


https://github.com/RedHat-UX/red-hat-design-system/assets/1466420/d6ea3835-598f-4ed1-8579-fb1ffd5fedb5

bennypowers:
@adamjohnson I removed the ariaLabel, and now both chrome and firefox appear to announce acceptably through voiceover. PTAL

bennypowers:
OK here's what I've discovered:

If `aria-valuetext` is a single character, voiceover loses its lunch on chrome and webkit. This is true regardless of whether we're using `ElementInternals`, aria attrs on a div, or aria-attrs on a meter element.

I do believe I have everything working exactly as intended on macOS / voiceover across chrome/ff/safari. 

Check out the [DP screen reader demo](https://deploy-preview-1549--red-hat-design-system.netlify.app/elements/health-index/demo/screen-readers/)

elements/rh-health-index/rh-health-index.ts
```diff
@@ -0,0 +1,81 @@
...
+   * Sets the size of the health index
+   * Defaults to `md`
+   */
+  @property({ reflect: true }) size: 'sm' | 'md' | 'lg' | 'xl' = 'md';
+
+  /**
+   * Sets the health grade
+   * Defaults to `A`
+   */
+  @property({ reflect: true, attribute: 'grade' }) grade: 'A' | 'B' | 'C' | 'D' | 'E' | 'F' = 'A';
```
bennypowers, line 32:
```suggestion
  @property({ reflect: true }) grade: 'A' | 'B' | 'C' | 'D' | 'E' | 'F' = 'A';
```

.changeset/clean-emus-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": major
```
bennypowers, line 2: why the changeset?

elements/rh-cta/demo/customization.html
bennypowers, line 1: why the new file?

bennypowers (changes requested):
@zeroedin looks like some unrelated changes got committed?

zeroedin:
> @zeroedin looks like some unrelated changes got committed?

Oops sorry will get that fixed up.

bennypowers:
this should get some a11y docs, regarding at least the meter role and the use of aria-label with aria-label and aria-valuetext for customizing (and a warning against doing so without due consideration)

we should also open a ticket to get our strings translated into all the official langs

markcaron:
@coreyvickery and @adamjohnson, I believe I've finished the docs. Can you give this a review?

Adam, please double-check the code as well.

adamjohnson:
> I do believe I have everything working exactly as intended on macOS / voiceover across chrome/ff/safari.

I am having trouble getting VO + Firefox to read the grade. [Video of my results](https://github.com/RedHat-UX/red-hat-design-system/assets/575865/037cbf41-0141-4ea5-ae86-62919110c6df). 

Environment: OSX + VO + evergreen Safari/Firefox.

elements/rh-health-index/docs/40-accessibility.md
```diff
@@ -0,0 +1,21 @@
+## Implementation
+
+Most accessibility concerns should be covered by implementing health index per our guidelines. You can double-check your implementation by ensuring the following:
+
+- Health index should not be interactive or keyboard-focusable
+- Health index should be placed in a context that makes its purpose clear to all users—with and without assistive technologies
+- Health index should communicate its grade both visually and to assistive tech
+
+Test your experience with a keyboard and <a href="/accessibility/qa-testing/#screen-readers">screen reader</a> to help verify these accessibility conformance checks.
+
```
adamjohnson, line 10: Maybe here add something like:

> Under the hood, health index uses the [meter role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/meter_role) and corresponding ARIA attributes to communicate to assistive technology. Inspect this element using the [DevTools Accessibility Tree](https://developer.chrome.com/docs/devtools/accessibility/reference#pane) to see each value.

adamjohnson (approved):
Looks like Benny has an environment where this works with VO + Firefox. If users have any issues with this, we can cross that bridge when that happens.

Issue #1550, opened 2024-05-03
[bug] 'New' `<rh-tag>` stretched on mobile


coreyvickery:
### Describe the bug

The `New` tag on the [Design/code status](https://ux.redhat.com/design-code-status/) page appears stretched on small viewports. Please make sure the `Beta` and `Experimental` tags are also correct as part of this issue.

### Which Component?

rh-tag

### Steps to reproduce

Go to the [Design/code status page](https://ux.redhat.com/design-code-status/) and reduce the window to mobile size (screenshot is 360px).

### Expected behaviour

All tags should have the same padding.

### Screenshots

<img width="365" alt="Screenshot 2024-05-02 at 5 00 31 PM" src="https://github.com/RedHat-UX/red-hat-design-system/assets/45767595/e3522a0c-24bd-4ba6-94fd-6b5574090f7c">

### Operating System (OS)

iOS, macOS, Linux, Windows

### Browser

Chrome

PR #1551, opened 2024-05-03
docs: fix repo-status tag padding in mobile

diwanshi:
## What I did

1. This MR fixes https://github.com/RedHat-UX/red-hat-design-system/issues/1550
2. Added a span around rh-tag (overall status) in the Repo Status table to fix the extra padding issue in mobile view

## Testing Instructions

1. Go to https://ux.redhat.com/design-code-status/
3. Check the mobile view for the repo status table
4. the overall status i.e New, Experimental, Beta should not have extra padding in mobile view

### DIFF
--- a/docs/_plugins/shortcodes/repoStatus.cjs
+++ b/docs/_plugins/shortcodes/repoStatus.cjs
@@ -251,9 +251,12 @@ function repoStatusTable() {
 <tr>
 <td data-label="Name">
   <a href="/elements/${listItem.name}">${listItem.name}</a>
-  ${listItem.overallStatus !== 'Available' ? `<rh-tag color="${STATUS_LEGEND[listItem.overallStatus].color}" variant="${STATUS_LEGEND[listItem.overallStatus].variant}">
-    ${listItem.overallStatus}${STATUS_LEGEND[listItem.overallStatus].icon}
-  </rh-tag>` : ''}
+  ${listItem.overallStatus !== 'Available' ? `
+  <span>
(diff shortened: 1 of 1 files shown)


docs/_plugins/shortcodes/repoStatus.cjs
```diff
@@ -251,9 +251,12 @@ function repoStatusTable() {
...
   <a href="/elements/${listItem.name}">${listItem.name}</a>
-  ${listItem.overallStatus !== 'Available' ? `<rh-tag color="${STATUS_LEGEND[listItem.overallStatus].color}" variant="${STATUS_LEGEND[listItem.overallStatus].variant}">
-    ${listItem.overallStatus}${STATUS_LEGEND[listItem.overallStatus].icon}
-  </rh-tag>` : ''}
+  ${listItem.overallStatus !== 'Available' ? `
+  <span>
+    <rh-tag color="${STATUS_LEGEND[listItem.overallStatus].color}" variant="${STATUS_LEGEND[listItem.overallStatus].variant}">
+      ${listItem.overallStatus}${STATUS_LEGEND[listItem.overallStatus].icon}
+    </rh-tag>
+  </span>` : ''}
```
zeroedin, line 259: Should we instead have the span wrap both the `<a>` and the `<rh-tag>` that way it stays on the same line in the mobile view instead of dropping down a line? 

Current:

image

Span wrapping both `<a>` and `<rh-tag>`

Screenshot 2024-05-06 at 4 35 15 PM

diwanshi: @zeroedin thanks for pointing that out, i've pushed a commit for the same

zeroedin (approved):

PR #1554, opened 2024-05-06
chore: improve test run speed by removing full build deps

zeroedin:
## What I did

1. Improved running speed of `npm run test` I believe we don't need the full build stack as a dependency for tests such as the docs and artifact builds.
2. Created `build:test` which reduces deps needed for test runs.

## Testing Instructions

1.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,10 +116,17 @@
         "docs"
       ]
     },
+    "build:test": {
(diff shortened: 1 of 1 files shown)


bennypowers (approved):
Nice pick up

PR #1557, opened 2024-05-08
fix(card): header font

bennypowers:
## What I did

1. set heading font for card
2. change the html and css for the 'image header with text heading' example in the variants.html demo file

## Notes to Reviewers
How will this affect localized pages?

### DIFF
--- a/elements/rh-card/demo/variants.html
+++ b/elements/rh-card/demo/variants.html
@@ -1,128 +1,130 @@
-<rh-card>
-  <h2 slot="header">Default card, slotted content and footer</h2>
-  <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.
-    Nullam eleifend elit sed est egestas, a sollicitudin mauris
-    tincidunt. Pellentesque vel dapibus risus. Nullam aliquam
-    felis orci, eget cursus mi lacinia quis. Vivamus at felis sem.</p>
-  <rh-cta variant="primary" slot="footer">
-    <a href="#">Call to action</a>
-  </rh-cta>
-</rh-card>
+<section id="card-variants">
+  <rh-card>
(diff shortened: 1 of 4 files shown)


elements/rh-card/rh-card.css
```diff
@@ -10,12 +10,13 @@ article {
 }
 
 ::slotted(*) {
-  font-family: inherit !important;
+  /* font-family: inherit !important; */
   line-height: inherit !important;
 }
 
 ::slotted(:is(h1, h2, h3, h4, h5, h6)) {
   font-size: inherit !important;
```
marionnegp, line 18: Does the font size have to be inherited here? When I had to use a heading in the card body, I needed it to be a different size than the body copy.

elements/rh-card/demo/variants.html
```diff
@@ -156,20 +158,25 @@ <h2 slot="header">Card with slotted image header. Full width image.</h2>
...
-
     & img {
       display: block;
       min-width: 100%;
       height: auto;
     }
+
+    &.heading {
+      & h2 {
+        font-weight: var(--rh-font-weight-heading-bold, 700);
```
marionnegp, line 174: Font weight for headers should be `--rh-font-weight-heading-medium`. Not sure if it matters in the demo, but it might confuse users to see a bolder-than-usual heading.

elements/rh-card/rh-card.css
```diff
@@ -10,12 +10,13 @@ article {
 }
 
 ::slotted(*) {
-  font-family: inherit !important;
+  /* font-family: inherit !important; */
```
adamjohnson, line 13: This seems like a holdover from being commented out. Remove it and you're G2G.

adamjohnson (approved):
Lower greater tonsil meteorite 👍

marionnegp (approved):

PR #1558, opened 2024-05-08
fix(card): hide header container when it's empty

### LINKED ISSUE #1383: [bug]: `<rh-card>` header when empty does not display none
If a `slot="header"` is not present, the `#header` shadow DOM element correctly receives a `.empty` class but the style is overwritten due the style `#header { display: flex }` having a higher specificity. `!important` is needed on the `.empty` class.

Reference #1380

bennypowers:
Fixes #1383

## What I did

1. hide empty containers

## Testing Instructions

1. [/elements/card/demo/](https://deploy-preview-1558--red-hat-design-system.netlify.app/elements/card/demo/)
2. run this in js console
    ```js
    $('rh-card').querySelector('h2').remove();
    ```
4. observe that header container is hidden.


elements/rh-card/rh-card.css
```diff
@@ -6,7 +6,7 @@ article {
 }
 
 .empty {
-  display: none;
+  display: none !important;
```
zeroedin, line 9: suggest using:

```css
#header.empty {
  display: none;
}
```

instead of an important.

.changeset/plenty-rabbits-scream.md
```diff
@@ -0,0 +1,4 @@
+---
+"@rhds/elements": patch
+---
+`<rh-card>`: hide header container when there is no header content
```
adamjohnson, line 4: Not just the header, but `#header`, `#body`, and `#footer`.

adamjohnson (approved):
Lattice Gopher Tomcat Melons

PR #1559, opened 2024-05-08
chore: attempt to make builds less flakey

bennypowers:
## What I did

1. use wireit's file cache to hopefully prevent failures like https://github.com/RedHat-UX/red-hat-design-system/actions/runs/8988133555/job/24688276034


zeroedin (approved):
LGTM

Issue #566, opened 2022-10-04
[feat] Inline promo


marionnegp:
### Description

The inline promo needs to be formalized as a component because it's being used consistently on .com and will be used as a personalization.

### Acceptance Criteria

- [x] Design done
- [x] Development done

### Image

**Featured inline promo:**

![Image](https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/f5038162-2355-4f73-8962-338cf7f3a90a)

**Standard inline promo:**

![Image](https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/52c0cb99-cf15-442e-82ea-9aa16f1e304c)

**Full-width promo:**

![Image](https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/5edc5d82-83cc-42ae-9c00-ec5d1aaf6ff5)

### Link to design doc

[Figma file](https://www.figma.com/file/Bfdo9e0qzDx6quuMWYdi9K/Promo-(WIP)?type=design&node-id=10%3A4869&mode=design&t=sH5oo85U3wslK85W-1)

marionnegp:
Questions:
- The inline promos are basically cards and should maybe be a card variant. How do we define inline promo so RHDS users know when to use promo vs a regular card?
- Designers like the option to have a promo without a border. This is the use case that full-width promo is trying to capture. Does this make sense to have as a formalized element/pattern in RHDS?

marionnegp:
@hellogreg @markcaron, since we're still working on documenting correct usage and finding out whether this can extend card accessibly, can this be moved to the Chansey milestone?

hellogreg:
@marionnegp Done!

adamjohnson:
@markcaron I'm looking into this component now and Marion [asked some great questions](https://github.com/RedHat-UX/red-hat-design-system/issues/566#issuecomment-1813273913).

> The inline promos are basically cards and should maybe be a card variant.

Looking at Inline Promo, do you think it should be its own component or should it be an `<rh-card>` [variant](https://ux.redhat.com/elements/card/demo/variants/) or [pattern](https://ux.redhat.com/patterns/card/)?

bennypowers:
FWIW I'm leaning towards card variant. Can we maybe put together a PoC and shop it around to studio devs?

PR #1514, opened 2024-04-04
feat(video-embed): add rh-video-embed

nikkimk:
## What I did

1. Created `rh-video-embed` element

## Testing Instructions

1. Review the [DP Docs](https://deploy-preview-1514--red-hat-design-system.netlify.app/elements/video-embed/).
2. Test all 4 demos.

PR #1569, opened 2024-05-15
feat(card): promo variant

adamjohnson:
## What I did

1. Added a new demo + styles for `<rh-card>` to be used for an "Inline Promo".
1. Fixed an issue with the `.empty` class not having enough specificity
1. Fixed a spacing issue with the `#body` when the `#footer` is `.empty`

## Testing Instructions

1. Review the [Inline Promo Issue](https://github.com/RedHat-UX/red-hat-design-system/issues/566)
1. Check the Deploy Preview for [`<rh-card>`'s Promo demo](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo/)
    * [Promo](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo/)
    * [Full Width Promo ](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo-full-width/)
    * [Standard](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo-standard/)
    * [Narrow](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo-narrow/)

## Notes to Reviewers

[Benny thinks](https://github.com/RedHat-UX/red-hat-design-system/issues/566#issuecomment-2106269997) that Inline Promo should be a `<rh-card>` variant. Take a look at what I've done here and see if you agree.

### Questions:

1. Who should we show this to? Who will take on that task? ("...shop it around to studio devs." -Benny)

Related to #566


elements/rh-card/demo/inline-promo.html
```diff
@@ -0,0 +1,243 @@
+<rh-card class="inline-promo">
```
bennypowers, line 1:
```suggestion
<rh-card variant="inline-promo">
```

bennypowers: i think this should be a variant

bennypowers (changes requested):

elements/rh-card/demo/inline-promo.html
bennypowers: to elaborate: we don't want to make the value of `class` a public api on our elements

.changeset/fair-buttons-punch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-card>`: Add Inline Promo demo + styles
```
bennypowers, line 5: these changesets should be user-facing, instead of developer-facing

so, instead of listing what you did as a programmer here ("added demo and styles"), instead say what was added (:sparkles: Added inline promo variant) and put a short description of the feature and a fenced code block demonstrating it's typical usage

elements/rh-card/demo/inline-promo.html
```diff
@@ -0,0 +1,85 @@
+<rh-card variant="inline-promo">
+  <svg slot="inline-promo" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 320 190">
```
bennypowers, line 2: i'm not sure we should call this slot `inline-promo`. we are likely to introduce other variants (see #1188 ) so i think we'd be better off calling this slot `image`. WDYT @marionnegp ?

elements/rh-card/demo/inline-promo.html
```diff
@@ -0,0 +1,85 @@
...
+    <title>A placeholder image in a card header slot</title>
+    <rect x="0" y="0" width="320" height="190" fill="#e7f0f9"/>
+    <line x1="0" y1="0" x2="320" y2="190" stroke="#d6e7f5"/>
+    <line x1="320" y1="0" x2="0" y2="190" stroke="#d6e7f5"/>
+  </svg>
+
+  <h2>Featured: left aligned with image</h2>
+  <p>You can customize the <code>#body</code> by using the <code>::part</code> selector. Using this technique, you can have padding on the top, bottom, left, right, or none at all.</p>
+  <rh-cta>
```
bennypowers, line 11:
```suggestion
  <rh-cta slot="footer">
```

elements/rh-card/demo/inline-promo.html
```diff
@@ -0,0 +1,85 @@
...
+    <a href="#">Call to action</a>
+  </rh-cta>
+</rh-card>
+
+<rh-card variant="inline-promo" reverse>
```
bennypowers, line 16: do you think it's possible to avoid adding all these new boolean attributes like `reverse`, `one-col`, `full-width`, etc? each feature we add is a potential breaking change down the road when we inevitably discover that someone needs to modify or remove it

how much of this can be accomplished with existing tokens and parts?

elements/rh-card/demo/patterns.html
bennypowers, line 1: see #1188

perhaps we should make this change there, or in a separate pr

bennypowers (changes requested):

elements/rh-card/demo/inline-promo.html
adamjohnson: Adding `slot="footer"` to the CTA breaks the design/layout. Can you explain this suggestion further?

elements/rh-card/demo/patterns.html
adamjohnson: Reverted in 4331e06.

.changeset/fair-buttons-punch.md
adamjohnson: Updated in [f603b8e](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/commits/f603b8e3dcc56eefc05748b6da40a4b5d6a68ed1). 

AFAICT, this is the first variant to follow this convention. Double check to make sure this is the correct way to do it.

elements/rh-card/demo/inline-promo.html
```diff
@@ -0,0 +1,85 @@
...
+
+<script type="module">
+  import '@rhds/elements/rh-cta/rh-cta.js';
+  import '@rhds/elements/rh-card/rh-card.js';
+</script>
+
+<style>
+  /*****************************************************************************
+   * FOR DEMO ONLY
+   *****************************************************************************/
```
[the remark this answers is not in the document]
adamjohnson: Removed in [57e96f7](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/commits/57e96f7b211bdcb27ac9078b0a75535c3c0ce8a3).

elements/rh-card/demo/inline-promo.html
adamjohnson: Resolved in [089d9ac](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/commits/089d9ac0b3de92225ae4d65cfba046e54340d7a2).

elements/rh-card/demo/inline-promo.html
adamjohnson: It's probably possible to avoid adding _most_ of them. I did use an adjacent sibling combinator for the `full-width` variant. Not sure if I can move that out of `rh-card.css` or not.

To be clear, do you want me to move any styles that can be accomplished via `::part()` to a `<style>` tag inside `inline-promo.html`? Then, if component authors would want a "one column" inline promo, they'd have to copy these styles to their pages as well.

marionnegp (changes requested):
- The demo for the full-width promo shouldn't have an inline margin since it's supposed to span the full width of a page/section. When this is shared with the devs, do you think it's worth having a demo that shows the inline promo and full-width promos within a layout?
- On desktop, the full-width promo image shouldn't have a 64px padding above it. It should basically look like the inline promo but without a border. 
Screenshot 2024-05-30 at 2 51 13 PM

Let me know if you have any questions! This is kind of an odd element/pattern.

elements/rh-card/demo/inline-promo.html
adamjohnson: Renamed the `inline-promo` slot to `image` in 75da1c7.

> Would a user have to add content to a slot?

No. If a slot doesn't have content, it gets: `display: none;`. Technically, they _should_ add content to the `body` slot.

> Is there a reason the image slot is an svg? Inline promos might not always use an svg image.

The SVG is just used for demo purposes. I have styled it to work just the same with an `<svg>` or ``. You can [test it](https://github.com/RedHat-UX/red-hat-design-system/assets/575865/0e2b0994-45d0-4cd5-8524-c2e4c2bd8b6e) by adding an `` in place of the SVG + a `slot="image"` attribute.

adamjohnson:
@marionnegp 

> The demo for the full-width promo shouldn't have an inline margin since it's supposed to span the full width of a page/section. 

Solved in ab48223.

> When this is shared with the devs, do you think it's worth having a demo that shows the inline promo and full-width promos within a layout?

I will defer to your expertise on this one. If yes, perhaps we can collaborate to accomplish this?

> On desktop, the full-width promo image shouldn't have a 64px padding above it. It should basically look like the inline promo but without a border.

Solved in ad21ced.

bennypowers:
progress here looks really good, @adamjohnson 
I've run some experiments and determined that `image` slot can be used to implement the Image pattern from [the figma library](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/Red-Hat-Design-System-library?node-id=439-4200&m=dev). We would be wise to expend some thought and care regarding the asset and logo patterns there as well, so make sure we're setting up for success

Here is an example of how we would implement asset pattern 

image

and image pattern

image

bennypowers:
with the small changes in #1590 we can implement all the patterns in figma without modifying card.

marionnegp (commented):
> video, quote, and logo appear to me to have both image and logo? is that correct @marionnegp ?

The quote and logo card patterns do have both, but the video would pull in `<rh-video>` (#1514) eventually.

>> When this is shared with the devs, do you think it's worth having a demo that shows the inline promo and full-width promos within a layout?

>I will defer to your expertise on this one. If yes, perhaps we can collaborate to accomplish this?

Yes! I think you could probably show it to some devs first, though, and see if they have questions that a layout demo could answer. Would hate to add that work if the other FEDs don't need it.

elements/rh-card/demo/inline-promo.html
adamjohnson: Following up here, I removed the `one-col`, `reverse`, `full-width`, and `standard` attributes/properties and converted them to classes on the `:host` in the demo (0309f97, decbb9b, f942a05).

I believe this is in line with what you're doing in #1590—minimizing our API surface and showing users it's possible to customize these components with their own CSS (via `::part` and styling the lightdom).

PTAL at [the DP](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/inline-promo/) and let me know if I'm on the right track here. TIA!

adamjohnson:
> > video, quote, and logo appear to me to have both image and logo? is that correct @marionnegp ?

> The quote and logo card patterns do have both, but the video would pull in <rh-video> (https://github.com/RedHat-UX/red-hat-design-system/pull/1514) eventually.

FYI: the `rh-video` PR has [a demo](https://deploy-preview-1514--red-hat-design-system.netlify.app/elements/video/demo/card-with-video/) that combines `rh-card` + `rh-video` which may be relevant here.

.changeset/fair-buttons-punch.md
bennypowers: looks good but might need updating after 75da1c74251ad775c50f8ea94fd90a1e28199d8d

adamjohnson: Updated changeset in [cef0ff3](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/commits/cef0ff3bd8fb203f32c7f6585562bd3dad59d738) to match 1:1 to the demo.

elements/rh-card/demo/inline-promo.html
bennypowers: yeah looks good

elements/rh-card/demo/inline-promo.html
bennypowers: my thought process is that
- ctas usually  go in card footers
- lots of existing html has cta in card footer
- so toggling the variant to inline-promo should (maybe) work?

markcaron (approved):
Low-key Gossip Triggers Mayhem

.changeset/fair-buttons-punch.md
```diff
@@ -0,0 +1,24 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-card>`: ✨ Added Inline Promo variant
```
zeroedin, line 5: @markcaron made an oopsie on review

This comment is a blocker to stop accidental auto merge. Resolve after full review has been done.

elements/rh-card/demo/inline-promo.html
adamjohnson: Indeed, that is a logical thought process and seems like it could certainly happen in the wild.

Technically, you _could_ put the CTA in the `footer` slot and style it ala:

```css
rh-card[variant="inline-promo"]::part(footer) {
    grid-column: 2;
}
```

Then it looks like this:

Inline promo with CTA in footer - spacing visualized

Here's what's happening with the grid behind the scenes:

Inline promo with CTA in footer - CSS grid visualized

Unfortunately, it's [not possible](https://stackoverflow.com/questions/45264354/is-it-possible-to-place-more-than-one-element-into-a-css-grid-cell-without-overl) to place more than one grid element into a cell without overlap. CSS grid doesn't have a concept of flow.

Hopefully, when people see the CTA in the wrong spot, they'll remove `slot="footer"` from the CTA (fixing their layout), actualize peak web component mastery and move on to sweating each pixel in their image for the Inline Promo `image` slot. 

Let me know how you would like to proceed.

bennypowers: I ran some experiments and it seems like the way forward is:

1. when the variant is inline-promo, the footer slot moves to becomes a child of the body container and loses it's margins.
2. the difference between a large inline promo and a large full-width promo layout is only a matter of container query.. i.e. a "full width" large inline promo is one that's larger than a given width.
3. a "standard" inline promo is one that has a body and a footer but no image or header.

Going this way, I believe it's possible that the only things users will have to change to get these patterns is the content, and in the case of the "standard" pattern, the color-palette attr, and even that we could enforce by requiring lighter bg color even for lightest when the content conditions for standard obtain.

bennypowers: @coreyvickery @marionnegp note that according to the approach i lay out above, users will have to apply page css in order to get the full width (i.e. borderless) pattern on a some screen sizes _in the case where the promo is full width within it's container, but not full page width_ i.e. there's a sidebar. Is that intended behavior?

bennypowers: consensus from office hours:
- it's `promo` not `inline-promo`
- `promo` is indeed a variant of card
- when promo doesn't have a heading or image, it appears like the "standard" type in the figma
- we'll use the approach above to align the content model of promo to card

bennypowers (approved):

.changeset/fair-buttons-punch.md
bennypowers: marionne's request for changes should block merge

marionnegp (changes requested):
- [x] Promo isn't displaying properly in the demo area. For example, the `Promo Standard` demo looks like this: 
Screenshot 2024-06-28 at 1 52 13 PM
It looks correct when I go to https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo-standard/ though!

- [x] Seeing some extra space between the image slot and the card border.
Screenshot 2024-06-28 at 1 48 25 PM

bennypowers:
WRT the demo area, that should resolve when we publish. It's expected that it won't work in the dp

marionnegp (changes requested):
If you have dev mode now, can you double check some of the spacing? If not, let me know, and I can get those specs to you.

adamjohnson:
@bennyp I noticed  you have a couple of [hard-coded](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-6231a58a58f5f5e718589c70803cca66ea1a31f4596c833d1334ed056bbc7444R54) [width values](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-b46e0f08bfba1e09875a41de2a00e6b67b12dfe9fba99f19747f1395d2e6bd7eR25) in the [demo CSS](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-6ae54541d0e883e3c3ed429e8de90d8723b14a2c8c46ca471324d4b727ec0e7cR43). These values are then [referenced in the container queries](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-4d26efc6c88db3bbcc6755c4b889ecb058b8f3f29aacfcaa8e1f5efbf397d7bcR265) in `rh-card.css`.

Component consumers are unlikely to bring the demo CSS with them. Can you explain your logic here?

If you remove the widths from the demo's, it breaks the design of all of the component demos except [Standard](https://deploy-preview-1569--red-hat-design-system.netlify.app/elements/card/demo/promo-standard/).

bennypowers:
> [hard-coded](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-6231a58a58f5f5e718589c70803cca66ea1a31f4596c833d1334ed056bbc7444R54) [width values](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-b46e0f08bfba1e09875a41de2a00e6b67b12dfe9fba99f19747f1395d2e6bd7eR25) ... [container queries](https://github.com/RedHat-UX/red-hat-design-system/pull/1569/files#diff-4d26efc6c88db3bbcc6755c4b889ecb058b8f3f29aacfcaa8e1f5efbf397d7bcR265) in `rh-card.css`.

The way I intended this to work was for the container size to dictate the layout of the promo. Therefore, in order for the demo to consistently display a given layout, the containing element for the promo needs to have the desired width.

> If you remove the widths from the demo's, it breaks the design

For example, here's `promo.html` with the container width removed:
image
and here's `promo-standard.html` with the container width removed:
image


I don't think this is broken, I think this works as intended - without the width on the container, the card takes up the maximum width of its container, in this case the whole width of the viewport. That is the definition of "full width" promo (as in the previous paragraph in this comment).
Does that track with your intentions @marionnegp and @adamjohnson ?

I did find some layout jank which occurred when removing the `display: grid` from the demo container element. I tracked that down to a container query mysteriously not applying for narrow promos. I fixed that by factoring out the mobile styles to the root of the stylesheet (i.e ."mobile first"), for that rule.

So after my fix, if you remove the `display: grid` from the demo `#container-width` element as well as it's width you get this:

image

Why doesn't this look like the full-width demo? because the full width demo's images are smaller (in accordance with the guidelines for full width.

So the user story for promo, as of this most recent commit, looks like this:

- Use the `variant=promo` attribute on card
  - Want it to be "standard"?
     - Don't include an image or header
     - **AND** abide by content guidelines vis-a-vis content length
  - Want it to be "narrow"?
    - Put it in a container smaller than mobile width
  - Want it to be "full width"?
    - Make it the width of the viewport
    - **AND** abide by content guidelines re: image size
  - Want the image to appear first?
    - Use the `::part(image)` in this pattern: `order: 0; justify-self: end`

marionnegp (changes requested):
- [ ] Standard promo in dark theme should use var(--rh-color-surface-darker, #1f1f1f)
- [ ] Can the image placeholder take up the whole image slot width, so it's easier to see how wide the image slot is?

marionnegp (approved):

PR #1571, opened 2024-05-16
chore: install esbuild for linux

bennypowers:
## What I did

1. install esbuild's linux binaries


package.json
```diff
@@ -265,6 +265,7 @@
     "@11ty/eleventy-plugin-rss": "^1.2.0",
     "@commitlint/cli": "^19.2.2",
     "@commitlint/config-conventional": "^19.2.2",
+    "@esbuild/linux-x64": "^0.21.3",
```
zeroedin, line 268: would we not want to do this in the optionalDeps?

zeroedin (approved):

PR #1581, opened 2024-05-30
fix: make context picker a radio group

bennypowers:
## What I did

1. made `<rh-context-picker>` a radio group
image

## Testing Instructions

1. check it out in dev server context demos
3. e.g. https://deploy-preview-1581--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/


bennypowers:
notes from standup: we feel justified merging this without a changeset

marionnegp (changes requested):
I think it would be worth adding a tooltip or something that shows the name of the surface color, so users would know how to reference the color palette/context.

marionnegp:
@bennypowers, once this is merged, this could be used within a non-demo page on ux dot, right? Like if we want to make a "Theme" section use this color picker instead of a static image?

lib/elements/rh-context-picker/rh-context-picker.ts
```diff
@@ -79,44 +68,25 @@ export class RhContextPicker extends LitElement {
...
-                 @input="${this.#onInput}">
-          <datalist id="palettes">${this.allow.map(palette => html`
-            <option id="option-${palette}"
-                    value="${palette}"
-                    title="${palette}"
-                    @click="${() => this.#setValue(palette)}">
-              <span class="visually-hidden">${palette}</span>
-            </option>`)}
-          </datalist>
+      <div id="container"
```
adamjohnson, line 78: Probably most semantically correct to wrap these inputs with `<fieldset>` and `<legend>` tags. [WebAIM example](https://webaim.org/techniques/forms/controls#radio).

adamjohnson (changes requested):

bennypowers:
> I think it would be worth adding a tooltip or something that shows the name of the surface color, so users would know how to reference the color palette/context.

I added a title attribute, is that enough?

bennypowers:
> @bennypowers, once this is merged, this could be used within a non-demo page on ux dot, right? Like if we want to make a "Theme" section use this color picker instead of a static image?

See the card pattern docs pr #1188

lib/elements/rh-context-picker/rh-context-picker.ts
bennypowers: to reiterate, it'd be 

```html
<form>
  <fieldset>
    <legend>Theming</legend>
    <label for="context">Palette</label>
    <rh-context-picker id="context"></rh-context-picker>
  </fieldset>
</form>
```
instead of

```html
<form>
  <rh-context-picker id="context">
    #shadow-root
      <fieldset>
        <legend>...</legend>
        ...
      </fieldset>
  </rh-context-picker>
</form>
```

Does that track?

adamjohnson: The first example is close. Here's what I suggest:

```html
<form>
  <fieldset>
    <legend>Color Palette</legend>
    <rh-context-picker></rh-context-picker>
  </fieldset>
</form>
```

👍

bennypowers: see if [3d76f52](https://github.com/RedHat-UX/red-hat-design-system/pull/1581/commits/3d76f52189bb747345b4cf01daeb9ec55153343b) helps

adamjohnson: Here's what I would suggest:

```diff
- <label for="radio-${palette}" slot="content">${palette}</label>
+ <span slot="content" class="capitalize">${palette}</span>
+ <label class="visually-hidden capitalize" for="radio-${palette}">${palette}</label>
```

For more context, here's the [entire render method](https://pastebin.com/QTYCB2FX).

Since the previous `<label slot="content">` was being hidden by `<rh-tooltip>`, I added a `<span>` that could take its place. 

This makes AT happy. Automated tests (Lighthouse, Axe, probably Lumar) won't surface false positive accessibility errors (they were before). Safari reads:

> Dark color palette, selected, radio button 2 of 6

Which is what we want ✔️ . 

Be sure to add `.capitalize { text-transform: capitalize; }` to the CSS file too.

adamjohnson: These changes look good. 👏 

As seen in Office Hours, the nested context picker still has some strange click/focus issues. [Video Demo (WebM)](https://github.com/RedHat-UX/red-hat-design-system/assets/575865/afe3d3d9-7046-49e5-990d-a5b19f7be7fb)

bennypowers: I'd like to spin that off into a separate issue. I suspect it's a problem with tooltip, rather than with context picker.

Repro: https://codepen.io/bennyp/pen/yLWPgew?editors=1000

adamjohnson: Perhaps I'm missing something, but tooltip in your repro appears to behave as expected.

adamjohnson: OK, yup. I see it now. It's definitely most apparent in Firefox. Thanks for clarifying.

Do you want to create and resolve that (separate) issue before resolving this conversation and merging this PR?

bennypowers: https://github.com/RedHat-UX/red-hat-design-system/issues/1618

bennypowers:
@adamjohnson I want to merge this anyways, and swing back to #1618, since I think it's unrelated to this feature. WDYT?

adamjohnson:
> @adamjohnson I want to merge this anyways, and swing back to #1618, since I think it's unrelated to this feature. WDYT?

Since the strange tooltip behavior only happens with the nested picker, I think we can merge this.

adamjohnson (approved):
Nested picker behaves strangely, but that can be addressed in https://github.com/RedHat-UX/red-hat-design-system/issues/1618.

Great work!

PR #1583, opened 2024-06-02
fix: elements/elements

### LINKED ISSUE #418: `/elements/elements`
With our package export maps, we can import with specifiers like:
```js
import '@rhds/elements/rh-footer/rh-global-footer.js';
```
but for CSS, we still have to:
```html
<link rel="stylesheet" href="https://unpkg.com/@rhds/elements/elements/rh-footer/rh-footer-lightdom.css" />
```

If we remove the export maps and copy everything in `/elements/*` to `/`, we wouldn't need to `/elements/elements` for assets.

BUT that would also break things like dev server and demo imports when running locally.

We could *keep* the export maps, and ship duplicates of all the files, but that's not ideal either

bennypowers:
## What I did

1. remove `elements/` prefix from CEM
2. copy lightdom assets to root

Closes #418 

Also fixes the import paths in the react wrappers, which are derived from the manifest


zeroedin (approved):
Let's get this moving

zeroedin:
> note that this moves the output files themselves. i think it's ok because we can change the export map, but wdyt?

I'm ok with that.

bennypowers:
I tested this out in the PoC we're working on and it seemed fine

PR #1552, opened 2024-05-03
fix(pagination): add dark theme support

adamjohnson:
## What I did

1. Added support for `<rh-pagination>` with dark themes / color contexts.

## Testing Instructions

1. View the color context demo in the deploy preview.

Issue #1457, opened 2024-01-23
[feat] `<rh-pagination>`: element updates and new pattern


marionnegp:
### Description

Feedback on current element:
- The current design looks heavy and too large on some pages, like Portal and Catalog.
- There should be an option for stacking the compact pagination with the "Go to page" input like in the screenshot below.
![Screenshot 2024-01-23 at 1 48 33 PM](https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/a2239190-c88f-4753-8a9a-878bf0d9bb64)
- There's also a need for the dropdown menu with options for changing the number of results per page. This is available in PatternFly.
<img width="438" alt="Screenshot 2024-01-23 at 1 46 25 PM" src="https://github.com/RedHat-UX/red-hat-design-system/assets/95588923/10f930ec-324e-45ef-9977-a4a706d6740c">

### Suggested solution

I'm proposing adjustments to the light theme colors, size adjustments, a compact variant, and a pagination + item per page selector. ([Figma link](https://www.figma.com/file/BVTQ5Fim2mLfstNGeZN1Qc/Pagination?type=design&node-id=0%3A1&mode=design&t=mTimrsdl2tizInMQ-1))

marionnegp:
@hellogreg @markcaron, should this go in Chansey or Charmander?

coreyvickery:
@markcaron @hellogreg I also would like to get this done in `Charmander`.

@zeroedin Could we also fold in [1393](https://github.com/RedHat-UX/red-hat-design-system/issues/1393) here too?

adamjohnson:
Looks like the color adjustments to the light theme were handled in #1393 (#1552).

I will look into implementing the `open`, `size` (small), and `compact` variants.

Should "Pagination + items per page selector" [use `<pf-dropdown>`](https://patternflyelements.org/components/dropdown/) or [wait for `<rh-dropdown>`](https://github.com/RedHat-UX/red-hat-design-system/issues/425)?

coreyvickery:
@adamjohnson Can we convert `pf-dropdown` (PF5) to `rh-dropdown` quickly? I think the styles are exactly the same.

cc @markcaron @zeroedin

adamjohnson:
I'm going to defer to @markcaron and @zeroedin's experience on this question from Corey:

> Can we convert pf-dropdown (PF5) to rh-dropdown quickly? I think the styles are exactly the same.

Issue #425, opened 2022-07-08
[feat] Create `<rh-dropdown-menu>` element


wesruv:
### Description

Make a dropdown that can accept links or buttons that has the correct a11y properties.

PFE-Dropdown was a solid attempt at this but the aria attributes are setup incorrectly, see: https://github.com/patternfly/patternfly-elements/issues/2054#issuecomment-1178799554

I don't know anything about Patternfly React's dropdown, but it looks like it's following the same design and has features I'd be interested in: https://www.patternfly.org/v4/components/dropdown

One pattern that PFE didn't cover that it looks like PF React Dropdown did was the dropdown should close if it loses focus, and you should only be able to have one open at a time.

### Acceptance Criteria

- [ ] Design done
- [ ] Development done

### Image

![image](https://user-images.githubusercontent.com/5607236/178032549-66cc417e-1ebd-4d8b-abfa-1058c1b7d49a.png)

### Link to design doc

https://www.patternfly.org/v4/components/dropdown/design-guidelines

wesruv:
I switched up the labels to what I think sounded right.

bennypowers:
See also https://github.com/patternfly/patternfly-elements/issues/2049

markcaron:
Blocked by PFE 3.0 release (https://github.com/patternfly/patternfly-elements/pull/2596)

hellogreg:
Exists as `rh-menu`

markcaron:
@adamjohnson @bennypowers @zeroedin, we should look into what's left to wrap up `<rh-menu>` to close this loop.

marionnegp:
@markcaron, since the design here is referencing PFv4, do we want to revisit it quickly to make sure it's still the design direction we want to use?

adamjohnson:
Indeed, I was just going to mention that we're missing a Figma for this component.

Also relevant for the design phase: `<rh-dropdown>` is likely made for freeform content (eg: primary nav dropdowns) whereas `<rh-menu>` follows the APG [Menu Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/):

> A [menu](https://w3c.github.io/aria/#menu) is a widget that offers a list of choices to the user, such as a set of actions or functions. Menu widgets behave like native operating system menus, such as the menus that pull down from the menubars commonly found at the top of many desktop application windows.

Looking at `<rh-menu>` right now, it looks like this:

[...]

PR #1587, opened 2024-06-04
feat(pagination): add open, small, and compact variants

adamjohnson:
## What I did

1. Added ~~three~~ two new attributes to pagination: `variant="open"` and `size="sm"`. Compact is now included by default in the main component between widths of 344px and 767px.

## Testing Instructions

1. Check out the deploy preview for the following variants:
    * [Open](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/open/)
    * [Open compact](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/open-compact/)
    * [Open compact small](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/open-compact-size/)
    * [Small](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/size/)
    * [Small compact](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/size-compact/)
    * [Compact](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/compact/)

## To do's

- [x] Update ux.redhat.com `rh-pagination` docs with the new variants/attributes
- [x] See [more docs to do's](https://github.com/RedHat-UX/red-hat-design-system/pull/1587#issuecomment-2197263029)
- [x] Change button dimensions for default size to 48px in Figma library to stick with token values.

## Notes to Reviewers

Relies on style changes made to pagination in #1552. Solves most of the to do's in #1457. 

### Pagination + items per page selector (blocked)

The [Figma spec](https://www.figma.com/file/BVTQ5Fim2mLfstNGeZN1Qc/Pagination?type=design&node-id=0%3A1&mode=design&t=mTimrsdl2tizInMQ-1) denotes another variant called "Pagination + items per page selector". This items per page selector relies on a dropdown menu. Since [rh-dropdown doesn't exist yet](https://github.com/RedHat-UX/red-hat-design-system/issues/425), that variant is blocked.


zeroedin:
> The [Figma spec](https://www.figma.com/file/BVTQ5Fim2mLfstNGeZN1Qc/Pagination?type=design&node-id=0%3A1&mode=design&t=mTimrsdl2tizInMQ-1) denotes another variant called "Pagination + items per page selector". This items per page selector relies on a dropdown menu. Since https://github.com/RedHat-UX/red-hat-design-system/issues/425, that variant is blocked.?

Let's roll that out into a new issue

elements/rh-pagination/rh-pagination.ts
```diff
@@ -123,32 +129,44 @@ export class RhPagination extends LitElement {
     const nextHref = this.#nextLink?.href;
     const lastHref = this.#currentLink === this.#lastLink ? undefined : this.#lastLink?.href;
     const currentPage = this.#currentPage.toString();
+    const numericContent = html`
```
zeroedin, line 132: I would consider breaking this out into its own function, this block should only render if compact is set.

```js
  #numericContent(currentPage: string, lastHref?: string, ) {
    return html`
      <div id="numeric" part="numeric">
        <span id="go-to-page" class="visually-hidden">
          <slot name="go-to-page">
              Go to page
          </slot>
        </span>
        <input inputmode="numeric"
            required
            min=1 max=${this.#links?.length ?? 1}
            aria-labelledby="go-to-page"
            @change=${this.#onChange}
            @keyup=${this.#onKeyup}
            .value=${currentPage}>
        <slot name="out-of">of</slot>
        <a href=${ifDefined(lastHref)}>${this.#links?.length}</a>
      </div>
    `;
  }
```

Then calling that in your method in `render()`:

```js
  render() {
    const { on = '' } = this;
    const { mobile, size } = this.#screen;
    const { dir } = this.#dir;
    const { label, labelFirst, labelPrevious, labelNext, labelLast } = this;
    const firstHref = this.#currentLink === this.#firstLink ? undefined : this.#firstLink?.href;
    const prevHref = this.#prevLink?.href;
    const nextHref = this.#nextLink?.href;
    const lastHref = this.#currentLink === this.#lastLink ? undefined : this.#lastLink?.href;
    const currentPage = this.#currentPage.toString();

    return html`
      <div id="container" class=${classMap({ mobile, [size as string]: true, [dir]: true, [on]: !!on })}>
        <a id="first" class="stepper" href=${ifDefined(firstHref)} ?inert=${!firstHref} aria-label=${labelFirst}>${L2}</a>
        <a id="prev" class="stepper" href=${ifDefined(prevHref)} ?inert=${!prevHref} aria-label=${labelPrevious}>${L1}</a>

        ${!this.variant?.includes('compact') ?
          html`
            <nav ?hidden=${mobile} ?inert=${mobile} aria-label=${label}>
              <slot></slot>
            </nav>
          `
          : html`
            ${this.#numericContent(currentPage, lastHref)}
          `}
        <a id="next" class="stepper" href=${ifDefined(nextHref)} ?inert=${!nextHref} aria-label=${labelNext}>${L1}</a>
        <a id="last" class="stepper" href=${ifDefined(lastHref)} ?inert=${!lastHref} aria-label=${labelLast}>${L2}</a>
        ${this.variant?.includes('compact') ? html`${this.#numericContent(currentPage, lastHref)}` : html``}
      </div>
    `;
  }
```

adamjohnson:
> Let's roll that out into a new issue

Issue created 👉 https://github.com/RedHat-UX/red-hat-design-system/issues/1625

elements/rh-pagination/rh-pagination.ts
adamjohnson: Updated in [e5a747d](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/e5a747d0f904968ac7fb3888fe049168f1fdb39b). 

Changed two things:

  1. The last `this.variant?...` should be `!this.variant?...` and now is.
  1. the `.visually-hidden` class in the `#numericContent` function needed to be conditionally applied.

I like this approach better than my previous one. Nice suggestion 👏

elements/rh-pagination/rh-pagination.ts
```diff
@@ -123,32 +129,45 @@ export class RhPagination extends LitElement {
...
     return html`
       <div id="container" class=${classMap({ mobile, [size as string]: true, [dir]: true, [on]: !!on })}>
         <a id="first" class="stepper" href=${ifDefined(firstHref)} ?inert=${!firstHref} aria-label=${labelFirst}>${L2}</a>
         <a id="prev" class="stepper" href=${ifDefined(prevHref)} ?inert=${!prevHref} aria-label=${labelPrevious}>${L1}</a>
 
-        <nav ?hidden=${mobile} ?inert=${mobile} aria-label=${label}>
-          <slot></slot>
-        </nav>
-
+        ${!this.variant?.includes('compact') ?
```
bennypowers, line 138: ok this is interesting...
this line introduces a new concept to the design system: multiple layered variants.

prior to this, variants were exclusive. this change means that an element can have 0-or-more variants at one time.

this opens up quite a bit of flexibility for users, however it could introduce explosive complexity and additional cognitive overhead. 

- "Do these two variants work together?"
- "Does this element support multiple variants?"
- "Why doesn't adding this variant to this other one work the way i thought it would?"
- "How do I validate every possible combination of variants on this element?"

Before introducing a broad-ranging concept like this, it's worth considering the knock-on effects.

Alternative approaches could be:
- `variant: 'open'|'compact'|'open-compact'|null = null`. we have precedent for this in `inline-promo` and in audio-player
- making `open` or `compact` a boolean attribute, but beware: 'open' is overloaded
 
@coreyvickery @marionnegp what do you think about that?

bennypowers (changes requested):

elements/rh-pagination/rh-pagination.ts
marionnegp: Would setting an element-specific token allow us to prevent people from using other background colors?

I would think that the first alternative approach would have been the easiest for engineers to use because the variant options are more apparent. Would going in that direction make the design system harder to maintain?

adamjohnson: I'd be into investigating `variant: 'open'|'compact'|'open-compact'|null = null`, especially if we think people might go wild assigning background colors that could be inaccessible.

marionnegp (changes requested):
- [ ] Box/default variant doesn't work in the `lighter` palette
Screenshot 2024-06-21 at 1 38 44 PM

- [ ] Active state of open variant should have a transparent background color
Screenshot 2024-06-21 at 1 39 42 PM

- [ ] This might be fixed in a separate PR, but can the "Go to page" text before the page input be shortened to just "Page"?

elements/rh-pagination/rh-pagination.css
```diff
@@ -209,3 +227,39 @@ input:invalid {
...
+
+:host([variant~="compact"]) #prev {
+  margin-inline-end: var(--rh-space-lg, 16px);
+}
+
+:host([variant~="compact"]) #numeric {
+  flex: 0;
+  margin-block: 0;
+  margin-inline-end: var(--rh-space-lg, 16px);
+  margin-inline-start: 0;
```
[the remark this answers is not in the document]
adamjohnson: Added in [8ca41fe](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/8ca41fe0edabf2ac0de0eb9747572afa069087c0).

adamjohnson:
## Color palette updates

Dark | Darker | Light | Lighter fixed in [5746c45](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/5746c453d609a0567b228a6027f9dc2ebaabeca9).

## Go to page

> This might be fixed in a separate PR, but can the "Go to page" text before the page input be shortened to just "Page"?

This is customizable via the [`go-to-page` slot](https://github.com/RedHat-UX/red-hat-design-system/blob/main/elements/rh-pagination/rh-pagination.ts#L133).

Right now, the default content is "Go to page". The docs have several screenshots that say, "Go to page". 

@marionnegp I can change it to "Page" if you want to update the docs screenshots.

marionnegp:
@adamjohnson, I can send updated docs content and images next week.

elements/rh-pagination/rh-pagination.ts
marionnegp: @bennypowers, re: making `open` into a pattern: The placement of the blue border in the interaction states is different in `open`, so setting an element-specific token will be more involved than changing a background color.

adamjohnson:
@marionnegp Changed "Go to page" to just "Page" in [69e8d3f](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/69e8d3fa47bd2d89bb94d62c8c7d2c510517b942). The slot is still called `go-to-page` to keep backwards compatibility.

- [ ] Update docs to reflect the change from "Go to page" to "Page".

elements/rh-pagination/rh-pagination.ts
bennypowers: so is the conclusion that we want to allow elements to have multiple, stacking variants?

adamjohnson: No. There would be a single variant value: 

```html
variant="open"
variant="compact"
variant="open-compact" 
...
```
...etc

and never: 

```html
variant="open compact"
```

bennypowers: Ok that's good

I would still like to expend some effort to see if we can factor out one of those into a pattern out and css vars

adamjohnson: Added `variant="open-compact"` in ab591d1.

<hr>

pagination-sm-vs-compact

It might be feasible to make the `size="sm"` property respond to CSS vars. That said, design control goes to the end user, which may be undesirable. FWIW, those variables are currently private.

As Marionne said, `open`, `compact`, and `open-compact` are different because they are doing more than changing one CSS variable:

  * `open` removes the border, background color, and changes the active paginator border from top to bottom in both shadowdom and lightdom CSS.
  * `compact` changes the DOM so that `#numeric` appears between `#prev` and `#next`.
  * `open-compact` combines the two ☝️ 

@bennypowers if you have any other ideas re: pattern/vars, please let me know. I'm open to exploring different approaches.

bennypowers: @marionnegp is there ever a case where pagination should present as "Compact" as in the above screenshot, even though there is enough horizontal whitespace available? In other words, is "compact" something that must be the choice of the page author, rather than a consequence of the available space for the element (via container query)?

If it were the case that "compact" pagination is something that happens **only** when the element doesn't have enough horizontal space within which to render at least seven buttons in the button group, then we could reduce the api surface to the following:

- `<rh-pagination>`
- `<rh-pagination size="sm">`
- `<rh-pagination variant="open">`
- `<rh-pagination size="sm" variant="open">`

IMO that would be much easier to handle for both maintainers and users (i.e. page authors). Note that the question of whether-or-not to display the numeric "Go to page" input is not directly relevant, because that control can always simply wrap to the next line, regardless of the container size.

marionnegp: @bennypowers, was your question answered adequately during today's office hours?

bennypowers: Yes. Summary: compact variant should be replaced with a container query, and we should document how to force it into compact layout by putting it in a narrow container

adamjohnson: This is a comparison to how `rh-pagination` looks on mobile currently versus making `compact` the default. Screenshot taken at 320px wide (ala [WCAG SC 1.4.10](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html)).

pagination mobile screenshot

At 576px+, paginator numbers show up in the current implementation.

`compact` needs roughly 392px before everything fits on one line. FWIW, the [compact demo](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/compact/) does fit on one line on my relatively narrow-screened Pixel 6a. It's probably worth opening the demo page on your device.

I just want to make sure everyone is on-board with these changes. I'm happy to implement them. Are folks concerned with design/layout of this component with screen sizes less than 392px...?

bennypowers: I would suggest  two container sizes: one with enough space to fit the numeric input in between the steppers, and a narrower layout where the numeric wraps around the inline-end direction

elements/rh-pagination/demo/compact.html
bennypowers, line 1: Could we please split this into multiple files, one per element, no text content?

elements/rh-pagination/rh-pagination.ts
adamjohnson: Resolved in [f243293](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/f2432935e4cede283026b95fb33e7a20f73dfe28). Here's what the standard `rh-pagination` now looks like on mobile, mobile-tablet ("compact"), and desktop:

Screenshot of rh-pagination at different viewports

Play with the [DP Demo](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/) to see it in action for yourself.

@bennypowers great suggestion integrating "compact" into the base component. 👏

elements/rh-pagination/demo/compact.html
adamjohnson: Resolved in [a0dca8b](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/a0dca8b69461288a69d19b29946794a5e3bf5ece).

bennypowers (approved):

marionnegp (changes requested):
- [x] Arrow icons when they're not disabled are `var(--rh-color-gray-50, #707070)` in the designs. Is it #4d4d4d for better color contrast?
- [x] Field input padding should be `var(--rh-space-sm, 6px)` for block, `var(--rh-space-md, 8px)` for inline.
- [x] Font size should be `var(--rh-font-size-code-md, 1rem)` for numbers in small pagination
- [ ] In the [small demo](https://deploy-preview-1587--red-hat-design-system.netlify.app/elements/pagination/demo/size/), the blue border for selected state looks thinner than 3px. Might be due to the `inset-block-start`? It looks correct in the color context demo though.
Screenshot 2024-07-19 at 10 50 30 AM

Some changes to the button bg based on the color context demo for the box version: 
- [ ] Bg in lightest and lighter context should be #f2f2f2. (It was originally #e0e0e0, but we got feedback that it looked a little too dark.)
- [ ] Bg in lighter context should stay #e0e0e0.
- [ ] Bg in dark, darker, and darkest contexts should be #292929. 
- [ ] Bg in dark-alt context should be #383838.

- Will users be able to choose the alignment of the pagination buttons and input field themselves?

Let me know if you have any questions!

adamjohnson:
@marionnegp:

> In the small demo, the blue border for selected state looks thinner than 3px. Might be due to the inset-block-start? It looks correct in the color context demo though.

Yeah, there is a -1px `inset-block-start` value there due to the focus state design. Without it, the square edges of the active border show through the rounded edges of the focus state.

FWIW, the computed state is `3px` for that border. Let me know if that should change.

## Color context demo / colors

Since this component has specific colors for light, lighter, lightest, dark, darker, darkest, when using the color context demo, you have to always go from light to dark to light or vice versa. Eg click:

> Darkest > Light > Dark
> Lightest > Darker > Lighter

The following will not work:

> Dark > Darkest > Darker
> Lightest > Lighter > Light

It's a limitation of our current color theme implementation. /shrug

With this in mind, what background color should lighter be?

>  Bg in lightest and **lighter** context should be #f2f2f2. (It was originally #e0e0e0, but we got feedback that it looked a little too dark.)
>  Bg in **lighter** context should stay #e0e0e0.

Also, the "dark-alt context" the same as the "darker context"?

I have a feeling some of the background color feedback from above may be moot after realizing the strange behavior of the Color Palette picker for this component (it got me, too!).

## Alignment

> Will users be able to choose the alignment of the pagination buttons and input field themselves?

Yes. You can now use `justify-content` on the `container` or `numeric` part to align elements as needed.

elements/rh-pagination/rh-pagination.ts
```diff
@@ -113,46 +118,69 @@ export class RhPagination extends LitElement {
...
+  #internalColorPalette?: string | null;
+
+  protected willUpdate(): void {
+    /**
+     * TEMPORARY: this fixes the need to access the parents color-palette in order to get the `lightest`
+     * value. This fix will only update the component when switching between light and dark themes as
+     * thats when the consumer requests an update. Switching between lighter -> light for example will
+     * not trigger the component to update at this time. Related: #1395.
+     */
+    this.#internalColorPalette = this.closest('[color-palette]')?.getAttribute('color-palette');
```
zeroedin, line 130: @bennypowers we stripped this idea out of switch, it still exists in table.  What are your thoughts here?

marionnegp:
@adamjohnson, all the light theme colors look correct now that I know about the tricky color context picker, so I marked those to do items as done.

Yep, dark-alt should look like darker. I went ahead and updated the pagination file because I realized I hadn't mocked pagination up in the darker context. 🤦🏽‍♀️

marionnegp (approved):
I realized that `dark-alt` isn't available as a color context AFAIK.

elements/rh-pagination/rh-pagination.css
```diff
@@ -1,28 +1,41 @@
 :host {
+  --_stepper-size: 48px;
+
   display: block;
   min-height: 4em;
 }
 
+:host([size="sm"]) {
+  --_stepper-size: 34px;
```
marionnegp, line 9: @adamjohnson, can the stepper size for the small variant be changed to `32px`, so it matches with our length token sizes. This can use `var(--rh-length-2xl, 32px)` if you'd like.

adamjohnson: Resolved in [47c3440](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/47c3440d6980140563b64f6b6903459ee08d36e5).

markcaron:
@adamjohnson @zeroedin @bennypowers, we're waiting on some doc updates for the new variants. I'd hate to hold up the show for docs. We can fast-follow if needed.

CC: @coreyvickery @marionnegp

elements/rh-pagination/docs/20-guidelines.md
```diff
@@ -68,116 +45,166 @@ In some edge cases, pagination may need to display a very low page count like on
...
+  <uxdot-best-practice dont>
+    <uxdot-example slot="image" width-adjustment="372px">
+      <img src="../pagination-guidelines-best-practices-3-dont.svg" alt="Truncated pagination with the input field on top and right aligned.">
+    </uxdot-example>
+    <p>Do not change the order or alignment of the page input field.</p>
+  </uxdot-best-practice>
+</div>
+
+
+### Compact vs. mobile
```
marionnegp, line 195: Is it possible to remove the top padding for these two Do/Don't images?

marionnegp (changes requested):

elements/rh-pagination/rh-pagination.ts
bennypowers: Ok the issue here is the same one we discussed on a call a ways back

Because we have not yet fully developed our theming process yet, it seems to me that shipping this will actually break theming going forward. 

Imagine that the user implements a bordeaux-and-rosewater colour theme for their page, then adds a paginator on a `lighter` container - they'll get paginator colours that assume stock surface colours, and it won't look as expected on the `lighter` rosewater surface colour.
 
For that reason may I suggest that we limit pagination to being used on `lightest` and `darkest` containers only for the time being, then circle back when we work on theming to address `lighter` `light` `dark` and `darker`?

Thanks to @adamjohnson for talking this out with me earlier today.

elements/rh-pagination/docs/20-guidelines.md
adamjohnson: Resolved in [b6308e3](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/b6308e3acd5d4e3737d2a0a940decf28f39ca1a5).

elements/rh-pagination/rh-pagination.ts
adamjohnson: Talked about this with @marionnegp. She's okay with this. So, I went ahead and removed light/lighter/dark/darker colors (and docs) so that we are just sticking with the normal darkest ("`.dark`")/lightest color palette.

Resolved in [63a2749](https://github.com/RedHat-UX/red-hat-design-system/pull/1587/commits/63a274962a434280c12d996b0e66399952977fc5).

FWIW, the only situation where the background color of the box variant bleeds into the surface background color is "Lighter" in the color context demo. I added a note in the docs suggesting consumers place Paginations on lightest and darkest surfaces to avoid confusion.

PR #1592, opened 2024-06-06
docs(tabs): link to tab pattern

bennypowers:
## What I did

1. add a pattern page for tabs
2. stub a pattern image for tabs (cc @coreyvickery)
3. write up a link-to-tab pattern

## Testing Instructions

1. view the [pattern page](https://deploy-preview-1592--red-hat-design-system.netlify.app/patterns/tabs/)
2. check that hash navigation to tabs or tab contents works

## Notes to Reviewers
cc @markcaron this pattern was requested by Lindsay Mitchel


zeroedin:
Should we provide the explicitly used as a codeblock on the pattern page itself?

bennypowers:
> Should we provide the explicitly used `html` / `js` code as a codeblock on the pattern page itself? Do patterns pages ultimately get a code tab as well long term?

yes, but maybe not in this pr

docs/patterns/tabs/index.md
```diff
@@ -0,0 +1,39 @@
...
+within the tab panel, or to the tab itself.
+
+{% alert state="warning" %}
+Use this pattern sparingly. If your tabs serve only as page navigation,
+use the [Subnav](/elements/subnavigation) element instead.
+{% endalert %}
+
+{% include './patterns/link-to-tab.html' %}
+
+<nav id="simulate-nav">
```
zeroedin, line 27: needs an aria-label

bennypowers: is a nested heading acceptable? cc @adamjohnson

bennypowers:
OK @zeroedin PTAL

@adamjohnson I noticed an unrelated issue: I wasn't able to tab to the "Best" link in the "West" tab panel on firefox, fedora 40, with orca screenreader on (or off). Instead, I tabbed to the tab panel itself, which I think should not be in the tab order. Can you confirm? If so, let's open a new issue, or maybe this is already addressed elsewhere?

zeroedin:
> wasn't able to tab to the "Best" link in the "West" tab panel on firefox

the link doesn't have a `href` on it.

adamjohnson:
> I tabbed to the tab panel itself, which I think should not be in the tab order. Can you confirm?

From ARIA [Authoring Practices Guide Tabs](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction) docs (Note `#4`):

> When the tabpanel does not contain any focusable elements or the first element with content is not focusable, the tabpanel should set tabindex="0" to include it in the tab sequence of the page.

So, in this case, this tab is behaving as expected.

zeroedin (approved):

PR #1603, opened 2024-06-10
fix(tabs): fix nested interactive controls

### LINKED ISSUE #1585: [bug] `<rh-tabs>` per axeDevTools: Interactive controls must not be nested
@rmscampos is getting the following issue on the `<rh-tabs>` [demo](https://ux.redhat.com/elements/tabs/demo/):

```Interactive controls must not be nested```

**Screenshot:**
image

### PR DESCRIPTION
**Pull request by zeroedin**
## What I did

1. Corrects nested controls for `<rh-tab>`.  No longer embeds a `<button>` element.  This same fix was made upstream prior in `pf-tabs`.

Closes #1585

## Testing Instructions

1. View [deploy preview docs demo](https://deploy-preview-1603--red-hat-design-system.netlify.app/elements/tabs/demo/)

## Notes to Reviewers

Check automated accessibility tests no longer fail on nested interactive controls.

### DIFF
--- a/elements/rh-tabs/rh-tab.css
+++ b/elements/rh-tabs/rh-tab.css
@@ -19,7 +19,7 @@ slot[name="icon"] {
   display: block;
 }
 
-button {
+#button {
   margin: 0;
   font-family: inherit;
   font-size: 100%;
@@ -40,8 +40,8 @@ button {
   max-height: 100%;
 }
 
-button:before,
-button:after {
+#button:before,
+#button:after {
   position: absolute;
   inset: 0;
   content: "";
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: elements/rh-tabs/rh-tab.ts [hunk 1]
```diff
@@ -34,7 +34,7 @@ export class TabExpandEvent extends Event {
  * @slot icon - Can contain an `<svg>` or `<pf-icon>`
  * @slot - Tab title text
  *
- * @csspart button - `<button>` element
+ * @csspart button - `<div>` element
```
bennypowers, line 37:
```suggestion
 * @csspart button - element that contains the interactive part of a tab 
```

elements/rh-tabs/rh-tab.ts
```diff
@@ -92,13 +89,13 @@ export class RhTab extends LitElement {
     const last = lastTab === this;
     return html`
       <div id="container" class="${classMap({ active, box, vertical, first, last, [on]: !!on })}">
-        <button part="button" ?disabled="${this.disabled}">
+        <div id="button" part="button" ?disabled="${this.disabled}">
```
bennypowers, line 92: I wonder if we could combine this with the container element?

bennypowers (changes requested):
One small docs change

elements/rh-tabs/rh-tab.ts
zeroedin: when you get the opportunity PTAL again!

elements/rh-tabs/rh-tab.css
```diff
@@ -104,64 +122,32 @@ button:after {
 }
 
 /* Context switch on attributes */
-:host([active]) #container,
-#container.active {
+#button.active {
```
adamjohnson, line 125: The [APG recommends](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/) using [`aria-selected="true"` and `aria-selected="false"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected) to communicate to AT which tab is active. That said, that change might be out of scope for this PR.

adamjohnson (approved):
Logic Gives Tabs Meaning 👏

elements/rh-tabs/rh-tab.css
zeroedin: This definitely was a regression, good catch!  Resolved in https://github.com/RedHat-UX/red-hat-design-system/pull/1603/commits/87e038b5210301ad01012f0a568350d315909831

bennypowers: Can you add an ax tree test for that using a11ySnapshot?

bennypowers (approved):
last good town meeting

PR #1608, opened 2024-06-11
feat(cta)!: href

### LINKED ISSUE #1607: [feat] `<rh-cta href="">`
### Description

Allow cta to take an href attr.

### Suggested solution

render the link in shadow root, adjust css and necessary

### Example API

```HTML
<rh-cta href="/">Home</rh-cta>
```

### Additional context

 see https://github.com/orgs/RedHat-UX/discussions/1594#discussioncomment-9728818

bennypowers:
Closes #1607 

## What I did

1. add `href` attr to cta
2. adjust css to work with shadow anchors
3. rewrite analytics demo to better represent actual usage and ensure that shadow link works as expected
4. remove deprecated apis

## Testing Instructions

1. unit tests
2. thoroughly review deploy preview demos, ensure they work as expected

## Notes to Reviewers

**NOTE WELL** that this PR removes the already-deprecated `color-palette` attr, and marks the `cta` property as deprecated.


zeroedin (approved):
Links generate traffic metrics

elements/rh-cta/rh-cta.ts
```diff
@@ -2,54 +2,36 @@ import { LitElement, html } from 'lit';
...
 import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
 
 import { DirController } from '../../lib/DirController.js';
 
-import { type ColorPalette } from '../../lib/context/color/provider.js';
 import { colorContextConsumer, type ColorTheme } from '../../lib/context/color/consumer.js';
 
 import style from './rh-cta.css';
 
 import '@patternfly/elements/pf-icon/pf-icon.js';
```
zhawkins, line 15: Can we dynamically import `pf-icon` if it's needed based on the attributes provided?

bennypowers:
@paceaux brings up potential problems with web scraping
@OneEightyFirst was also interested

elements/rh-cta/rh-cta.ts
bennypowers: FOUC?

bennypowers: i suppose we could delay first render until icon is registered

zhawkins: There's going to be FOUC no matter what with our current approach. 

Since it's already going to take longer to load if we're using pf-icon (vs the inline SVG that's there for the normal CTA styles), might as well dynamically import the JS if it's needed?

bennypowers: https://github.com/RedHat-UX/red-hat-design-system/pull/1634

bennypowers:
@heyMP can you take a look at the [analytics demo](https://deploy-preview-1608--red-hat-design-system.netlify.app/elements/call-to-action/demo/analytics/) ([source](https://github.com/RedHat-UX/red-hat-design-system/blob/fb8584a2bea096310914a0216ff6c2f3e11c9352/elements/rh-cta/demo/analytics.html)) and let me know if I'm headed in the right direction? Does this comport with the current practice?

bennypowers:
Leo Zavala says:

> The changes for `rh-cta` are aligned with the new approach we are taking with EDDL; using `data-analytics-*` attributes. For anyone who is using the new approach for analytics the update [to `href`] would be seamless.
> One thing to note is we are providing a fallback for those that have not implemented the new approach. The fallback would not work w/the new way of using the `href` attribute on `rh-cta.` It seems like a minor change to [this function in the C2EDDL](https://github.com/searchdiscovery/redhat-datalayer/blob/main/ClicktoEDDL/src/AdobeLaunchClicktoEDDL.js#L574) script will allow for the fallback to continue working. (Though it may be unnecessary if teams are adopting the new approach as they update)

marionnegp (commented):
- [ ] The default video variant is missing the little play icon at the end.
- [ ] Should the brick icon version have an icon in the demos?

Slightly off-topic for this PR, but I'm seeing this for the dark palette CTAs in the no javascript demo. Is it what I'm supposed to see?
Screenshot 2024-06-20 at 2 20 43 PM

bennypowers:
> * [ ]  The default video variant is missing the little play icon at the end.
> 
>     * [ ]  Should the brick icon version have an icon in the demos?
> 
> 
> Slightly off-topic for this PR, but I'm seeing this for the dark palette CTAs in the no javascript demo. Is it what I'm supposed to see? Screenshot 2024-06-20 at 2 20 43 PM

i saw this in another job. it's the wrong icon name.

bennypowers:
I want to update the docs to use `href` attr for ctas, but i would rather do that after making sure ssr works for cta.

and that seems out of scope for this PR so let's push it off to next cycle.

bennypowers:
> > Slightly off-topic for this PR, but I'm seeing this for the dark palette CTAs in the no javascript demo. Is it what I'm supposed to see? Screenshot 2024-06-20 at 2 20 43 PM
> 
> same on main https://ux.redhat.com/elements/call-to-action/demo/no-javascript-customization/

https://github.com/RedHat-UX/red-hat-design-system/pull/1654

PR #1613, opened 2024-06-12
docs: elements > code > accordion + table layout tweaks

adamjohnson:
## What I did

1. Tweaked the design of the table on the code tab. See #1606 for background.
    * Removed some padding, changed a background color, and a few border colors

## Testing Instructions

1. Go to Elements > {any element} > Code and scroll down to the table. Compare changes to ux.redhat.com.
    * [DP of Call To Action Code tab](https://deploy-preview-1613--red-hat-design-system.netlify.app/elements/call-to-action/code/#rh-cta)


docs/styles/styles.css
```diff
@@ -303,6 +303,32 @@
...
+    border-block-end: 1px solid var(--rh-color-border-strong-on-light, #151515) !important;
+  }
+
+  .code-accordion rh-accordion-panel[expanded] rh-table tbody tr:last-child {
+    border-block-end: 0 !important;
+  }
+
+  .code-accordion rh-accordion-panel[expanded] rh-table :is(th, td) p {
+    margin-block: 0;
+  }
```
zeroedin, line 331: If these are page specific (code tab) I would suggest creating a `/styles/pages/{{page}}.css` and linking that file to those pages and move these styles there instead of creating a non global class.  This way we keep styles compartmentalized and don't start down the path of classitis again.

zeroedin: The `uxdot-` component prolly isn't a bad idea that way no need for the additional page css, and its just composition so easy peasy dsd.  @adamjohnson ping me if you want to pair on that.

adamjohnson: Steve and I paired on this—trying to turn these changes into a `<uxdot...>` component. 

Since the styles rely on descendant selectors, it was actually more code to create a custom `<uxdot...>` component with a lightdom CSS file versus moving these styles into their own stylesheet.

bennypowers: what about the whitespace notes from #1606 ?

adamjohnson: > what about the whitespace notes from #1606?

Can you elaborate on this a little?

I purposefully added the background color and kept padding to clearly denote the [parent/child relationship](https://github.com/RedHat-UX/red-hat-design-system/assets/575865/cc671f77-2378-4112-baf2-8abbe8de808c). 

From a pure design perspective, I like your [flush left version](https://github.com/RedHat-UX/red-hat-design-system/assets/575865/400125e1-9c6f-4637-9442-b5cb0d3fbe44), but I think the UX is less clear because the `<thead>` looks nearly identical to an `<rh-accordion-header>`.

zeroedin: I agree with @adamjohnson on this point.  I think the table needs padding in the panel to separate itself visually from the panel.  @marionnegp @coreyvickery what are your thoughts on this?

coreyvickery: @zeroedin @adamjohnson Looks good to me.

adamjohnson:
Resolved a merge conflict and had to move the CSS to the new layout (`docs/_includes/layouts/pages/elements.njk`). 

Nothing has really changed but need someone to give the 👍 to merge.

👀 Code tab > table in the [deploy preview](https://deploy-preview-1613--red-hat-design-system.netlify.app/elements/call-to-action/code/#rh-cta)

zeroedin (approved):
This looks good to me!

PR #1622, opened 2024-06-13
fix(footer): add additional slot to support rh-site-status composition

### LINKED ISSUE #1620: [bug]: `rh-footer` does not have a slot that currently supports `rh-site-status`
If you try and add `rh-site-status` to footer there is no simple way of doing so without overriding footer's CSS pretty dramatically. An optional additional slot in `rh-footer` would be preferable to easily compose `rh-site-status` inside of footer.

References #1614

zeroedin:
## What I did

1. Added `main-tertiary` slot to support `rh-site-status` composition.
2. Added SlotsController in order to place `empty` class and display none this new class when there is nothing slotted in the new `main-tertiary` slot.
3. Added demo showing `rh-site-status` composition using this new slot

Closes #1620 

## Testing Instructions

1. [ View Deploy Preview](https://deploy-preview-1622--red-hat-design-system.netlify.app/elements/footer/demo/with-site-status/)


zeroedin:
Closing.  After further evaluation, we've decided to compose site-status using the current footer API workarounds: Example: https://codepen.io/zeroedin/pen/rNgYgJq

PR #1630, opened 2024-06-16
feat!: generate entrypoint module

bennypowers:
removes `rhds.min.js` from the npm tarbal in favour of `elements.js`

## What I did

1. remove the "bundle" step (the bundle script is still available as a helper for users that want to generate a bundle)
2. add an entrypoint file `elements.js`, which re-exports everything from the elements dirs.

## Testing Instructions

1. try running the bundle script - it shouldn't crash

## Notes to Reviewers
What happens if we ever produce a name collision? Should we have some automated tests for that?


heyMP:
Checking our repos, it doesn't look like we use the previous `rhds.min.js` outside of some demos.  I think it's a good change. 👍

zhawkins:
To confirm, we don't use `rhds.min.js` file 👍

zeroedin (approved):
Linking Globally To Modules

PR #1634, opened 2024-06-17
perf(cta): lazy-load icon dependency

bennypowers:
## What I did

1. lazy load the `pf-icon` dep in cta


zeroedin (approved):
Leaping goats traverse mountains

zhawkins:
Nice! TY!

PR #1654, opened 2024-06-21
docs: load lightdom css on demo pages

bennypowers:
## What I did

1. account for `-lightdom-shim.css` in the 11ty build

## Testing Instructions

1. [check dp /elements/call-to-action/demo/no-cta-javascript/
](https://deploy-preview-1654--red-hat-design-system.netlify.app/elements/call-to-action/demo/no-cta-javascript/)


marionnegp (commented):
CTA demo looks good for light palettes. Is the dark supposed to look like this in the demo?
Screenshot 2024-06-21 at 2 54 08 PM

Dark versions look good when I change the page's bg color though.
Screenshot 2024-06-21 at 2 57 19 PM

bennypowers:
@marionnegp i think it's a problem with the cta lightdom shim approach, and one of the reasons we should prefer ssr. I may be mistaken here, @zeroedin can perhaps provide clarity.

zeroedin:
That demo appears to be not using `<rh-surface>` but `<rh-context-provider>` will update.

PR #1656, opened 2024-06-21
fix(navigation-secondary): improved accessibility customization

### LINKED ISSUE #1655: `<rh-navigation-secondary>`:  aria-label on the nav in the shadowroot should be customizable
For other elements we've used `accessible-label` property to set and customize labels such as this, this would also be an improvement for translations.

zeroedin:
## What I did

1. applied navigation role to host using internals
2. removed shadow nav element

Closes #1655

## Testing Instructions

1. View deploy preview, change `aria-label` on host to a different value, that value should update in accessibility tree, and removing it will change it back to "secondary"

### DIFF
--- a/.changeset/strong-snails-reflect.md
+++ b/.changeset/strong-snails-reflect.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-navigation-secondary>`: improved accessibility'
--- a/.changeset/strong-snails-reflect.md
+++ b/.changeset/strong-snails-reflect.md
@@ -2,4 +2,4 @@
 "@rhds/elements": patch
 ---
 
-`<rh-navigation-secondary>`: improved accessibility'
+`<rh-navigation-secondary>`: improved accessibility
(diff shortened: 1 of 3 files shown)


adamjohnson (approved):
Labels Guide Tidy Menus 👏

.changeset/strong-snails-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-navigation-secondary>`: improved accessibility'
```
bennypowers, line 5:
```suggestion
`<rh-navigation-secondary>`: improved accessibility
```

elements/rh-navigation-secondary/rh-navigation-secondary.ts
```diff
@@ -389,11 +386,8 @@ export class RhNavigationSecondary extends LitElement {
     this.removeAttribute('role');
     // remove aria-labelledby from slotted `<ul>` on upgrade
     this.querySelector(':is([slot="nav"]):is(ul)')?.removeAttribute('aria-labelledby');
-    // transfer the aria-label to the shadow <nav>
-    if (this.hasAttribute('aria-label')) {
-      this.#label = this.getAttribute('aria-label') ?? 'secondary';
-      this.removeAttribute('aria-label');
-    }
+    // if the accessibleLabel attr is undefined, check aria-label if undefined use default
+    this.#internals.ariaLabel = this.getAttribute('aria-label') ?? 'secondary';
```
bennypowers, line 390: We shouldn't need this. Internals are defaults which can be overridden by host attrs

At most, we may choose to provide an `accessible-label` attr, for the sake of congruency with other elements. That attr would set the internals property, but would still be overridden by aria-label on host

zeroedin: makes sense, and fixed, thanks

PR #1669, opened 2024-06-26
docs: fix whitespace in repo status rh-tags

zeroedin:
## What I did

1. Fixed whitespace in repostatus that causes markdown to convert it to `<p>` tags when they shouldn't exist.

## Testing Instructions

1.


markcaron (approved):
Little gnome tinkers machines

Issue #1676, opened 2024-06-26
[bug] `<rh-stat>` broken css selector in chrome


zeroedin:
### Describe the bug

View https://codepen.io/zeroedin/pen/zYQbJRr in Chrome version 126.0.6478.116 and 126.0.6478.127 (only versions tested so far) 

```
:not(.hasCta) #cta, :not(.hasIcon) #icon, :not(.hasStatistic) #statistic, :not(.hasTitle) #title {
    display: none;
}
```

The above selector does not function correctly in chrome and will ignore the `:not()` it appears and applies the `display: none` do to `#statistic`

Same selector appears to work in Firefox and Safari

```
div:not(.hasCta) #cta, div:not(.hasIcon) #icon, div:not(.hasStatistic) #statistic, div:not(.hasTitle) #title {
    display: none;
}
```
or 
```
:not(.hasCta) > #cta, :not(.hasIcon) > #icon, :not(.hasStatistic) > #statistic, :not(.hasTitle) > #title {
    display: none;
}
```

both work, the div obviously applies the div with not class, and  `>` applies the same with the direct parent being called for the id.

### Screenshots

![image](https://github.com/RedHat-UX/red-hat-design-system/assets/13666/7ffbad24-6049-451d-804b-d8d9270906ee)

### Steps to reproduce

https://codepen.io/zeroedin/pen/zYQbJRr

### Expected behaviour

The slotted stat to display

### Operating System (OS)

macOS

### Browser

Chrome

bennypowers:
Please reopen for cubone if it is still relevant

PR #1678, opened 2024-06-27
docs: various demo fixes

bennypowers:
## What I did

1. exclude rh-card from heading style selectors globally

## Testing Instructions

1. check card demos
2. check heading usage


zeroedin:
Looks like this same issue in chrome will applys here as well https://github.com/RedHat-UX/red-hat-design-system/issues/1676

elements/rh-accordion/demo/color-context.html
```diff
@@ -42,6 +42,13 @@ <h3>Item Three</h3>
   </rh-accordion>
 </rh-context-demo>
 
+<style>
```
zeroedin, line 45: this doesn't seem to work 


Maybe this change to the `rh-context-demo`
```
#provider {
    padding: var(--rh-context-demo-padding, var(--rh-space-xl, 24px));
    height: auto;
    min-height: 100%;
}
```

docs/styles/typography.css
```diff
@@ -4,50 +4,50 @@ body {
   line-height: var(--rh-line-height-body-text, 1.5);
 }
 
-:where(h1, h2, h3, h4, h5, h6) {
+:where(h1, h2, h3, h4, h5, h6):not([slot]) {
```
zeroedin, line 7: I have questions about this approach.
 
Isn't it the components responsibility to ensure the correct h1-h6 styles are applied?  
Shouldn't the component apply a `!important` rather then saying the page styles don't apply to when the headers are in "lightdom"?
Is this a violation of those separation of concerns, as we won't be able to control other sites page styles but still want to ensure the components output the correct styling?

elements/rh-accordion/demo/color-context.html
bennypowers: That will fix it on next release

In prod, playgrounds load packages from unpkg

docs/styles/typography.css
bennypowers: Maybe if we add `:is([slot])` to the slotted selector we'll get a tiny specificity bump

zeroedin: I'm still not sure why the specificity from the page css here would matter to the concerns laid out above.

If we are trying to enforce a specific style shouldn't this be the approach: https://codepen.io/zeroedin/pen/MWddJPQ

marionnegp (changes requested):
I'm still seeing some heading size weirdness. The default demo is showing a 16px `<h2>` and the other demos have really large headings and titles. 

Once card headings are excluded, will RHDS users be able to set the heading sizes later?

elements/rh-card/rh-card.css
```diff
@@ -55,7 +55,7 @@ article {
   display: none;
 }
 
-#header ::slotted(:is(h1, h2, h3, h4, h5, h6)),
+#header ::slotted([slot]:is(h1, h2, h3, h4, h5, h6)),
```
bennypowers, line 58:
```suggestion
#header ::slotted(:is(h1, h2, h3, h4, h5, h6)),
```

elements/rh-card/rh-card.css
```diff
@@ -80,7 +80,7 @@ article {
   margin-block: auto var(--rh-space-xl, 24px);
 }
 
-#header ::slotted(:is(h1, h2, h3, h4, h5, h6)) {
+#header ::slotted([slot]:is(h1, h2, h3, h4, h5, h6)) {
```
bennypowers, line 83:
```suggestion
#header ::slotted(:is(h1, h2, h3, h4, h5, h6)) {
```

docs/styles/typography.css
bennypowers: #1701 might help here. I reverted these changes to page css, this pr will now focus on other improvements, like getting rid of the `data-demo` layout css. If individual demos want to put their elements in a padded container, that should be explicit in the demo css, rather than implicit in the dev server css

bennypowers:
*sigh*. I rebased this to staging/charmander, then github auto merged this before it was strictly ready.

I'll open a new branch off of staging/charmander

bennypowers:
https://github.com/RedHat-UX/red-hat-design-system/pull/1713

PR #1684, opened 2024-07-01
docs: hide demos tab when there are none

bennypowers:
## What I did

1. hide the demos tab when the list is empty

## Testing Instructions

1. check that demos tab exists for e.g. alert
2. ensure that it does not for e.g. popover


markcaron (approved):
Loop Gets That Manifest

Issue #1591, opened 2024-06-05
[bug] <rh-tile> using wrong font size for headings


marionnegp:
### Describe the bug

The tile demo on its docs [Overview page](https://ux.redhat.com/elements/tile/) is showing an `<h2>` with a 28px font size because the font size is coming from `typography.css`. The element should define the heading size.

### Which Component?

rh-tile

### Steps to reproduce

Visit demo on [Tile's Overview page](https://ux.redhat.com/elements/tile/)

### Expected behaviour

Heading for tile should be:
- 20px for link tile at default size
- 18px for link tile at compact size
- 18px for selectable tiles

### Operating System (OS)

macOS

markcaron:
@marionnegp I think quite a bit of the `font-sizes` are off or at least left open to inheriting the `1rem` / `16px` body text from `<rh-tile>`.

The default slot `font-size` is `16px`, which seems like it should be `18px` in the specs.

PR #1692, opened 2024-07-03
fix(tile): tile headline size

markcaron:
## What I did

1. Fixed the headline, title, body text, and footer font-sizes to match the [Tile specs](https://www.figma.com/design/8cZsqpCsq3KyiHf5VyBuRO/Element---Tile?node-id=507-2284&m=dev) for both the default and compact variants.

## Testing Instructions

1. Inspect the font-sizes of the headlines, title, body, and footer in the demos and ensure they match both variants in the spec.
2. Dev: check that my CSS looks good too, and that there should be no repercussions (e.g., with my `!important` flags).

Closes: #1591

### DIFF
--- a/.changeset/tiny-melons-help.md
+++ b/.changeset/tiny-melons-help.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+Fixed the headline, title, body text, and footer font-sizes to match the Tile specs for both the default and compact variants
--- a/elements/rh-tile/demo/color-context.html
+++ b/elements/rh-tile/demo/color-context.html
@@ -36,7 +36,7 @@
   <h2>Desaturated heading</h2>
 
   <rh-tile desaturated>
-    <img slot ="icon" src="logo-red-hat.svg" alt="Red Hat">
+    <img slot="icon" src="logo-red-hat.svg" alt="Red Hat">
     <div slot="title">Title</div>
     <h2 slot="headline"><a href="#top">Link</a></h2>
     Lorem ipsum dolor sit amet, consectetur adipiscing elit.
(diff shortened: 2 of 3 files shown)


markcaron:
CC: @OneEightyFirst — FYI. We're fixing the Tile heading sizes. LMK if you think this will mess up anything on the implementation side.

zeroedin (changes requested):
Needs a changeset

zeroedin (approved):
Laying Grout Tile Mosaics

marionnegp (changes requested):
I'm still seeing the heading being bolder than it should be:
Screenshot 2024-07-10 at 9 59 26 AM
Screenshot 2024-07-10 at 9 59 11 AM

Should be `var(--rh-font-weight-body-text-medium, 500)`

Is it Netifly/UX dot styles interfering?

markcaron:
@marionnegp it is picking up the heavier weight (700) from ux dot styles. Will fix

zeroedin:
> 
> Should be `var(--rh-font-weight-body-text-medium, 500)`
> 
> Is it Netifly/UX dot styles interfering?

Looks like we need to add an `!important` to the font-weight as well:

https://github.com/RedHat-UX/red-hat-design-system/blob/aa02acbffe418dab434280ba0281f29ebe48ad86/elements/rh-tile/rh-tile.css#L38

```
font-weight: var(--rh-font-weight-heading-medium, 500) !important;
```

marionnegp (approved):
Lowbrow Gibberish, Toddlers Mimic

bennypowers:
> Lowbrow Gibberish, Toddlers Mimic

funny *and* topical. 10 points

PR #1697, opened 2024-07-03
docs: add aria-labels to sample icons on status page

hellogreg:
## What I did

1. Added an `aria-label` value to each of the five sample icons at the top: "{status} icon"

## Testing Instructions

1. Go the [DP Design/code status page](https://deploy-preview-1697--red-hat-design-system.netlify.app/design-code-status/).
2. Inspect the sample tag icons in the Overview section's table.
3. Verify that each icon has a populated `aria-label` attribute.

There are other ways to test (e.g., via a11y plugins), so do as you see fit. But the above does the trick.

Sample tag icons screencap

### DIFF
--- a/docs/design-code-status/index.md
+++ b/docs/design-code-status/index.md
@@ -2,7 +2,7 @@
 layout: layouts/pages/basic.njk
 title: Design/code status
 hasToc: true
-importElements: 
+importElements:
   - rh-table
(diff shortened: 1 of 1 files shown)


markcaron (approved):
Laudable: Greg Tackling Micro-improvements

PR #1711, opened 2024-07-12
docs: update Typography page

marionnegp:
## What I did

Updated the Typography page based on [this Figma file](https://www.figma.com/proto/ehnLuOYqK8zZyCfVeMmdzV/Foundations?page-id=615%3A1236&node-id=615-1240&viewport=-3302%2C110%2C1&t=cWRWiAtz8iMyIp1P-1&scaling=scale-down&content-scaling=fixed).

## Testing Instructions

View the DP's [Typography page](https://deploy-preview-1711--red-hat-design-system.netlify.app/foundations/typography/).

## Notes to Reviewers

@coreyvickery, tagged you for content and design review.
- I don't think the Brand team typically links their email publicly, so I've just included a link to their site each time. 
- I linked the "Get support" page instead of our design system email address on the "Overview" subpage.
- For the blockquote and title tables, I switched the order of the rows so that they go from largest to smallest like the other tables.
- Are "Italic" and "Medium" used as style names? Questioning whether they should be capitalized. I changed it on the "Usage" page, but I noticed that they were captalized on the "Scale" page as well.

@zeroedin, tagged you to check that I'm editing the `style.css`, `rhds.cjs`, and  `basic.njk` files correctly.


coreyvickery (changes requested):
@marionnegp Really good first pass. Some quick notes:

### Overview page
- [ ] Ensure the image under the `Font family` heading is `1140px` wide
- [ ] The table under the `International` heading is not rendering for me, is it for you?

### Scale page
- [ ] Is there a way to make all of the table columns the same width? Maybe using percentages?

### Usage page
- [ ] I updated the images under `Spacing in between text styles`, look for them in Slack.

coreyvickery (approved):

docs/_includes/layouts/pages/basic.njk
```diff
@@ -5,6 +5,7 @@ importElements:
   - rh-footer
   - rh-alert
   - rh-subnav
+  - rh-table
```
zeroedin, line 8: Instead of adding `rh-table` to the page level layout, please add It to the individual pages in the front matter.

`foundations/typography/index.md` & `foundations/typography/index.md`
```
importElements:
  - rh-table
```

Otherwise, we are polluting table JS across all pages that might not use it.  It is also true that `rh-alert` should be evaluated in this very same way here (which was not added by you).  basic should be kept to the lowest common denominator used in all pages.

docs/styles/styles.css
```diff
@@ -325,4 +325,30 @@
...
+  Changes text color to red */
+  td.altered-text p:last-child {
+    color: var(--rh-color-text-brand-on-light, #ee0000);
+  }
+
+  /* Used in Foundations > Typography.
+  Changes text color to green */
+  td.not-altered-text p:last-child {
+    color: var(--rh-color-green-60, #3d7317);
+  }
```
zeroedin, line 353: Because this is specific to `foundations/typography/scale.md` I would add it as a page-specific style and possibly a special table style.

If this style becomes pervasive throughout the site then we'd move it to the patterns @layer in styles.css

Example:
```
<style>
rh-table td.altered-text p,
rh-table td.not-altered-text p {
  margin-block-start: 0;
  margin-block-end: 0;
}
```

zeroedin (approved):
Letters Gaining Typographic Meaning

PR #1720, opened 2024-07-17
feat(tile): add border-block-start custom props

markcaron:
## What I did

1. Added `--rh-tile-border-block-start-width` and `--rh-tile-border-block-start-color` CSS custom props
2. Updated the Custom Props demo to illustrate the use

## Testing Instructions

1. Check out [Custom Props demo locally](http://localhost:8000/elements/tile/demo/custom-props/) to verify the new custom props work
2. Check my CSS skills
3. Proofread the [Custom Props info in the docs](https://deploy-preview-1720--red-hat-design-system.netlify.app/elements/tile/code/#rh-tile)

PR #1721, opened 2024-07-18
feat(tile): accented tile CSS pattern demo

markcaron:
## What I did

1. Added an `Accented tiles` CSS pattern demo based on [Benny's idea](https://github.com/RedHat-UX/red-hat-design-system/pull/1720#issuecomment-2235349886)

## Testing Instructions

1. Checkout the [Accented tiles demo](https://deploy-preview-1721--red-hat-design-system.netlify.app/elements/tile/demos/#demo-accented-tiles)


elements/rh-tile/demo/accented-tiles.html
```diff
@@ -0,0 +1,66 @@
...
+        display: block;
+        inset-block-start: 0;
+        inset-inline: 0;
+        border-block-start-color: var(--rh-color-brand-red-on-light, #ee0000);
+        border-block-start-width: 3px;
+        border-block-start-style: solid;
+        pointer-events: none;
+      }
+    }
+  }
```
zeroedin, line 65: IIRC we are still on the stance of not using CSS nesting in production code.  I suggest rewriting this in plain CSS format for now. The holdout was ESR Firefox being 115 which didn't support CSS Nesting.  Given the recent change of Firefox ESR becoming v128, we should be able to do all CSS in nested format soon.

elements/rh-tile/demo/custom-props.html
```diff
@@ -39,13 +39,15 @@ <h2 slot="headline"><a href="#top">Link 2</a></h2>
     column-gap: var(--rh-space-2xl, 32px);
     row-gap: var(--rh-space-2xl, 32px);
     margin-inline-end: var(--rh-space-md, 8px);
+    & > * {
+      width: 320px;
+    }
```
zeroedin, line 44: nesting here as well.

elements/rh-tile/demo/accented-tiles.html
markcaron: Good point. I just took Benny's code mostly verbatim.

Looking at this again, I think what I've got in mind is probably a better approach or at least one more consistent to how implementors might apply this anyway.

zeroedin (commented):
Thinking this change should be a pattern instead of a demo of the component.  What we are demonstrating is a pattern implementation, so we should move this under Patterns.

marionnegp (approved):

zeroedin (approved):
Laying Geometric Tiles Meticulously

docs/patterns/tile/index.md
```diff
@@ -0,0 +1,89 @@
...
+  inset-inline: 0;
+  border-block-start-color: var(--rh-color-brand-red-on-light, #ee0000);
+  border-block-start-width: var(--rh-border-width-lg, 3px);
+  border-block-start-style: solid;
+  pointer-events: none;
+}</script>
+    </rh-code-block>
+    <rh-cta><a href="/elements/tile/demos/#demo-accented-tiles">View accented tile demo</a></rh-cta>
+  </div>
+</div>
```
marionnegp, line 89: Is there a way to get the bottom of the code block (or the CTA) and image container to line up when they're next to each other?

zeroedin: Maybe consider moving the CTA out of the grid

image

bennypowers:
In that case, https://github.com/RedHat-UX/red-hat-design-system/pull/1188 needs to be resolved first

docs/patterns/tile/index.md
markcaron: @zeroedin I was thinking about this exact thing last night. Will do.

bennypowers: Also, prefer the `<rh-cta href=`

markcaron:
@bennypowers ok. I'll wait to merge until I can pull in #1188 and use a similar `patterns` directory structure.

hellogreg (approved):
Literary genius. Top marks!

PR #1729, opened 2024-07-30
feat(tile): include grid layout in `rh-tile-group` element

### LINKED ISSUE #1303: [bug] `<rh-tile-group>`: built in style guidelines
### Description

The style guidelines give some terse advice about tile group grid widths. 

Tile and Tile Group should implement those styles, not page devs.

### Suggested solution

If you check the tile demos, you'll see that the demo files include global css in order to layout the tile groups. These styles should be built-in to the tile group element, such that we'd be able to delete all the global css from the tile demo files, and they would still 
a) look like they do now
b) work in multiple ds grid layout circumstances

### Example API

_No response_

### Additional context

Good excuse to work on grid tokens, play with subgrid as well.

adamjohnson:
## What I did

1. Removed CSS grid / flexbox layout from any demos that had an `<rh-tile-group>` element.
1. Added CSS grid to apply to the `:host` of `<rh-tile-group>`.
1. Improved the layout of the Color Context demo.
1. Removed the `tile-group-flex.html` demo.
1. Removed the `tile-layout.html` demo.
1. Set `<main>` to `display: block;` to improve layout in some `rh-tile` demos.
1. Removed `<rh-tile-group>` from the custom props demo.

## Testing Instructions

Check the DP of the following demos:

  * [Tile Group](https://deploy-preview-1729--red-hat-design-system.netlify.app/elements/tile/demo/tile-group/)
  * [Tile Group Radio](https://deploy-preview-1729--red-hat-design-system.netlify.app/elements/tile/demo/tile-group-radio/)
  * [Tile Group Radio Disabled](https://deploy-preview-1729--red-hat-design-system.netlify.app/elements/tile/demo/tile-group-radio-disabled/)
  * [Tile Group Disabled](https://deploy-preview-1729--red-hat-design-system.netlify.app/elements/tile/demo/tile-group-disabled/)
  * [Color Context](https://deploy-preview-1729--red-hat-design-system.netlify.app/elements/tile/demo/color-context/)
  * [Tile Group Custom Props](https://deploy-preview-1729--red-hat-design-system.netlify.app/elements/tile/demo/custom-props/)

## Notes to Reviewers

This PR does not add any grid/flexbox styles to the demos that don't have a `<rh-tile-group>` element in them. Currently, `<rh-tile>` styles are untouched. If `rh-tile` styles should implement a grid layout or max-width, let me know and I can make that change. 

Closes #1303.


zeroedin (commented):
Should the tile layout demo also be updated to use `<rh-tile-group>`. I am confused what the difference is.  

Maybe tile layout demo really is just a pattern and should be moved to the patterns section of the site. I feel like it causes confusion here in the demos.  Thoughts?

.changeset/eleven-ligers-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-tile>`: `<rh-tile-group>` will now apply a CSS grid to child `<rh-tile>` elements.
```
zeroedin, line 5:
```Suggestion
`<rh-tile-group>`:  corrected application of grid layout of slotted elements.
```

Any slotted element in this case will be applied to the grid.  Feel like we should be more general here given that the grid won't just apply to `rh-tile` children, but any element that is slotted.

adamjohnson:
> Should the tile layout demo also be updated to use `<rh-tile-group>`. I am confused what the difference is.

Wrapping a few `<rh-tile>` elements in a `<rh-tile-group>` automatically changes the `<rh-tile>` elements to radios/checkboxes. So, in this case, `<rh-tile-group>` should not be used here.

> Maybe tile layout demo really is just a pattern and should be moved to the patterns section of the site. I feel like it causes confusion here in the demos. Thoughts?

My vote would be to delete this demo. It doesn't do anything except put them in a 3-column grid—which just about anyone will know how to do anyway. WDYT @marionnegp?

.changeset/eleven-ligers-share.md
adamjohnson: Updated in [0a18877](https://github.com/RedHat-UX/red-hat-design-system/pull/1729/commits/0a188776dfed8626abf08f74a6fa17b86494d441).

marionnegp (commented):
I'm good with removing the Tile Layout demo. 

The tiles in most of the demos stretch to fill the height of the window. Is that something that should be fixed? I tried putting the `<rh-tile-group>` in a `<div>`. That fixed it, but I don't know if that's something we'd actually want to include.
Screenshot 2024-08-02 at 2 43 41 PM

marionnegp (approved):

zeroedin (approved):
Like Grided Tiles Man...  🏄

PR #1731, opened 2024-07-30
fix(button): center play icon in circle

adamjohnson:
## What I did

1. Centered the icon in the "play" variant of `<rh-button>`.

## Testing Instructions

1. First, view the "play" variant of `<rh-button>` in production [on uxdot](https://ux.redhat.com/elements/button/demo/variants/).
    * Note how it is not centered in the black circle
2. Now, go to the [variants demo](https://deploy-preview-1731--red-hat-design-system.netlify.app/elements/button/demo/variants/) in this DP.
3. Notice how the play icon is now centered in the circle.

## Notes to Reviewers (rounded corners or not?)

If you look at the `<rh-button>` [Figma](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/Red-Hat-Design-System-library?m=auto&node-id=451-3138), you'll notice the play button does _not_ have rounded corners. [The docs](https://ux.redhat.com/elements/button/) for `<rh-button>` also reflect this.

If you look at the [current implementation](https://ux.redhat.com/elements/button/demo/variants/) of `<rh-button variant="play">` and the [Red Hat Icons Play SVGs](https://red-hat-icons.netlify.app/#ui), they _do_ have rounded corners. 

Should the corners be rounded?

## To Do's

- [x] Decide whether we should proceed with rounded corners or not for the play icon (Update 8/5/24: rounded!)
- [ ] Update docs accordingly (to be handled in a separate issue/PR)

### DIFF
--- a/.changeset/loud-spies-work.md
+++ b/.changeset/loud-spies-work.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-button variant="play">`: Play icon is now centered in the circle of the button.
--- a/elements/rh-button/rh-button.css
+++ b/elements/rh-button/rh-button.css
@@ -470,7 +470,7 @@ button:hover {
   --_hover-color: var(--_play-color);
   --_hover-background-color: var(--_play-hover-background-color);
   --_hover-background-opacity: var(--_play-hover-background-opacity);
-  --_icon-size: var(--rh-size-icon-03, 32px);
+  --_icon-size: var(--rh-size-icon-02, 24px);
 }
 
 :host(:is([variant="play"])) [part="icon"] {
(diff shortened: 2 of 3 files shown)


marionnegp (changes requested):
Let's remove the rounded corners, so we don't have to change more stuff. The play icon from Brand doesn't round them either.

Can the icon size be `var(--rh-size-icon-02, 24px)`?

marionnegp:
> The play icon from Brand's updated set doesn't round them either.

Wait, just kidding. I had to zoom in a bunch to see the rounded corners for the new play button icon. Let me know if you need an SVG for the triangle icon.

marionnegp (approved):

zeroedin (approved):
Look! Guided to middle!

Issue #1733, opened 2024-08-01
[docs] Put an icon next to elements or pattern that don't have a corresponding web component yet


zhawkins:
### Description

Recently we've run into designs coming through that use RHDS components that aren't code ready yet.

This would be a nice way to designate something isn't quite ready to be used yet.

Relevant slack conversation: https://redhat-internal.slack.com/archives/C0686DG8YSJ/p1722539567096989

### Links

- ux.redhat.com

### Screenshots

<img width="1970" alt="Screenshot 2024-08-01 at 2 23 43 PM" src="https://github.com/user-attachments/assets/8c2cc48b-99f2-462b-8c7c-1f6f40973935">

PR #1734, opened 2024-08-01
docs: quick and dirty icon update for in progress elements

zhawkins:
## What I did

Added a quick and dirty icon fix so that in-progress elements are more easily distinguished. 

## Testing Instructions

1. View the docs site, check to see if there is a caution icon next to the element navigation item AND next to the title within the element main page.

Screenshot 2024-08-01 at 2 23 43 PM

## Notes to Reviewers

Full disclaimer, I realize that there's likely a much much much better way to do this 😅. However, seeing as it's a pretty sore pain point, this would band-aid the issue until a proper fix is in place.

Thanks for considering it!

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/1733

### DIFF
--- a/CONTRIBUTING_DESIGN.md
+++ b/CONTRIBUTING_DESIGN.md
@@ -10,7 +10,7 @@ contributing guide](./CONTRIBUTING_DEV.md#prerequisites)
 
 Once set up, run the 11ty development server with the following command:
 ```bash
-npm run watch:docs
+npm run serve
 ```
 
 This will start eleventy and a watch process that will reload the site if `*.njk`, `*.md`, `*.scss`,
--- a/docs/_data/repoStatus.yaml
+++ b/docs/_data/repoStatus.yaml
@@ -167,8 +167,7 @@
       status: N/A
     - name: Documentation
       status: Ready
-- tagName: rh-navigation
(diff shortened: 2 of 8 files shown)


zeroedin:
Heads up this will have to be reworked with `staging/charmander` as we use a different data structure basing the repoStatus data off `tagName` ie: "rh-accordion" rather than the human friend name "Accordion".  Same concept though.

@coreyvickery I added Progress steps & Navigation primary to the repoStatus data, can you let me know what the actual status of those are an ill update it here.

zhawkins:
@zeroedin I like it! I worry the grey is too easy to overlook. 🤔 How about "purple" for "planned"?

image

markcaron:
@zhawkins that would be a design change and would have to be approved by @coreyvickery, since it affects the [whole status table](https://deploy-preview-1734--red-hat-design-system.netlify.app/design-code-status/). 

At the moment we're trying to get Charmander out the door, so we can address the color later.

zhawkins:
@markcaron ah gotcha! In that case, looks good to me! Ship it!

markcaron (approved):
Look, Gotta Tag Mapped (Elements)

Issue #1748, opened 2024-08-13
`<rh-accordion>`: remove  deprecated  icon property


zeroedin:
Remove deprecated icon property 

https://github.com/RedHat-UX/red-hat-design-system/blob/bdef81f6a269f51bf0356e8c0bc3757299c61077/elements/rh-accordion/rh-accordion-header.ts#L70-L71

Deprecated 4months ago

PR #1760, opened 2024-08-14
fix(accordion): remove deprecated icon property from header

zeroedin:
## What I did

1. Removed deprecated icon property from header

Closees #1748 

## Testing Instructions

1.

## Notes to Reviewers

Unless I'm missing it, the icon property didn't appear to be used or its functionality had already been removed.  Please check me here.

### DIFF
--- a/.changeset/silly-turtles-hang.md
+++ b/.changeset/silly-turtles-hang.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-accordion-header>`: removed deprecated icon property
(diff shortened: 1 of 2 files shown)


bennypowers (commented):
did this just not do anything? is there an alternative people should use?

adamjohnson (approved):
I concur with Steve. It doesn't appear to get used in the [original PR](https://github.com/RedHat-UX/red-hat-design-system/pull/594/files#diff-7b327c8f9fb343b5cf193cef0b288dbc3be2a4cb3323cbadb0d9dfd9b886ef62R49-R58). There's no mention of this property in the demos.

Maybe it was added somewhere in the middle? That said, it appears unused at the moment.

PR #1763, opened 2024-08-15
chore: integrate pfe 4

bennypowers:
## What I did

1. adjust apis and such

## Testing Instructions

1. for each changed element, check tab behaviour
2. check docs (manifest) in 11ty and dev server

## Notes to reviewers
In order for this PR to make sense, you need to use the PFE 4 packages

1. Build PFE 4 from branch
2. `npm -ws pack` in pfe root dir
3. delete the contents of `@patternfly/elements`, `@patternfly/pfe-core`, `@patternfly/pfe-tools` from node_modules
4. copy in the appropriate files from the tarballs made in step 2


zeroedin:
@bennypowers I believe the test failing for dialog is due to the switch to `@observes`.  Maybe a timing issue?  Seem's the element itself never `opens` before it is "cancelled" and closed, therefore the property never actually changes and the event never triggers.  Seem plausible?  I'll push and update if I can solve it by EOD.

bennypowers:
in fact there was a bug in dialog - it determined the type of event to fire only *after* completing the update lifecycle. the fix was the decide on which event to fire before waiting on the update lifecycle.

we may want to change this further, to make the events cancellable and only perform the update if the events are not cancelled.

PR #1790, opened 2024-08-28
docs: remove sidenav links underlines

markcaron:
## What I did

1. Removed the inherited underline from links in the sidenav

## Testing Instructions

1. Check it looks right on the deploy preview
2. Check the code in the lightdom CSS is the right way to go


adamjohnson (approved):
Links Gain Trendy Modifications

PR #1792, opened 2024-08-28
feat: theme properties

### LINKED ISSUE #1779: [docs] patterns html files should be self-sufficient
### Description

the files in `docs/patterns/*/patterns/*.html` contain `<uxdot-pattern>` and `<rh-code-block>` elements, 
in order to improve readability, we should move those out into the docs pages themselves, 
such that the html files contain *only* the code necessary to implement the pattern, 
and no special docs helpers
i think it's ok for those files to link to pattern-definition css files like `logo-wall.css` for example, but not for them to contain uxdot helpers

That way, if a user would copy-and-paste the contents of a pattern html file into a blank html document (with an appropriate import map already available), they would receive the desired outcome.

Certainly, there must not be any `<rh-code-block>` elements in the pattern files, (except of course, the files related to code block patterns)

The uxdot-pattern element, and the docs-browsing experience it provides is excellent and we should keep it, just that we should wrap the contents of those files in that element, and not the other way around

### Screenshots

_No response_

### Steps to reproduce

see https://github.com/RedHat-UX/red-hat-design-system/blob/98e54690d82c35039f7f2743444b7177b46bd748/docs/patterns/card/patterns/asset.html#L51 

### Expected behaviour

there shouldn't be uxdot helper elements in those files

### Operating System (OS)

iOS, macOS, Linux, Windows

### Browser

_No response_

_No response_

### LINKED ISSUE #1793: [docs] page styles affecting demos
### Description

Context demos are getting page styles, which can have unexpected effects

this patch demonstrates how to fix the problem in dev tools

```css

/* Element | https://deploy-preview-1792--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/ */

#provider {
  height: 100%;
  height: auto;
}

/* styles.css | https://deploy-preview-1792--red-hat-design-system.netlify.app/styles/dev-server/styles.css */

@layer layout {
  rh-context-demo {
    /* height: auto; */
    height: 100%;
  }
}

/* Element | https://deploy-preview-1792--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/ */

body > main:nth-child(1) > rh-context-demo:nth-child(1) {
  height: 100%;
}

/* Element | https://deploy-preview-1792--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/ */

body > main:nth-child(1) {
  height: 100%;
}

/* Element | https://deploy-preview-1792--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/ */

body {
  height: 100%;
}

/* Element | https://deploy-preview-1792--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/ */

html {
  height: 100%;
}
```

### Screenshots

image

### Steps to reproduce

See https://ux.redhat.com/elements/accordion/demo/color-context/

### Expected behaviour

Demos should have the bare minimum of styles applied - basically just `global.css` and maybe some very basic page typography.

### Operating System (OS)

iOS, macOS, Linux, Windows

### Browser

_No response_

### LINKED ISSUE #1848: [docs][bug] patterns/card page has several display errors
### Description

On ux.redhat.com/patterns/card, there are several display errors. 

1. The `<uxdot-pattern>` element is displaying inline and this is overlapping the content that follows, including the footer
2. The `<p>` following `#explicit-card-theming` overlaps a code container

### Screenshots

image

image

### Steps to reproduce

1. go to the page ux.redhat.com/patterns/card
2. Scroll down the page to "explicit card themeing"

### Expected behaviour

Hard to say what I'd expect to see, but I'm guessing `<uxdot-pattern>` would be `display: block` and not overlap other content. It would be more clear what code example went with which card. 

### Operating System (OS)

macOS

### Browser

Chrome, Safari, Firefox

_No response_

bennypowers:
Closes #1779 
Closes #1793 
Closes #1848

# IMPORTANT!

- [x] :point_right: **PLEASE SEE** https://github.com/RedHat-UX/red-hat-design-tokens/pull/138 :point_left: 
- [x] https://github.com/patternfly/patternfly-elements/pull/2834

## What I did

1. added theme properties (see notes below)
2. refactored color context decorators
3. updated card pattern docs, removing as many `[color-palette]` as i could and replacing them with theme properties

### TODO
- [x] update the changesets with latest version of the pr
- [x] add more theming tokens
- [x] add those tokens to the tokens repo
- [ ] lots of tests
- [x] we can probably remove almost all `-on-light` and `-on-dark` in component definition css

## Testing Instructions

1. check out the card patterns pages
2. for all elements updated in "Files Changed", please pay close attention to DP demos, particularly color-context demos.

## Notes to Reviewers

:point_right: **Reminder** that the playground demos use the _published_ versions of packages, but the "View in own tab" demos (e.g. `/elements/jazz-hands/demo/ragtime.html`) use the on-branch versions, so test those ones. :point_left: 

Hopefully this PR will help to demonstrate the concepts behind the theming system. To that end, we introduce new themable tokens. Here is the initial draft of the new color-context consumer css file, shared with all context consumer elements

```css
.on.dark {
  --rh-color-text-primary: var(--rh-color-text-primary-on-dark, #ffffff);
  --rh-color-border-subtle: var(--rh-color-border-subtle-on-dark, #707070);
  --rh-color-border-strong: var(--rh-color-border-strong-on-dark, #ffffff);
  --rh-color-border-interactive: var(--rh-color-border-interactive-on-dark, #92c5f9);
}

.on.light {
  --rh-color-text-primary: var(--rh-color-text-primary-on-light, #151515);
  --rh-color-border-subtle: var(--rh-color-border-subtle-on-light, #c7c7c7);
  --rh-color-border-strong: var(--rh-color-border-strong-on-light, #151515);
  --rh-color-border-interactive: var(--rh-color-border-interactive-on-light, #0066cc);
}
```

These new tokens are practically-speaking "read-only", meaning, they respond to the current `light|dark` context and provide their children with the correct values for a given token


marionnegp:
@bennypowers, this sounds similar to what PF is doing with their tokens because they have an intermediary layer between the crayon tokens and the usable semantic tokens.

Does mean that we'd deprecate cards that use `--rh-color-surface-lighter` and other elements that change based on `color-palette`?

bennypowers:
I think the surface tokens are still useful because they represent the color-palette values

Whereas the theming tokens represent the colors for elements `on` those surfaces

marionnegp (changes requested):
Feedback for the "Global color tokens" page:
- [ ] Any tokens in a "Theme tokens" container overlap with each other.
Screenshot 2024-09-06 at 9 32 50 AM

- [ ] I'm also not seeing a change when I toggle between themes for brand, interactive, status, or surface tokens.

- [ ] Anchor link icons aren't horizontally aligned with `<h5>`s properly.
Screenshot 2024-09-06 at 9 35 06 AM

bennypowers:
> Feedback for the "Global color tokens" page:
> 
>     * [ ]  Any tokens in a "Theme tokens" container overlap with each other.
> 
> 
> Screenshot 2024-09-06 at 9 32 50 AM
> 
> -[ ] I'm also not seeing a change when I toggle between themes for brand, interactive, status, or surface tokens.
> 
>     * [ ]  Anchor link icons aren't horizontally aligned with `<h5>`s properly.
> 
> 
> Screenshot 2024-09-06 at 9 35 06 AM

the styling works on ff but not chrome. samp.css will need some attention

the context switcher doesn't work on either, something must have changed in a recent commit

marionnegp:
@bennypowers, the forced palette accordion in the `<rh-accordion>` demo allows you to see only the lightest and darkest surface colors. I'm seeing the same issue in `<rh-card>`. Is this expected behavior based on the changes you're proposing or an issue with the context switcher?

Screenshot 2024-09-06 at 10 45 24 AM
Screenshot 2024-09-06 at 10 47 04 AM

zeroedin:
> @bennypowers, the forced palette accordion in the `<rh-accordion>` demo allows you to see only the lightest and darkest surface colors. I'm seeing the same issue in `<rh-card>`. Is this expected behavior based on the changes you're proposing or an issue with the context switcher?


This has come up a couple times before in variant discussions:

For card:

```
   */
  @colorContextProvider()
  @property(
    {
      reflect: true,
      attribute: 'color-palette',
    }
  ) colorPalette?: 'darkest' | 'lightest' | 'lighter';
```
 The code restricts to `darkest`, `lightest`, and `lighter`
 
 For Accordion the CSS was written in a way where it only responds to the context so only `light` or `dark`.
 
 The question I always wondered was this a lost in translation between the Figma specifications and the components or are they meant to be restrictive?

bennypowers:
@marionnegp @zeroedin us I think that's correct. These elements were crafted to only allow certain color palettes.

If we want, we can add support for more palettes to Those elements. I think it would be worthwhile adding to this PR a paragraph or two of prose, or astructured  section (of rh tags?) to each element's style or code docs section on which colour palettes it supports.

But I think adding more colour palette support to card or accordion elements is out of scope of this PR.

paceaux:
I have two thoughts:

### One
Could you sub-title the "[color palettes"](https://deploy-preview-1792--red-hat-design-system.netlify.app/patterns/card/#color-palettes) section "implicit theming", so that there's parity with the "explicit card theming that comes later" ?

### Two

In the "Custom Theming" section, there's a card that displays with a redish background in the lighter color palettes, but it goes to purple in the darker ones (via `--rh-color-border-subtle-on-dark`) 

It's not immediately obvious to me _why_ that's the color that's used on that dark background. 

The reason this stands out to me is because any card with a red border/background is going to stand out to me as signaling "error/warning/danger", and when the background-color changes, this feels like a loss of information. Purple just doesn't seem like a "dark-themed equivalent to red".

If you could either explain the red-purple equivalence, or chose another color, that might help. I get that this is demoing custom themes .... so maybe also I'm just having a weird opinion about nothing meaningful.

bennypowers:
> I have two thoughts:
> ### One
> 
> Could you sub-title the "[color palettes"](https://deploy-preview-1792--red-hat-design-system.netlify.app/patterns/card/#color-palettes) section "implicit theming"

Goal number one is to clearly define brand-new terms invented for RHDS called "color-palette", "on", and "theme" - we're still workshopping "on" (maybe "background" is better?) and "theme" but, "color-palette" is here to stay, since that's the name of the attribute on all the elements
 
> ### Two
> 
> In the "Custom Theming" section

My colour-sense notwithstanding, if you can do a better wine-themed color scheme I'm all for it. there's no meaning to these particular colours outside of demonstrating that it can be done. It's certainly been a highly-requested feature and it's in use all over the place.

marionnegp (commented):
If you think any of this feedback should be tackled in separate PRs, I'm good with merging #1792 into the staging branch.

**accordion:**
- Background color of closed accordion header should be transparent. When open, its bg color should match the panel’s bg color.

**alert:**
- Alert on dark is using white text. Maybe we have to add a token for situations like this, or should alert be a color context provider?

**audio player:**
- Needs a `--rh-color-border-subtle-on-light` or `--rh-color-border-subtle-on-dark`
- Needs a dark theme demo

**badge:**
- The critical badge should be using a darker red orange color, and the important should also be using a red orange shade. I think this could be updated by adding status tokens to the other repo.

**blockquote:**
- `border-inline-start` should be using `--rh-color-accent-brand-on-light` or `--rh-color-accent-brand-on-dark`

**button:**
- Disabled button on light should be using `--rh-color-gray-50`, but there’s no semantic token for this. Should probably add disabled color tokens.
- Disabled button on dark is using same colors as the on light version, but it should be different based on [Figma file](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/Red-Hat-Design-System-library?node-id=9-521&m=dev).

**card:**
- When I change the card’s color palette, it also changes the color palette of the rest of the demo. Is this a bug in the demo or expected behavior based on new theming rules?

**health index:**
- Border colors for light and dark themes don’t look quite the same as [what we have in Figma](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/Red-Hat-Design-System-library?node-id=2709-157&m=dev)

**stat:**
- On dark, statistic and title slots should use `--rh-color-text-brand-on-dark`

**table:**
- `--rh-color-interactive-primary-default-on-dark` doesn’t have enough contrast with the bg color of the hover state.

**tabs:**
- Links in tab panels aren’t showing correct colors in dark or light theme. I think this might be due to site styles messing with the demo.
- Hover and focus states aren’t showing for any of the color context demo options. (I can see the focus ring though.)

**tile:**
- Footer slot and arrow are misaligned. (Not due to color context, but wanted to note it in case it's a demo-only bug and not an issue in the element.)

PR #1796, opened 2024-08-28
docs: fixing repoStatus tag icons

markcaron:
## What I did

1. Fixed the `rh-tag` icon situation in the `repoStatus` tables since they were either not showing or the positioning and spacing was off.

For example:
image

image

## Testing Instructions

1. Double check the [Design/code status page](https://deploy-preview-1796--red-hat-design-system.netlify.app/design-code-status/) and at least one element's Overview page to ensure the status tables look ok.
2. Double check code is quality

Should look something like this:
image

### DIFF
--- a/docs/_plugins/shortcodes/repoStatus.cjs
+++ b/docs/_plugins/shortcodes/repoStatus.cjs
@@ -7,45 +7,53 @@ const STATUS_LEGEND = {
     color: 'gray',
     variant: 'filled',
     icon: html`<rh-icon set="ui" icon="notification-fill">Planned</rh-icon>`,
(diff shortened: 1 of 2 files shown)


adamjohnson (approved):
I changed the variable from `iconname` to `iconName` and referenced it with the bracket syntax. If you think this works, merge it. If not, just revert my last commit and merge. Nice work 👏

markcaron:
@adamjohnson I prefer `iconName`. Thanks!

diwanshi:
@markcaron hey do we need icon variable if we're not using it?

PR #1802, opened 2024-08-29
release: clefairy

### LINKED ISSUE #1849: [docs][bug] multiple elements with `id="summary"` on subpages
Across ux dot subpages---230+ times, according to our Lumar crawler--the "On this page" text is within two elements per page: one for smaller screens (within a `<summary>`) and one for larger screens (within a `<div>`).

Both elements have the same `id` value: `"summary"`.

Though one of these elements should always have a `display: none` applied to it (assuming CSS is enabled), this still can cause validation and potential assistive tech issues--especially as there is an `aria-describedby` pointing to this `id`. 

I'd recommend giving these elements different `id` values (and then reworking site JS and CSS to account for this).

Screencap of a page inspector showing the duplicate id values

adamjohnson:
Closes #1849


adamjohnson (commented):
Adding this comment here to prevent accidental merging. When we're ready to release, resolve this comment.

Update: looks like I have to add myself as a reviewer. Hiding this comment for now.

bennypowers (changes requested):
This comment blocks merge

bennypowers (approved):

PR #1804, opened 2024-08-29
feat(table): adding automatic data-labels on col-scoped tables

### LINKED ISSUE #1390: [feat] `rh-table` no data-label
`<rh-table>` when no `[data-label]`'s are present on headers and columns, the table should still present itself in a manner that is still visually logical/readable.

Also to note after a discussion with @LyndseyR `data-label` will not get translated via the XML for RHDC.  Because we use these labels as visual output to the user, on translations they would show as english counterparts which also isn't ideal.  A special translation case maybe able to be made here, but this also might be a good reason to look at alternatives.

markcaron:
## What I did

1. Added auto-generated `data-label` attributes on cells for `col`-scoped tables

## Testing Instructions

1. Check Auto Data Labels demo
2. `data-label` attributes should be automatically generated on every cell within `tbody` based on their corresponding `thead` column.
3. It should also respect preexisting `data-label` attributes, like the last row of the table with `Custom #` labels.

image

Closes #1390


elements/rh-table/rh-table.ts
```diff
@@ -138,6 +143,19 @@ export class RhTable extends LitElement {
...
+    if (this.#table?.tHead && this.#colHeaders) {
+      this.#rows.forEach((row, index) => {
+        const cells = row.querySelectorAll(':is(td,th)');
+        cells.forEach((cell, index) => {
+          if (!cell.hasAttribute('data-label')) {
+            const colHeader = this.#colHeaders[index].innerText;
+            cell.setAttribute('data-label', colHeader);
+          }
+        });
+      });
```
adamjohnson, line 157:
```suggestion
    if (this.#table?.tHead && this.#colHeaders && this.#rows) {
      for (const row of this.#rows) {
        if (!row) {
          continue;
        }

        const cells = row.querySelectorAll(':is(td, th)');
        let headerIndex = 0;

        for (const cell of cells) {
          if (!cell.hasAttribute('data-label')) {
            const colHeader = this.#colHeaders[headerIndex]?.innerText;
            cell.setAttribute('data-label', colHeader);
          }

          headerIndex++;
        }
      }
    }
```

This cleans up these typescript warnings but there are still questions here.

1. This change alters the users HTML / lightdom. IIRC this may have problems associated with it (perhaps other folks can elaborate on that more).
2. If a user has a table with a `rowspan` or `colspan`, this returns `undefined`.
3. This focuses on table headers that are on the top row. As you mentioned, what about table headers in the left column?

Overall, this is probably more complicated than we may be expecting. Perhaps it could be done in specific cases (equal number of rows and columns, etc)? Probably some discussion to be had here.

markcaron:
Per @adamjohnson (from resolved convo above):

> ...but there are still questions here.
> 1. This change alters the users HTML / lightdom. IIRC this may have problems associated with it (perhaps other folks can elaborate on that more).
> 2. If a user has a table with a rowspan or colspan, this returns undefined.
> 3. This focuses on table headers that are on the top row. As you mentioned, what about table headers in the left column?
> Overall, this is probably more complicated than we may be expecting. Perhaps it could be done in specific cases (equal number of rows and columns, etc)? Probably some discussion to be had here.

Re: 1): I'm not sure of another way to add these "tablesaw-esque" labels without altering the light DOM HTML. We either utilize `data-label` attributes and CSS pseudo elements _-OR-_ we append the labels inside the cells in a `<span>` or something. Would need @bennypowers or @zeroedin 's advice if this is problematic.

Re: 2): Agreed that this might be an issue. One I'm hoping to tackle, hence the "draft PR" for now. I'm only accounting for column headers _so far_.

Re: 2 + 3): Introducing `rowspan` and `colspan` would require even more auto-generated support (if it wasn't marked-up by the developer, like `ids` on the `<th>` elements (in the `<thead>` and `<tbody>` and `headers=""` attributes across their respective cells.

If it's too complex, then maybe this is "opt-in" through a `responsive` boolean on `<rh-table>` with documented use on "simple tables"?

Definitely worth a deeper discussion.

elements/rh-table/rh-table.ts
```diff
@@ -60,6 +60,11 @@ export class RhTable extends LitElement {
     return this.querySelectorAll('tbody > tr') as NodeListOf<HTMLTableRowElement> | undefined;
   }
 
+  get #colHeaders(): NodeListOf<HTMLTableCellElement> | undefined {
+    return this.querySelectorAll(
+      'thead > tr > th') as NodeListOf<HTMLTableCellElement> | undefined;
```
bennypowers, line 65:
```suggestion
    return this.querySelectorAll<HTMLTableCellElement>('thead > tr > th');
```
you can assert in the call, and this function will never return `undefined`, so change the signature as well

elements/rh-table/rh-table.ts
```diff
@@ -138,6 +143,27 @@ export class RhTable extends LitElement {
     if (this.#table && this.#summary) {
       this.#table.setAttribute('aria-describedby', 'summary');
     }
+
+    /* Auto-assign `data-label` attributes based on column headers */
+    if (this.#table?.tHead && this.#colHeaders && this.#rows) {
```
bennypowers, line 148:
```suggestion
    if (this.#table?.tHead && this.#colHeaders.length && this.#rows) {
```

elements/rh-table/rh-table.ts
```diff
@@ -138,6 +143,27 @@ export class RhTable extends LitElement {
...
+      for (const row of this.#rows) {
+        if (!row) {
+          continue;
+        }
+
+        const cells = row.querySelectorAll(':is(td, th)');
+        let headerIndex = 0;
```
bennypowers, line 155: instead of tracking and incrementing this, I suggest using the `forEach` method on the NodeList, and including the second `index` parameter in the callback

bennypowers (changes requested):

elements/rh-table/rh-table.ts
markcaron: Ah ha! I had it this way originally 😉

I will see about looping w/ `forEach`

elements/rh-table/rh-table.ts
markcaron: @bennypowers `this.#colHeaders.length` I'm getting a TS error `Object is possibly 'undefined'` on this with the added `length`. Removing `length` for now, since it seems to work without it.

bennypowers: You could use optional chaining and check with `this.#colHeaders?.length`

That would be the safest solution

elements/rh-table/rh-table.ts
markcaron: @bennypowers can you give this a re-review?

bennypowers: Pushed a few commits dealing with this function

elements/rh-table/rh-table.ts
```diff
@@ -23,6 +24,11 @@ import styles from './rh-table.css';
 export class RhTable extends LitElement {
   static readonly styles = [styles];
 
+  /**
+   * Auto-generates header labels for responsive layout
+   */
+  @property({ reflect: true }) responsive?: 'stack';
```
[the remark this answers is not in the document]
markcaron: No longer needed. Going responsive by default.

markcaron:
Related feature: https://github.com/RedHat-UX/red-hat-design-system/pull/1881

.changeset/khaki-rings-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-table>`: added auto-generating table cell headings for responsive layout on small screens.
```
bennypowers, line 5:
```suggestion
`<rh-table>`: added auto-generated table cell headings for responsive layout on small screens.
```

bennypowers (approved):

PR #1816, opened 2024-09-03
fix(back-to-top): position

bennypowers:
## What I did

1. apply the [guidelines](https://ux.redhat.com/elements/back-to-top/guidelines/) to back-to-top
2. remove the (undocumented, and bad for l10n) position css props


zeroedin (approved):
Links Goes To Main

PR #1820, opened 2024-09-03
docs: refactor styles and accessibility fixes

### LINKED ISSUE #1819: docs: toc padding issue
image

Table of contents having padding-inline causes the above issue on the expanded state of the details summary on mobile sized viewports.

zeroedin:
## What I did

1. Fixes padding issue 
2. Corrects labelling of `<nav>` element in shadowroot of `uxdot-toc`

Closes #1819 

## Testing Instructions

1.

### DIFF
--- a/docs/_includes/layouts/pages/basic.njk
+++ b/docs/_includes/layouts/pages/basic.njk
@@ -22,9 +22,11 @@ importElements:
 
   {% include 'partials/component/header.njk' %}
   {% if hasToc and (content | toc).length > 0 %}
-  <uxdot-toc summary="On this page">
-    {{ content | toc | safe }}
-  </uxdot-toc>
+  <aside>
+    <uxdot-toc summary="On this page">
+      {{ content | toc | safe }}
+    </uxdot-toc>
+  </aside>
   {% endif %}
 
   <div class="container">
--- a/docs/_includes/layouts/pages/elements.njk
+++ b/docs/_includes/layouts/pages/elements.njk
@@ -69,9 +69,11 @@ eleventyComputed:
     </uxdot-header>
 
     {% if hasToc and (content | toc).length > 0 %}
-    <uxdot-toc summary="On this page">
-      {{ content | toc | safe }}
-    </uxdot-toc>
+    <aside>
+      <uxdot-toc summary="On this page">
+        {{ content | toc | safe }}
+      </uxdot-toc>
+    </aside>
     {% endif %}
 
     <div class="container">
--- a/docs/_includes/layouts/pages/pattern.njk
+++ b/docs/_includes/layouts/pages/pattern.njk
@@ -23,9 +23,11 @@ importElements:
 
   {% include 'partials/component/header.njk' %}
   {% if hasToc and (content | toc).length > 0 %}
-  <uxdot-toc summary="On this page">
-    {{ content | toc | safe }}
-  </uxdot-toc>
+  <aside>
+    <uxdot-toc summary="On this page">
+      {{ content | toc | safe }}
+    </uxdot-toc>
+  </aside>
   {% endif %}
 
   <div class="container">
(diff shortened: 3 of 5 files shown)


bennypowers:
Can we get the same benefits without lifting markup and styles into the lightdom? Maybe use a complementary role on the host?

zeroedin:
> Can we get the same benefits without lifting markup and styles into the lightdom? Maybe use a complementary role on the host?

Maybe, the idea originally was to only have 1 aside, but if we lock the aside into the shadowroot then we also need to provide `aria-label` prop on the toc.  Also we were using it as a container for styling which is why the padding got wonky when it was moved into the shadow.  

What real benefit does it serve to move it or use a complementary role?  I'd personally rather just use the semantic tag here so we have the `<main><article><aside>` cascade.  The toc is not IMHO an aside, it is the content of an aside. YMMV

bennypowers:
> need to provide `aria-label` -> `accessible-label` prop on the toc. 

Not terrible, we're talking about <= 2 call sites

> a container for styling which is why the padding got wonky when it was moved into the shadow.

I'm confident that with some juggling of styles this can be resolved all in shadow styles. one approach could be to make the host the container, and put the complementary role on it

> What benefit

One stylesheet to rule them all. when issues come up with the toc, i want to solve them either in the css or in the template, encapsualted, abstracted away, with as little leakage into other moving parts as we can muster

zeroedin:
> i want to solve them either in the css or in the template, encapsualted, abstracted away, with as little leakage into other moving parts as we can muster

I feel like the way I have it already abstracts this given the separation of concerns of the semantics tags which are then styled for the layout.  The toc is only then inserted into that layout.  Maybe this is a Potaaaayto Potaaaaatoe... thing though.  I'll explore

zeroedin:
Given a conversation with @adamjohnson on this we determined due to [Best Practices](https://dequeuniversity.com/rules/axe/4.10/landmark-complementary-is-top-level?application=AxeEdge) it was better just to remove the `<aside>` all together.  The `aria-labelledby` in the shadowroot of the uxdot-toc already provides enough surface for assistive technology and its best to avoid using aside inside the `<main>` landmark all together.

https://html.spec.whatwg.org/multipage/sections.html#the-aside-element
> The element can be used for typographical effects like pull quotes or sidebars, for advertising, for groups of [nav](https://html.spec.whatwg.org/multipage/sections.html#the-nav-element) elements, and for other content that is considered separate from the main content of the page.

A heavy refactor would be needed to property incorporate a semantic usage of aside throughout the site given current layout constraints.

bennypowers:
sounds good to me

can we lose the `<div class="aside">`? the host element should be able to handle the whole kebab

zeroedin:
> can we lose the <div class="aside">? 

that is needed for `grid-area: aside`.  I'd rather not do that on the host element if we want to add another element to that region of the layout.

https://github.com/RedHat-UX/red-hat-design-system/blob/4ae9fd85ce42f3a89d069a747e0872343add99d3/docs/styles/pages/backpage.css#L60-L75

bennypowers:
In that case put aside class on the host

zeroedin:
> In that case put aside class on the host

My preference is to have a container here still.

```
<article>
 <div class="aside"><div class="container">
</article>
```

Reasoning:
If not for the best practices for screen readers not yet handling the [5.1 html spec of aside ](https://html.spec.whatwg.org/multipage/sections.html#the-aside-element)inside other landmarks, semantically the aside element would IMHO make the most sense.  

My argument/position here is that the container `.aside` is separate from the toc itself.  The container  `.aside` could contain MORE than the just the toc as a child in the future and therefore the layout shouldn't be tied specifically to the toc host element.  In an attempt to make content abstracted layout styles.

If you feel that strongly in opposition to that position, please feel free to push the commit and resolve.

bennypowers:
Fine by me

PR #1822, opened 2024-09-04
docs: sidenav height and scrolling

bennypowers:
## What I did

1. fixes background surface for sidenav. without the fix, it's transparent below the fold
2. makes sidenav vertically scrollable

### DIFF
--- a/docs/assets/javascript/elements/uxdot-sidenav.css
+++ b/docs/assets/javascript/elements/uxdot-sidenav.css
@@ -18,7 +18,7 @@
   display: none;
 }
 
-[part='close-button'] {
+[part="close-button"] {
   color: var(--rh-color-text-primary-on-light);
   background-color: transparent;
   border: none;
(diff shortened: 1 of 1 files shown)


docs/assets/javascript/elements/uxdot-sidenav.css
```diff
@@ -32,7 +31,7 @@
   position: relative;
   background-color: var(--rh-color-surface-lightest);
   width: auto;
-  height: 100%;
```
zeroedin, line 35: I think you still need a `height: 100dvh` here.

Before:
Screenshot 2024-09-04 at 8 58 51 AM
After:
Screenshot 2024-09-04 at 8 59 06 AM

bennypowers: how's that now?

zeroedin (approved):

PR #1825, opened 2024-09-04
docs: fix repoStatus naming convention

zeroedin:
## What I did

1.  Align components names to match the naming convention used in the subnav.
2.  Fixes missing rh-tag's in subnav when a component is still in planning.

## Testing Instructions

1.

### DIFF
--- a/docs/_data/repoStatus.yaml
+++ b/docs/_data/repoStatus.yaml
@@ -2,7 +2,7 @@
 # Creating repoStatus data for a new component
 #
 # - tagName: rh-jazz-hands 
-#   name: Jazz hands  - name should match the same format as sidenav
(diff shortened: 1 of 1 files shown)


bennypowers (commented):
We have utilities to convert from tagname to slug to pretty alias

Feels like this table- or parts of it- should be a function call

zeroedin:
> We have utilities to convert from tagname to slug to pretty alias
> 
> Feels like this table- or parts of it- should be a function call

Can we split that off to a different issue.  This resolves the problem for the immediate.

adamjohnson (approved):
The guidelines are an awesome addition 👏 

Links Get Tagged (with) Meaning

adamjohnson:
Related to #1734.

PR #1845, opened 2024-09-09
chore: lit css nesting

bennypowers:
## What I did

1. remove `@patternfly/pfe-tools/typescript/transformers/css-imports.cjs`
2. add `typescript-transform-lit-css`

## Testing Instructions

1. make some nested css in a shadow style
2. build clean
3. ensure that *all* the file css is legit in the output

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,13 +1,12 @@
 {
   "name": "@rhds/elements",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@rhds/elements",
-      "version": "2.0.0",
-      "hasInstallScript": true,
+      "version": "2.0.1",
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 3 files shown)


zeroedin (commented):
Looks to have some issues:
```css
:host {
  --_border-color: transparent;
  
  &[color-palette='dark'] {
    --_border-color: var(--rh-color-border-subtle-on-dark, #707070);
  }
}
```

outputs as:
```css
:host {
  --_border-color: transparent;
}

[color-palette="dark"]:host {
  --_border-color: var(--rh-color-border-subtle-on-dark, #707070);
}
```

flips the attribute value in front of host.

bennypowers:
```css
:host {
  --_border-color: transparent;
  
  &[color-palette='dark'] {
    --_border-color: var(--rh-color-border-subtle-on-dark, #707070);
  }
}
```

This would need to be 
```css
:host([color-palette='dark']) {...}
```
so the output above is correct as far as the transform goes.

PR #1854, opened 2024-09-11
fix(audio-player): theming, icon, playback rate

bennypowers:
## What I did

1. remove `-on-` tokens from player
2. use rh-icon everywhere in player
3. refactor css to use nesting and shadow classes (TODO in another pr: css grid)
4. add playback rate controls to mini layout menu

# Testing Instructions
Compare to clefairy DP

# Notes to Reviewers
Let's keep the review focused on bugs related to this PR. Unrelated issues please open up in a new ticket


elements/rh-audio-player/rh-audio-player-rate-stepper.css
```diff
@@ -0,0 +1,48 @@
...
+  border: none;
+  background: transparent;
+  height: var(--_button-size, 40px);
+  min-width: var(--_button-size, 40px);
+  padding: 0;
+}
+
+select:focus,
+button:focus {
+  outline: var(--_outline);
```
adamjohnson, line 39: This var is undefined in `rh-audio-player.html` and focus is lost on each one of these selector's controls. A11y issue.

elements/rh-audio-player/rh-audio-player.css
```diff
@@ -448,337 +443,294 @@ rh-menu > button:focus {
...
-#full-playback-rate {
-  appearance: none;
-  line-height: 36px;
-  padding: 2px 0;
-  border: none;
-  color: var(--_text-color);
-  background-color: var(--_surface-color);
+#play,
+#full-play {
+  color: var(--rh-audio-player-icon-background-color, var(--_surface-color));
```
adamjohnson, line 448: This causes the play button in `rh-audio-player.html` to be invisible (white foreground, white background).

adamjohnson (changes requested):
The decrease and increase playback speed buttons inside the dropdown are not focusable / usable to keyboard + screen reader users. 

rh-audio-player playback controls screenshot

It does appear that keyboard and SR users are able to activate the `<select>` to change the playback rate. All three of those controls should be accessible to keyboard and SR users. 

Would it be possible to have the playback rate in the menu at the smallest screens? Then, at 576px+ move the playback rate controls in between the volume and menu buttons?

Alternatively, hide the playback rate at the smallest screen sizes (far from ideal). Then show it when there's room.

marionnegp (changes requested):
For all screenshots, I have the DP for Clefairy on the left and the DP for this PR on the right.

- [x] play button isn't showing up in the full screen demo
Screenshot 2024-09-16 at 9 39 15 PM
---
- [x] play button isn't showing up when I make the transcript visible
- [ ] border around transcript isn't showing up
Screenshot 2024-09-16 at 9 37 13 PM
---
- [ ] scrollbar isn't showing up for the "About the episode" view
Screenshot 2024-09-16 at 9 38 15 PM
---
@coreyvickery, how do you feel about the contextual menu icon being included when the content panel is open? I actually like having it there, but this isn't what's in Figma or the doc images. Wasn't sure if there's a reason it wasn't added previously.
Screenshot 2024-09-16 at 9 42 41 PM

bennypowers:
@marionnegp PTAL, should be fixed now

@adamjohnson it seems to me there are two possible solutions here
1. disable or remove RTI from menu (perhaps by opting out with a special property?)
2. add the stepper controls to the RTI for the menu, again perhaps with a special property

I'll try the second approach first

elements/rh-audio-player/rh-audio-player-rate-stepper.css
bennypowers: i think this is resolved now PTAL

coreyvickery:
> @coreyvickery, how do you feel about the contextual menu icon being included when the content panel is open? I actually like having it there, but this isn't what's in Figma or the doc images. Wasn't sure if there's a reason it wasn't added previously. Screenshot 2024-09-16 at 9 42 41 PM

@marionnegp I don't think I mind, but I'd like to discuss further. Maybe we revisit for `Cubone` or `Diglett`.

PR #1856, opened 2024-09-11
docs: block padding in uxdot sidenav 

zeroedin:
## What I did

1. readding missing block padding present in uxdot sidenav spec

## Testing Instructions

1.


zeroedin:
> You might consider 16px/LG on mobile. YMMV.

I'll go back and check what the design specs are

PR #1862, opened 2024-09-12
chore: align style use between docs and demo

zeroedin:
## What I did

1.  Clean up how we are calling different stylesheets across the use cases.   Ensure's tokens are included in demos and docs as a layer in the main stylesheets for both in a modular approach.  Now a single source of truth for both use cases which is more cleanly organized and can be extended using layers.

## Testing Instructions

1. View in localhost:8000 and doc demos localhost:8080 demos should now have the same styles across both environments.


bennypowers:
Im ok with global tokens style sheets in the demos, but nothing more

Users should be able to copy demo contents onto a blank page and get results

zeroedin:
> Users should be able to copy demo contents onto a blank page and get results

This changes nothing about that, this is more just house keeping on our end.

docs/styles/demo/styles.css
```diff
@@ -1,13 +1,15 @@
...
   typography,
   base,
   layout,
   third-party;
 
 @import url('../reset.css') layer(reset);
 @import url('../fonts.css') layer(fonts);
+@import url('/assets/packages/@rhds/tokens/css/global.css') layer(tokens);
 @import url('../typography.css') layer(typography);
 @import url('../third-party/prism-rhds.css') layer(third-party);
```
zeroedin, line 14: @bennypowers I can't remember why we loaded prism into the demo styles.  This maybe something we can remove, especially with changes to codeblock, can you verify?

zeroedin:
> Im ok with global tokens style sheets in the demos, but nothing more

@bennypowers PTAL a look again, this last commit might have clarified stuff a bit more.

zeroedin:
Just realized this the part where i run into the dev-server trying to call global.css from the node_modules and gets wrapped in js...

.pfe.config.json
```diff
@@ -16,7 +16,7 @@
     "description": "Red Hat Design System",
     "componentSubpath": "elements",
     "stylesheets": [
-      "/docs/styles/demo/styles.css"
+      "/docs/styles/demo/dev-server.css"
```
bennypowers, line 19: We can lose patch on pfe this way, right?

bennypowers (approved):

.pfe.config.json
zeroedin: I think I'd like to break this out to another issue.  IMHO I'd still like to make this configurable upstream, both the html and the CSS.  We should be able to control/override all the CSS we want downstream.

zeroedin: https://github.com/patternfly/patternfly-elements/issues/2846

Issue #1238, opened 2023-09-26
[bug] <rh-dialog>


heyMP:
### Describe the bug

Triggering an rh-dialog component to "open" via a keyboard, the focus doesn't correctly transition from the trigger button to the open dialog.  The focus remains on the trigger.

### Which Component?

rh-dialog

### Steps to reproduce

1. Go to https://ux.redhat.com/elements/dialog/demo/
2. Using your keybaord, tab to the first "open" button
3. Hit "Enter"
4. Verify that the focus has remained on the "open"button

### Expected behaviour

The focus should move the the dialog element when the trigger is clicked and focus moved back to the trigger button when the dialog is closed

### Screenshots

![Screenshot 2023-09-26 at 10 31 02 AM](https://github.com/RedHat-UX/red-hat-design-system/assets/3428964/413d821d-2523-415e-81f7-db53b9d38e7f)
![Screenshot 2023-09-26 at 10 31 08 AM](https://github.com/RedHat-UX/red-hat-design-system/assets/3428964/6a7b5e5b-1c20-4d90-b885-e29246c997b1)

### Operating System (OS)

macOS

### Browser

Chrome

heyMP:
I have a fix from rhpt-elements that I can put into a PR.

bennypowers:
we should just use `<dialog>` internally https://caniuse.com/dialog

brianferry:
@bennypowers - Switching to dialog internally was going to be the follow up https://github.com/RedHat-UX/red-hat-design-system/issues/1242

bennypowers:
`<dialog>` provides browser-built-in focus trap. should we spend time implementing that in javascript now then immediately turn around and delete that code?

markcaron:
@adamjohnson can you look into this for Clefairy?

markcaron:
Seems like this may now be something that could be fixed in #1242

adamjohnson:
Focus is correctly applied with the implementation of the native `<dialog>` in #2078.

PR #1865, opened 2024-09-13
feat(dialog): use native HTML dialog element

### LINKED ISSUE #1242: [feat] `<rh-dialog>`: switch to native <dialog> element
### Description

`<rh-dialog>` should be set up to use the native [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element if possible and decouple from using the `PfModal` Base Class. 

### Suggested solution

Switch over the code and the styles from PfModal to be in the `<rh-dialog>` element first of all, then convert it to a native `<dialog>` html element inside the shadow dom.

### Example API

```HTML
Same slots and api as currently exists, just updated internals
```

### Additional context

_No response_

adamjohnson:
## What I did

1. Change internal structure of `rh-dialog` to use the native HTML `<dialog>` element.
2. Overhaul much of the CSS.
3. Change how we handle `#overlay` to use the CSS native `::backdrop` psuedo-element. 
4. Left notes about deprecating `#overlay`
5. Added the `accessible-label` property for when users want to manually label their modals + added docs.
6. Modal now has logic to decide when to use `accessible-label`, `aria-labelledby` + IDREF, and the `aria-label` of the trigger element.
7. Improved layout of modal contents when scrollbars are present
8. Increased the touch target of the close modal button.
9. Trap focus inside the modal for improved accessibility (WIP).

## Testing Instructions

1. Visit [the DP demo](https://deploy-preview-1865--red-hat-design-system.netlify.app/elements/dialog/demo/). 
2. Add all kinds of content to the modal header, body, and footer, see if you can break it.
3. See if focus trapping works with various different interactive elements (or none at all).

## To Do's

- [x] Fix Shift + Tab focus trapping
- [ ] Fix focus trap with Tab in video modal with YouTube embed
- [ ] Update/improve changeset
- [ ] Test across different browsers and AT
- [x] Look into why trigger/`rh-button` displays two buttons (one without a label) in the DP but not locally

## Notes to Reviewers

Some docs on the Guidelines page are incorrect as they have the modal body scrolling independently from the `header` and `footer`. That needs updated. 

WIP right now. Come back later for a more accurate description. 😇 

Closes #1242 and #1238.


adamjohnson:
## Current status

With a _lot_ of help from @zeroedin, he and I were able to get <kbd>Shift + Tab</kbd> mostly working for `<rh-dialog>`. Check out [the DP](https://deploy-preview-1865--red-hat-design-system.netlify.app/elements/dialog/demo/).

## Setbacks

Just when we thought we had it cracked, we added an `<rh-tabs>` component to the body of the `<rh-dialog>` and found that it did not trap focus as intended if an `<rh-*>` element is the last focusable item in the dialog.

## Possible solutions

We suspect that we'd have to add each focusable `<rh-*>` element to the focusable selectors list and possibly add `delegatesFocus` to each of those elements. We'd then have to maintain this focusable selectors list. 

_OR_ we could explore implementing [a function like this](https://southleft.com/insights/design-systems/focus-trap-accessible-web-components/) that recursively traverses the Shadow DOM looking for focusable elements. Good to know we're not the first people to run into complexities with web components and focus traps.

_OR_ we could add one of the [off the shelf solutions](https://github.com/ment-labs/focus-trap) as a dependency.

If anyone has strong feelings about any of these solutions, weigh in!

## Video dialog + YouTube

Note that, on a video dialog with a YouTube embed, <kbd>Shift + Tab</kbd> works, but <kbd>Tab</kbd> does not trap focus. [DP Video dialog demo](https://deploy-preview-1865--red-hat-design-system.netlify.app/elements/dialog/demo/video-modal/)

NB: Right now on uxdot for a dialog + YouTube embed, focus trapping works with <kbd>Tab</kbd> but not <kbd>Shift +  Tab</kbd>. 🔁 [uxdot production video demo](https://ux.redhat.com/elements/dialog/demo/video-modal/)

## APG Guidelines vs. browser implementations

It's also worth noting that when a `<dialog>` is open, it does not allow other elements on the page to be focused. APG guidelines explicitly define the [required keyboard behavior](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#keyboardinteraction); however, the native `<dialog>` element does not follow this guidance. Accessibility practitioners are split as to which behavior is preferrable.

bennypowers (commented):
Does the surface need to be in the outside because of the overlay? Can we fix that with CSS and put the surface inside the dialog?

If we're using `<dialog>`, why do we still need kitty's focus trap javascript? Should we just call showModal by default?

Are we capturing and exposing the return value?

adamjohnson:
Will look into moving `<rh-surface>` inside the dialog. 👍 

> If we're using `<dialog>`, why do we still need kitty's focus trap javascript? Should we just call showModal by default?

We are calling `showModal()` on line 370. [APG keyboard interaction guidelines](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#keyboardinteraction) clearly state:

<blockquote>
<ul>
<li>Tab:<ul>
<li>Moves focus to the next tabbable element inside the dialog.</li>
<li>If focus is on the last tabbable element inside the dialog, moves focus to the first tabbable element inside the dialog.</li>
</ul>
</li>
<li>Shift + Tab:<ul>
<li>Moves focus to the previous tabbable element inside the dialog.</li>
<li>If focus is on the first tabbable element inside the dialog, moves focus to the last tabbable element inside the dialog.</li>
</ul>
</li>
<li>Escape: Closes the dialog.</li>
</ul>
</blockquote>

Again, browser's have implemented `<dialog>` to allow Tab access to things like other tabs and the URL bar instead of truly focus trapping the dialog.

adamjohnson:
Closing this PR due to merge conflicts. Continue this discussion in #2078.

PR #1880, opened 2024-09-17
docs: icon design update

### LINKED ISSUE #1870: [docs] Icon page update
### Description

This is an update to the formatting and copy for the Icon page made in #1844.

What do you think about using [this layout](https://www.figma.com/proto/L3wKkxQafEmkO1hoG1m4tt/Foundations---Icons?node-id=0-1&t=pmyK6H7lgMA8T7Os-1)? It keeps the icons pretty dense, but using link button without the icon inside the button helps with scanning imo. I also bumped up the size of the icons to make them easier to see.

Other changes:
- I think this should be added to the "Foundations" section. I'm not sure this warrants being a top-level page, and PatternFly also has an "Icons" page under their "Design foundations" section. 
- Added an "Overview" section with info about how to use the buttons on the page and links to `<rh-icon>` docs.
- Removed copy under each section heading. I'd like to avoid having to update usage info on two different pages.
- Added cross links that most of our pages have at the bottom.

### Links

_No response_

### Screenshots

_No response_

### Additional context

@bennypowers, feel free to unassign yourself if you have too much on your plate! I only added you because you made the original version.

@markcaron @hellogreg, I added this to the Clefairy milestone to update what's already been merged to `staging/clefairy`.

bennypowers:
## What I did

1. implement designs from #1870 

Closes #1870


marionnegp (changes requested):
- [ ] Do we have the ability to left align buttons that wrap to a second line? Alignment looks good for other buttons.
Screenshot 2024-09-17 at 12 52 14 PM

- Do we have the ability to adjust the order of the sets? The `<rh-icon>` page uses the following order: standard, UI, microns, social media. This is not a very important edit to complete for Clefairy imo, if it's a heavy lift.

marionnegp (changes requested):
@bennypowers, I pushed a commit to update text only. One last thing that @coreyvickery caught:

- [x] Can the Foundation nav dropdown show the pages in alphabetical order? I can make this a separate issue if necessary.

PR #1881, opened 2024-09-17
fix(table): add container query support

### LINKED ISSUE #1476: [feat]: add container query styles to `<rh-table>`
Tables could benefit from container style queries.  We might want implement either a feature set that allows you to switch between basing the style switch off @media or @container

zeroedin:
## What I did

1. Switched from `@media` to `@container` in lightdom.css
2. Nested CSS
3. Switched to logical properties 

Closes #1476 

## Testing Instructions

1. View [deploy preview](https://deploy-preview-1881--red-hat-design-system.netlify.app/)
2. Please verify styles haven't been broken here given nesting of the CSS.  
3. Consider the implications of the @container query vs @media query.  Should we adjust the breakpoint?  Right now it is set at 768px.  That however maybe too large considering some content containers?  Should this drop to 576px instead now that we are taking into account only the width of the host element?


marionnegp (changes requested):
- I agree that 768px feels a little large, but I'd defer to what @coreyvickery thinks because he's been working with JVP's team on a unified grid. He may have some insight into whether they're using certain container breakpoints.

- [ ] In the "Row And Column Headers" demo, can the row header cells be left aligned on mobile?
Screenshot 2024-09-17 at 1 05 54 PM

- [ ] Do the headings for the "Column Headers" and "Row Headers" demos have to be switched?

- [ ] `padding-block` for cells should be `var(--rh-space-lg, 16px)`

- [ ] This is nitpicky, but can we use sentence case for demo headings?

Overall, styling looks like it's in really good shape!

coreyvickery:
> I agree that 768px feels a little large, but I'd defer to what @coreyvickery thinks because he's been working with JVP's team on a unified grid. He may have some insight into whether they're using certain container breakpoints.

@marionnegp @zeroedin I don't remember if we've gotten that far yet. Maybe we revisit when JVP, I, and team have something more concrete.

bennypowers:
Can we defer the discussion of the container width to cubone and merge this without overt design changes?

zeroedin:
> * I'd defer to what @coreyvickery thinks because he's been working with JVP's team on a unified grid.

I think this has less to do with grid and more to do with the component content itself.  Different table data is going to need to be handled differently with additional CSS by the implementation anyways.  A large table with many columns wide may want to break earlier.  Where a smaller column table much later and maybe not at all if the content fits on small viewports.

bennypowers:
@marionnegp please open new issues against cubone for those helpful notes above, as well as for the discussion on container width and content length

bennypowers (approved):

PR #1905, opened 2024-09-22
feat(alert): dpo tokens

bennypowers:
## What I did

1. use DPO tokens in alert
2. deprecate `info`, `default`, and `error`

## Testing Instructions

1. check out the DP, make sure colours are all as expected


brianferry:
On both the demos tab and the individual demo page for the deprecated states I am seeing them all just being shown as `default` styles though the state does look correct in the markup.  

On the demos tab I am seeing the first two examples with headers but not the third example.

image

On the individual demo page I am seeing all three but for some reason the header slot isn't showing.  (Actually, it looks like this is the case for every individual demo page?)

image

It appears that perhaps after the state has been updated the component needs to be rerendered to get the updated styles?

brianferry:
I think this line

`<header ?hidden="${this.#slots.isEmpty('header')}">` in `rh-alert.ts` might need to be `<header ?hidden="${this.#slots.hasSlotted('header')}">` ? wdyt?

bennypowers:
> On the demos tab I am seeing the first two examples with headers but not the third example.
> 
> image

This is expected - at the moment, the playgrounds can only get the most recent *published* version - they can't access the local builds.

> On the individual demo page I am seeing all three but for some reason the header slot isn't showing. (Actually, it looks like this is the case for every individual demo page?)
> 
> image
> 
> It appears that perhaps after the state has been updated the component needs to be rerendered to get the updated styles?

There are some SSR shenanigans here which can hopefully be resolved by importing dependencies.

brianferry (approved):
@bennypowers That makes sense to me, code looks good, great job!

marionnegp (changes requested):
- [x] Neutral alerts should be using grays.
Screenshot 2024-09-25 at 11 24 40 AM
(screenshot is from [this Figma file](https://www.figma.com/design/FAhXoct6VdXsoRDkwHBLMK/Status-colors?node-id=193-525&m=dev))

---

- [ ] Most alert headings need to use crayon colors for now. I didn't add status tokens for typography because using non-gray colors for alert headings is something that we _might_ want to phase out in a later release. For now, here are the  colors for each state:
- Danger: `--rh-color-red-orange-70`
- Warning: `--rh-color-yellow-70`
- Caution: `--rh-color-orange-70`
- Neutral: `--rh-color-text-primary-on-light`
- Info: `--rh-color-purple-70`
- Success: `--rh-color-green-70`

These would apply for both light and dark theme and to headings for inline and toast alerts.

---

**Notes but not requested changes:**

- The teal color scheme currently using can be used when there's no state attached to an alert to avoid breaking anything. (I think that's how alert is already styled, so this is just FYI!)

- I'm seeing the same style for both `danger` and `error` in toast alerts. I think this is fine if we're going to remove the `error` state in a future release. (I'm not sure how the two would be used differently, and we don't show an error state in our [alert severity guidelines](https://deploy-preview-1905--red-hat-design-system.netlify.app/elements/alert/guidelines/#severity).)
Screenshot 2024-09-25 at 11 21 53 AM

bennypowers:
@marionnegp OK i made your changes so this is ready to merge. Please consider using relative-color-in-oklch for the headings, either from the border tokens or from the surface tokens, then we could easily apply color palettes and theming

marionnegp (approved):

PR #1941, opened 2024-09-30
docs: installation tabs code tabs

bennypowers:
## What I did

1. use code-block tabs pattern for installation tabs
2. add copy and wrap actions to most codeblocks
3. remove the code.md files which were only boilerplate
4. remove renderCodeDocs and renderInstallation shortcodes
5. make the code page an 11ty.cjs template, including code docs, installation, lightdom, and shims
6. automatically detect lightdom and shims
7. render usage codeblock based on content of primary demo.

## TODO
- [x] ensure table of contents works (probably need to abstract out a "has toc" template from base.njk)


bennypowers:
This should also handle light DOM shims instructions as well

PR #1943, opened 2024-09-30
fix(card): computed context

zeroedin:
## What I did

1. Compute context.  First check if `this.colorPalette` is set, if so use that as basis for `light|dark` if undefined check to see if `this.on` is set, if so use that as basis, if not default to light.

## Testing Instructions

1. [View Deploy preview](https://deploy-preview-1943--red-hat-design-system.netlify.app)

## Notes to Reviewers

Ensure color-context and card patterns function as expected.

### DIFF
--- a/elements/rh-card/rh-card.ts
+++ b/elements/rh-card/rh-card.ts
@@ -101,10 +101,14 @@ export class RhCard extends LitElement {
   }
 
   override render() {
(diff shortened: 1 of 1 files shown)


bennypowers:
Note: we're planning to reevaluate this after Clefairy release, to see if there's a better solution in the context system, or if this is genuinely a card-specific problem, we can leave as is

PR #1971, opened 2024-10-08
docs: ssr for uxdot-pattern

bennypowers:
## What I did

1. remove the `{% uxdotPattern %}` shortcode
2. reimplement the lit-ssr eleventy plugin
3. patch lit/context so that it runs to completion in ssr
4. patch TabsAriaController for ssr-ability
5. add a serverside controller to uxdot-pattern which reads, highlights, and renders codeblocks and content
6. modify the design of uxdot-pattern to more closely resemble the playgrounds mockup

## Testing Instructions

1. check out any of the patterns of theming pages - see "Files Changed" for relevant pages


docs/assets/javascript/elements/uxdot-pattern.css
```diff
@@ -1,89 +1,96 @@
...
-  grid-template-columns: 1fr;
-  grid-template-areas: 'controls' 'example' 'code';
-  gap: var(--rh-space-2xl, 32px) var(--rh-space-4xl, 64px);
-  padding: var(--rh-space-2xl, 32px);
-  border: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle);
-  border-radius: var(--rh-border-radius-default, 3px);
-  background-color: var(--rh-color-surface);
-  color: var(--rh-color-text-primary);
-  border-color: var(--rh-color-border-subtle);
+  grid-template-columns: auto auto;
```
zeroedin, line 11: we only have single column at this point.  I believe this should be the following: 
```Suggestion
grid-template-columns: auto;
```

docs/assets/javascript/elements/uxdot-pattern.css
```diff
@@ -1,89 +1,96 @@
...
+  grid-template-areas: 'heading' 'controls' 'content' 'code';
+  gap: var(--rh-space-2xl) var(--rh-space-4xl);
 
-  --rh-color-surface: inherit;
-}
+  @container host (width >= 530px) {
+    grid-template-areas:
```
zeroedin, line 16:
```Suggestion
    grid-template-columns: auto auto;
    grid-template-areas:
```
follow up to previous line fix

docs/assets/javascript/elements/uxdot-pattern.css
```diff
@@ -1,89 +1,96 @@
...
+      'heading controls'
+      'content content'
+      'code code';
```
zeroedin, line 19:
```Suggestion
    grid-template-columns: auto auto;
    grid-template-areas:
        "heading controls"
        "content code"
        "content code";    
```
follow up to previous line fix

docs/assets/javascript/elements/uxdot-pattern.css
```diff
@@ -1,89 +1,96 @@
...
-#container ::slotted(h6) {
-  margin-block: 0 var(--rh-space-lg, 16px) !important;
-  font-size: var(--rh-font-size-body-text-lg, 1.125rem) !important;
+  @container host (width >= 992px) {
+    grid-template-areas:
+      'heading heading'
+      'controls controls'
+      'example example'
+      'code code';
+  }
```
zeroedin, line 28: given the changes above might not need this additional breakpoint not sure what grid-area `example` is so maybe it comes as another element not yet present?

docs/assets/javascript/elements/uxdot-pattern.css
```diff
@@ -1,89 +1,96 @@
...
+  grid-area: heading;
   display: flex;
   align-items: center;
-  gap: var(--rh-space-lg, 16px);
 }
 
-#example {
-  grid-area: example;
-}
+#description { grid-area: content; }
```
zeroedin, line 37: likely this was meant to be:
```Suggestion
#description { 
  grid-area: description; 
}
```

docs/assets/javascript/elements/uxdot-pattern.css
```diff
@@ -1,89 +1,96 @@
...
+  border-radius: var(--rh-border-radius-default);
+  background-color: var(--rh-color-surface);
+  color: var(--rh-color-text-primary);
+  border-color: var(--rh-color-border-subtle);
 
-#container.noColorPicker {
-  grid-template-areas: 'example' 'code';
+  @container host (width < 300px) {
+    padding: var(--rh-space-lg);
+  }
```
zeroedin, line 63: I'm thinking here we'd want to base this off the more narrow container so maybe we need a wrapping div around the `rh-surface` called `#surface-container` and `{ container: surface / inline-size }` ?

docs/assets/javascript/elements/uxdot-pattern.css
zeroedin: might also need a `display: none` on it unless something is slotted and the grid is setup to handle the display somewhere which currently isn't.  but the duplicate content grid-areas were also conflicting

docs/assets/javascript/elements/uxdot-pattern.css
zeroedin: we may way to have a intermediate between 0 and >= 530 px before this to move to 2 columns when the heading and controls can fit on the same line but before we break out to two column for the whole component.

docs/_plugins/lit-ssr/lit.cjs
zeroedin, line 1: formally requesting a more in depth walk through of these tweaks you have made that may not be visible just in the diff.  👔 🚶‍♂️

docs/assets/javascript/elements/uxdot-pattern.css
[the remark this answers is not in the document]
bennypowers: from discussions with @markcaron: always single column, narrow elements centered.

docs/_plugins/lit-ssr/lit.cjs
bennypowers: three changes from stock lit plugin:
- strip out the `mode: 'vm'` code, which lit team may anyways remove
- use `collectResult`, which the plugin does not (as it predates it)
- hardcode in an elementRenderer for uxdot pattern, which does 2 things necessary for this pattern:
  1. passes in eleventy page data to the element
  2. waits on the async work to read and highlight pattern html files

step 2 is accomplished via the cumbersome RenderResult api, which is a sync iterable which might contain promises, or nested RenderResults. we can't just return a promise of a templateresult, which means we have to do all the async work upfront and yield the promise in the iterator before we get to the actual element render. UxdotPatternSSRControllerServer/Client are factored to store the results from that async work on the class fields.

at that point you have to play whack-a-mole with hydration mismatch errors, which in this case had two parts:
1. no ternaries in the template (we'll have to disable the empty tabs)
2. make sure the classfields on the controller are the DirectiveResults returned from unsafeHTML, not the string parameters thereto. I guess that's so that reference equality is maintained in the hydration check?? actually no clue... but it works.

docs/assets/javascript/elements/uxdot-pattern.css
bennypowers: why do we need that? i think that would just add complexity. We only have one vertical layout, and all we need to do is adjust padding to suit smaller containers (IMO)

bennypowers: [6a6eabc](https://github.com/RedHat-UX/red-hat-design-system/pull/1971/commits/6a6eabc1072356fb05337ad7441b314b7296934b)

bennypowers:
Reviewed on call with @zeroedin

PR #1975, opened 2024-10-09
docs: reapply "docs: installation tabs code tabs (#1941)"

bennypowers:
This reverts commit 5123961c301525efb20d56b89edeb8e23d452c2b.


docs/_data/repoStatus.yaml
```diff
@@ -231,7 +231,7 @@
       status: N/A
     - name: Documentation
       status: Ready
-- tagName: rh-navigation-primary      
+- tagName: rh-navigation
```
zeroedin, line 234: I thought we had discussed at one point that the `rh-navigation` would still have the `-primary` on it?  maybe I'm mistaken. @markcaron

docs/_includes/partials/component/footer.njk
```diff
@@ -1,3 +1,8 @@
+<script data-helmet type="module">
+  import '@rhds/elements/rh-footer/rh-footer-universal.js';
+  import '@rhds/elements/rh-footer/rh-footer-copyright.js';
+</script>
```
zeroedin, line 4: These kinds of changes that move these deps to the file they are needed in are 👨‍🍳 💋

zeroedin (approved):
Little Glitches Trigger Mayhem

docs/_data/repoStatus.yaml
markcaron: @zeroedin I believe that was the decision, yeah.

PR #1976, opened 2024-10-10
docs(home): fixing get started tiles

markcaron:
## What I did

1. Put the images in the `image` `slot` in Get Started tiles on the Docs Home, so that it looks better.

## Testing Instructions

1. Does it look cool?

### DIFF
--- a/docs/index.njk
+++ b/docs/index.njk
@@ -40,6 +40,7 @@ title: Home
     <div class="grid xs-two-columns">
       <rh-tile>
         <img src="/assets/home/foundations.svg"
+            slot="image"
             alt="Foundations"
             width="300"
             height="131">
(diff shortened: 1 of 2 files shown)


adamjohnson (changes requested):
Looks better but I think the following could improve it further. I would:

1. Remove all `rh-tile h2` styles targeting this area from `home.css` (margin-block-start and font-size)
1. Remove the margin-block from the slotted `rh-tile` paragraph tag via home.css: `#get-started rh-tile p { margin-block: 0; }` (or maybe just `rh-tile p`, YMMV).

[Screenshot after changes](https://github.com/user-attachments/assets/0a6c8d5b-5609-43b0-b29e-966605e0f90e).

adamjohnson (approved):
Logical gains tiles make 👏

Issue #1959, opened 2024-10-04
`<rh-health-index>` z-index set too high


zeroedin:
The `z-index` value for the current active box in health index is set to 10.  This value is static in health index and cannot be adjusted with a component level css prop.  I do not believe this value should be set this high when setting to z-index of 1 is sufficient for the internals of the component.  I think our components should use the lowest possible z-index when needed as a best practice, or provide the ability to adjust them per use case.

[This issue shows up](https://github.com/RedHat-UX/red-hat-design-system/pull/1949#discussion_r1786752898) on the ux. website where the health-index self set z-index of 10 becomes higher then any z-index used ux. and therefore overlays the lower set site z-index values.

markcaron:
Temporarily patching with #1995 until we can discuss our approach to handling `z-index` across components.

bennypowers:
Can you check this in the color-scheme pr dp? (or on main after it lands)?

markcaron:
@bennypowers this still exists in the [color-scheme PR](https://github.com/RedHat-UX/red-hat-design-system/pull/2138).

<img width="390" alt="Image" src="https://github.com/user-attachments/assets/0d311ba8-2d97-437c-87f8-2651d25f6b81" />

See z-index explicitly set to `10`:
<img width="157" alt="Image" src="https://github.com/user-attachments/assets/77c2470a-e014-4898-9e8f-766e1b32e0d1" />

Can we not fix this in the [current patch](https://github.com/RedHat-UX/red-hat-design-system/pull/1995)?

CC @zeroedin

bennypowers:
Closed by #1995

PR #1995, opened 2024-10-17
fix(health-index): prevent graphical elements breaking out of stacking context

markcaron:
## What I did

1. Added `--rh-health-index-z-index` CSS custom prop to allow implementers to patch/adjust any z-index issues they may have with `<rh-health-index>` and other elements on the page.

Quick solution to #1959


.changeset/eighty-windows-exist.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
```
bennypowers, line 2: this is a minor change

Can we fix this without exposing a new css property? maybe with position: relative on the host element?

bennypowers (changes requested):

.changeset/eighty-windows-exist.md
bennypowers: 1. this is only a problem in chrome

image

2. I think we can fix this with `position: relative; z-index: 1` on the container element, without requiring a minor release

image

markcaron:
@bennypowers nice! Thanks for fixing this.

PR #2000, opened 2024-10-18
feat(navigation-universal): add `<rh-navigation-universal>`

adamjohnson:
## What I did

1. Add the `<rh-navigation-universal>` element.

## Testing Instructions

1. View the DP for Universal Navigation:
    * [Demo](https://deploy-preview-2000--red-hat-design-system.netlify.app/elements/navigation-universal/demo/)
    * [Docs](https://deploy-preview-2000--red-hat-design-system.netlify.app/elements/navigation-universal/)
2. Compare to Figma and what exists on live properties.

## To Do's

- [ ] Add docs
- [ ] Add changeset
- [ ] Lightdom shim (?)
- [x] Add tests
- [x] Add "All Red Hat" dropdown
- [x] Fix background color for dropdown (currently not defined)
- [x] Implement default value for `accessible-label` in docs

## Notes to Reviewers

Right now, this mimics what exists on redhat.com.

  * This design is not in the Figma library. Should it exist?
    * If so, designers: make sure this component meets design specs.
  * In the Figma Library, there's 16px of padding around each link. This results in what I perceive to be much larger height than what's in production.
    * Should it be 16px? 
    * FWIW, I have it set to 8px as of this writing.
  * Double check that my values in `repoStatus.yaml` are correct.
  * Some noticeable FOUC. What should we do about that?
      * This may soon be handled by [rh-disclosure](https://github.com/RedHat-UX/red-hat-design-system/pull/2043) and DSD 
  * Should this have a dark variant (may only apply to dropdown content)?
  * Mobile:
      * On redhat.com and elsewhere, the universal nav is included inside the main navigation on mobile.
      * What should happen on mobile here? Should we dictate that? Or should end users decide how they want to handle mobile (hamburger, hide, etc)?
  * ~~Developers: added an `any` type to fix `Argument of type 'DocumentFragment | undefined' is not assignable to parameter of type 'Node'.` Would love some help getting the `#getDetailsContentHTML()` function properly typed, if applicable.~~ Resolved via sub component.


markcaron:
We should limit the design to a single `universal` nav, preferably the right-aligned nav with the borders on hover (like redhat.com). The left-aligned design competes with the RH logo—almost "blocking" the eye, IMO.

coreyvickery:
@adamjohnson Looks like we also have to also account for personalizations.

Screenshot 2024-10-28 at 11 15 25 AM

coreyvickery:
@adamjohnson Personalization being the `Large retailer delivers ...` text in the upper left corner.

adamjohnson:
@coreyvickery will the info icon (left) and/or chevron (right) change? Or are those spots always those particular icons?

coreyvickery:
> @coreyvickery will the info icon (left) and/or chevron (right) in the personalization message change? Or are those spots always those particular icons?

@adamjohnson Not to my knowledge no, but let's make sure we're using current icons from Brand for both.

PR #2010, opened 2024-10-27
docs: convert repoStatus.yaml to typescript

bennypowers:
this will help us catch bugs in our repostatus configuration

## What I did

1. convert `repoStatus.yaml` to a well typed typescript file

## Testing Instructions

1. try logging out `{{ repoStatus | log }}` in an njk template, it should print in the terminal

## Notes to Reviewers

We might later on want to adopt https://github.com/uncenter/eleventy-plugin-validate for runtime validations


zeroedin (approved):
Libraries Grouped, Typed Mechanics

PR #2016, opened 2024-10-28
docs(foundations-typography-scale): update title values

gradymcgee:
## What I did

1. Updated the `Title, lg` value to:

Font size: 20 (1.25)
Line height: 30 (1.5) 

## Testing Instructions

1. Check the foundations-typography-scale page in the deploy preview.


bennypowers:
I'm noticing just now that we don't have tokens for these values @coreyvickery

coreyvickery:
@marionnegp Benny's comment above is why I was asking you about adding tokens.

coreyvickery (approved):

PR #2039, opened 2024-11-06
fix(code-block): pre-rendered copy and word wrap actions

zeroedin:
## What I did

1. Fixed copy and wrap functionality with prerendered code-block content
2. Updated prerendered demos to include copy and wrap buttons

## Testing Instructions

1.  [View Deploy Preview demos](https://deploy-preview-2039--red-hat-design-system.netlify.app/elements/code-block/demo/prerendered-prism-highlighting/)


elements/rh-code-block/rh-code-block.css
```diff
@@ -127,6 +127,28 @@
   color: inherit;
 }
 
+#content::slotted(:is(code[class*='language-'], pre[class*='language-'])) {
```
bennypowers, line 130: Not sure this will work in all cases. Prism usually outputs a code wrapped in a pre, hence the need for lightdom css.

bennypowers (approved):
LGTM, with the proviso that lightdom css should be carefully reviewed

elements/rh-code-block/rh-code-block.css
zeroedin: It was the only way we could figure out how to pivot the styles without reflecting "wrap" back to the host element as the only switch is on the shadow container.  Otherwise we could have modified `prism.css.ts` with a `rh-code-block[wrap]` style selector and style it on this layer.

zeroedin: @bennypowers not sure how to effectively test for the outliers or when they would pop up, if you have an gumption to take a stab.

PR #2042, opened 2024-11-07
fix(button): aria disabled button

### LINKED ISSUE #2041: [bug] Disabled `<rh-button>` should still be focusable
Currently, `<rh-button>` has a `disabled` attribute which, when set to `true`, adds a `disabled` attribute to its `<button>` element.

This has the effect of making the button unreachable to keyboard users, even when `tabindex` is applied.

In order to make the button focusable, I recommend applying an `aria-disabled="true"` attribute to the `<button>` instead of `disabled`, when `<rh-button>` has the `disabled` attribute applied to it. This also means we can attach actions and/or helpful error messages when anyone (mouse or keyboard, visual or non-visual) tries to interact with it: "You must update at least one option before re-saving" or some such.

Basically...
- Using `aria-disabled` allows screen reader users to focus on it.
- The button can still _look_ disabled to visual users. 
- We can add styles like `cursor: not-allowed`, if desired.
- JS can be used to prevent clicks, if desired.

By doing the above, we wind up with what the `<button disabled>` attribute should be by default, imo.

Might want to pair this issue with #1884, which also relates to disabled `rh-button`.

If you feel like reading, here's a [good article on this topic](https://css-tricks.com/making-disabled-buttons-more-inclusive/#aa-aria-to-the-rescue).

zeroedin:
## What I did

1.  Applied `aria-disabled` to internals if button is disabled

Closes #2041 

## Testing Instructions

1. Test button using [form control demo](https://deploy-preview-2042--red-hat-design-system.netlify.app/elements/button/demo/form-control/)

### DIFF
--- a/.changeset/twelve-cups-yawn.md
+++ b/.changeset/twelve-cups-yawn.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-button>`: improved accessibility by allowing focus when disabled'
--- a/elements/rh-button/rh-button.ts
+++ b/elements/rh-button/rh-button.ts
@@ -98,15 +98,6 @@ export class RhButton extends LitElement {
     }
   }
 
-  connectedCallback(): void {
-    super.connectedCallback();
-
-    // button should be focusable when disabled
-    if (this.disabled || this.#internals.formDisabled) {
-      this.#internals.ariaDisabled = 'true';
-    }
-  }
-
   override render() {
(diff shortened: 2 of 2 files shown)


hellogreg (approved):
Los gatos tienen mojitos 
image

bennypowers:
If Greg likes it, it's fine, but I would like to point out that this does not match the behaviour of the built-in `<button>`

https://codepen.io/bennyp/pen/RwXEKvo

PR #2043, opened 2024-11-07
feat(disclosure): add `<rh-disclosure>`

### LINKED ISSUE #1293: [feat] rh-disclosure
### Description

The catalog has a need for an `rh-disclosure` element. It's currently classified as a pattern on [ux.redhat.com](https://ux.redhat.com/patterns/disclosure/), but after discussing in office hours, it's really an element. 

image

### Acceptance Criteria

- [ ] Design done
- [ ] Development done

### Image

_No response_

### Link to design doc

_No response_

### Other resources

_No response_

adamjohnson:
## What I did

1. Created the `<rh-disclosure>` element
2. Added a few tests

## Testing Instructions

1. Visit the [`<rh-disclosure>` demo](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/) on the DP.
2. Also visit the [Nested Disclosures](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/nested-disclosures/) demo.
3. [Events demo](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/events/)
4. [Color Context demo](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/color-context/)
5. [Slotted Summary demo](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/slotted-summary/)
6. [Docs](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/)

## To do's

- [x] Docs
- [x] Fix git `main` 👉 `staging/cubone` + diff
- [x] Fix border / border colors in DP

## Notes to Reviewers

~~This element is lightdom based. We're planning to use it standalone as `<rh-disclosure>` and also using it in `<rh-navigation-XYZ>` for dropdowns.~~

Shout out to @zeroedin for [the initial concept](https://codepen.io/zeroedin/pen/qBeygYB).

Closes #1293.


bennypowers:
I'm curious why you opted to make `<rh-disclosure>` a decorator. From a strictly custom-element perspective, and based on prior discussions we've had about it, "disclosure" and "details" should be interchangeable concepts, so having one explicitly depend on the user using the other is on some level confusing. Late-2000s haute cuisine trends aside, it would be a touch annoying to order a burger and have the patty, buns, veg, and sauce, all come in separate containers with assembly instructions stapled to the bad.

Tn other words, if `<rh-disclosure>` is merely a decorator element, why not just make this a pattern and recommend some global CSS rules? Can you elaborate on the decision to make user light-dom a requirement?

if downstream teams "need" lightdom details (which is in my view an antipattern) what about making it [opt-in](https://codepen.io/bennyp/pen/OJKqMQx?editors=1011)?

bennypowers:
consider [`sl-details`](https://shoelace.style/components/details):

```html
<sl-details summary="Toggle Me">
  Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna
  aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
</sl-details>
```
image

marionnegp (changes requested):
- [ ] In Safari only, I'm seeing the caret icon get duplicated. 
Screenshot 2024-11-15 at 9 58 31 AM

adamjohnson:
@bennypowers:

> Can you elaborate on the decision to make user light-dom a requirement?

So that it works with JavaScript disabled, avoids FOUC, and drives adoption. (see this [draft PR demo](https://deploy-preview-2000--red-hat-design-system.netlify.app/elements/navigation-universal/demo/) of Universal Nav with shadowdom'ed `details`/`summary` to see what the potential FOUC story looks like).

We thought that rh-disclosure might be used for `rh-navigation-{universal/primary/secondary}`. Having `detais`/`summary` in the lightdom made it easy to re-use and re-style this component for those use cases. That said, it's possible that DSD'ing those components may make this lightdom use case moot.

Your opt-in example is indeed interesting and something I'd be willing to pursue if we decide that path is our way forward. If that's not amenable, then I would probably do something similar to what [`rh-universal-nav-dropdown` does currently](https://github.com/RedHat-UX/red-hat-design-system/pull/2000/files#diff-7a179a846ba68f88ca66a5dff46fd2b1b3e04395ef25f887c737cb1f4d83cbabR23-R34), being sure to include a bespoke slot for `<summary>` in the case of JS failure. But, again, if this approach is used, FOUC is an issue short of SSR.

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -1,52 +1,47 @@
...
+
+import '@rhds/elements/rh-icon/rh-icon.js';
 
 import styles from './rh-disclosure.css';
 
 /**
  * @summary A disclosure is a widget that enables content to be either collapsed (hidden) or expanded (visible).
- * @slot - Place the `details`, `summary`, and `.details-content` elements in the default slot.
+ * @slot - Place the content you want to disclose in the default slot. This content is hidden by default.
+ * @slot summary-label - The title of the disclosure
```
bennypowers, line 12: if we had a `summary` slot but also needed a `label` for that `summary`, then I would understand the need to hyphenate, but in this instance, maybe we can get away with just `summary`, for brevity

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -1,52 +1,47 @@
...
+  @query('details') private detailsEl!: HTMLDetailsElement;
+  @query('summary') private summaryEl!: HTMLElement;
 
   render() {
     return html`
-      <slot></slot>
+      <details @keydown=${!isServer && this.#closeDetails}>
+        <summary>
+          <rh-icon part="caret" id="caret" set="ui" icon="caret-down"></rh-icon>
+          <slot name="summary-label">Panel Title</slot>
```
bennypowers, line 28:
```suggestion
          <slot name="summary-label">${this.summary}</slot>
```

elements/rh-disclosure/rh-disclosure-lightdom.css
bennypowers, line 1: We can do all of this in shadow CSS

We only need the light DOM shim as an optional optimization to prevent CLS ie for box size

https://github.com/RedHat-UX/red-hat-design-system/wiki/CSS-Styles#lightdom-css

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  render() {
+    return html`
+      <details
+        ?open="${this.open}"
+        @keydown="${this.#onKeydown}"
+        @toggle="${this.#onToggle}">
+        <summary>
+          <rh-icon part="caret" id="caret" set="ui" icon="caret-down"></rh-icon>
+          <slot name="summary"></slot>
```
bennypowers, line 36:
```suggestion
          <slot name="summary">${this.summary}</slot>
```
we should make this an attr/slot pair like we did with [avatar](https://github.com/RedHat-UX/red-hat-design-system/blob/42aeb3c2762f9b1ac0feaff4e0f20f00e4f83344/elements/rh-avatar/rh-avatar.ts#L98), [switch](https://github.com/RedHat-UX/red-hat-design-system/blob/42aeb3c2762f9b1ac0feaff4e0f20f00e4f83344/elements/rh-switch/rh-switch.ts#L112), and [tooltip](https://github.com/RedHat-UX/red-hat-design-system/blob/42aeb3c2762f9b1ac0feaff4e0f20f00e4f83344/elements/rh-tooltip/rh-tooltip.ts#L82)

demos should prefer the attr, but we should have a dedicated demo for using the slot (i.e. slotting in an icon)

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -44,16 +56,34 @@ export class RhDisclosure extends LitElement {
 
   #onToggle(): void {
     this.open = this.detailsEl.open;
+    const event = new DisclosureToggleEvent();
+    this.dispatchEvent(event);
   }
 
   #onKeydown(event: KeyboardEvent): void {
+    const preventEscElements = `
```
bennypowers, line 64: i think we'll need to use the event composed path for this

consider

```html
<rh-disclosure>
  <super-cheeky-element>
    #shadow-root
      <sc-input>
        #shadow-root
          <input>
```


so instead of `document.activeElement` (which might be a custom element with an arbitrary tag name, and we might not even be in the document while checking, we might be in another shadow root), so something like:

```ts
if (!event.composedPath().findLast(isLegitEscElement)) {
  // close...
}
```

bennypowers (changes requested):

marionnegp (changes requested):
- [ ] Light theme focus ring color should use `--rh-color-interactive-primary-default-on-light` for active and focus states.
Screenshot 2024-12-11 at 9 59 18 AM

- [x] When the disclosure is expanded, it's showing a 24px space between the summary slot and the body content. According to our Figma file, it should be 32px (`--rh-space-2xl`). The Figma file's been updated more recently than the disclosure docs page, so apologies for the inconsistency there!
Screenshot 2024-12-11 at 10 12 21 AM

adamjohnson:
> Light theme focus ring color should use `--rh-color-interactive-primary-default-on-light` for active and focus states.

@marionnegp I used `--rh-color-interactive-primary-default` with a fallback so that it works on light and dark contexts. PTAL.

marionnegp (commented):
Focus ring changes look good!

I wanted to check the focus ring in the nested disclosures, so I wrapped that demo with `<rh-surface>`. I noticed that not everything ended up being themed correctly. (Links aren't light blue and the third nested disclosure's caret icon is black.) Is this expected based on how `<rh-surface>` works (maybe I used it wrong), or is it a bug?
Screenshot 2024-12-17 at 10 49 50 PM

marionnegp (changes requested):
Sorry one more thing! Disclosure should use the same interaction state styles as [accordion](https://ux.redhat.com/elements/accordion/demo/color-context/).

marionnegp (approved):
Thanks!

.changeset/nasty-ravens-joke.md
```diff
@@ -0,0 +1,16 @@
...
+
+A disclosure is a widget that enables content to be either collapsed (hidden) or expanded (visible).
+
+```html
+<rh-disclosure>
+  <span slot="summary">
+    Collapsed panel title
+  </span>
+  <p>Lorem ipsum dolor sit amet consectetur adipisicing, elit.</p>
+</rh-disclosure>
```
bennypowers, line 15:
```suggestion
<rh-disclosure summary="Collapsed panel title">
  <p>Lorem ipsum dolor sit amet consectetur adipisicing, elit.</p>
</rh-disclosure>
```

elements/rh-disclosure/README.md
```diff
@@ -0,0 +1,17 @@
...
+Place the following markup on your page:
+
+```html
+<rh-disclosure>
+  <span slot="summary">
+    Collapsed panel title
+  </span>
+  <p>Lorem ipsum dolor sit amet consectetur adipisicing, elit. Velit distinctio, nesciunt nobis sit.</p>
+</rh-disclosure>
+```
```
bennypowers, line 17:
~~~suggestion
<rh-disclosure summary="Collapsed panel title">
  <p>Lorem ipsum dolor sit amet consectetur adipisicing, elit. Velit distinctio, nesciunt nobis sit.</p>
</rh-disclosure>
```

### Rich summaries

When summary content should be rich HTML, use the `summary` slot instead of the `summary` attribute

```html
<rh-disclosure>
  <span slot="summary">
    <strong>Rich</strong> summary <em>content</em>
  </span>
  <p>Lorem ipsum dolor sit amet consectetur adipisicing, elit. Velit distinctio, nesciunt nobis sit.</p>
</rh-disclosure>
```
~~~

elements/rh-disclosure/rh-disclosure.css
```diff
@@ -0,0 +1,96 @@
+:host {
+  /* stylelint-disable-next-line rhds/token-values */
```
bennypowers, line 2: why do we need to disable this lint rule here? Shouldn't this use the computed value?

elements/rh-disclosure/rh-disclosure.css
```diff
@@ -0,0 +1,96 @@
...
+  &:hover,
+  &:active,
+  &:focus {
+    background-color: var(--rh-color-surface-lighter);
+  }
+
+  &:active,
+  &:focus {
+    outline: var(--rh-border-width-md, 2px) solid;
+    /* stylelint-disable-next-line rhds/token-values */
```
bennypowers, line 30: same as above

elements/rh-disclosure/rh-disclosure.css
```diff
@@ -0,0 +1,96 @@
...
+  margin-block-end: 0 !important;
+}
+
+:host:has(details[open]) {
+  box-shadow: var(--rh-box-shadow-sm, 0 2px 4px 0 rgba(21, 21, 21, 0.2));
+  position: relative;
+
+  &:before {
+    content: '';
+    /* stylelint-disable-next-line rhds/token-values */
```
bennypowers, line 83: same as above

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -0,0 +1,114 @@
...
+  render() {
+    const { on = '' } = this;
+    return html`
+      <details
+        class=${classMap({ [on]: !!on })}
+        ?open="${this.open}"
+        @keydown="${this.#onKeydown}"
+        @toggle="${this.#onToggle}">
+        <summary>
+          <rh-icon part="caret" id="caret" set="ui" icon="caret-down"></rh-icon>
```
bennypowers, line 55: what's the rationale requiring us to expose this part?

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -0,0 +1,114 @@
...
+  @query('summary') private summaryEl!: HTMLElement;
+
+  render() {
+    const { on = '' } = this;
+    return html`
+      <details
+        class=${classMap({ [on]: !!on })}
+        ?open="${this.open}"
+        @keydown="${this.#onKeydown}"
+        @toggle="${this.#onToggle}">
```
bennypowers, line 53:
```suggestion
      <details
          class=${classMap({ [on]: !!on })}
          ?open="${this.open}"
          @keydown="${this.#onKeydown}"
          @toggle="${this.#onToggle}">
```

bennypowers (changes requested):
preferred attr to slot in docs, but updated readme accordingly

elements/rh-disclosure/demo/slotted-summary.html
```diff
@@ -0,0 +1,26 @@
...
+
+<link rel="stylesheet" href="../rh-disclosure-lightdom-shim.css">
+
+<script type="module">
+  import '@rhds/elements/rh-disclosure/rh-disclosure.js';
+  import '@rhds/elements/rh-icon/rh-icon.js';
+</script>
+
+<style>
+  span[slot='summary'] {
```
bennypowers, line 17:
```suggestion
  span.icon {
```

elements/rh-disclosure/rh-disclosure-lightdom-shim.css
```diff
@@ -0,0 +1,26 @@
+rh-disclosure:not(:defined) {
+  /* stylelint-disable-next-line rhds/token-values */
```
bennypowers, line 2: why do we need to override the lint rule here? shouldn't we use the computed value?

elements/rh-disclosure/rh-disclosure.ts
adamjohnson: Resolved in [4f92e56](https://github.com/RedHat-UX/red-hat-design-system/pull/2043/commits/4f92e5607b9604396ef08bd22fa7f190ad4d40d2).

elements/rh-disclosure/rh-disclosure.css
adamjohnson: When a `<rh-disclosure>` is **not** wrapped by a `<rh-surface>`, the computed value is undefined.

This results in incorrect colors (no brand colors, no blue interactive colors—just the browser's default [transparent]).

You can see what happens if we remove the fallback colors by visiting [the demo](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/) on the DP, inspect element on the host, then remove the fallback for the `:host` border.

Screenshot of rh-disclosure variable without fallbacks, demonstrating that the border disappears sans fallbacks

So, what's the best way to tackle this?

  1. Add an `<rh-surface>` to the shadowdom every time this issue appears
  1. Keep the fallbacks, keep the `/*stylelint-disable... */`
  1. Something else...?

bennypowers: Should disclosure have a color palette (ie be a provider)? cc @coreyvickery @marionnegp

bennypowers: If, like cta, the intention is for disclosure to always only ever be a consumer, then yes, surface or another palette container is necessary (for the time being)

marionnegp: I think it should be a provider, like accordion. I can see users wanting to slot similar content into a disclosure.

adamjohnson: Resolved in [0c69b1b](https://github.com/RedHat-UX/red-hat-design-system/pull/2043/commits/0c69b1b609c16d33d2f7a315dea392273122ec57).

elements/rh-disclosure/rh-disclosure-lightdom-shim.css
adamjohnson: Resolved in [32cb9e0](https://github.com/RedHat-UX/red-hat-design-system/pull/2043/commits/32cb9e02c6140e83ad1bfa66338db391857e60f7).

elements/rh-disclosure/rh-disclosure.css
adamjohnson: Resolved in [32cb9e0](https://github.com/RedHat-UX/red-hat-design-system/pull/2043/commits/32cb9e02c6140e83ad1bfa66338db391857e60f7).

elements/rh-disclosure/rh-disclosure.css
adamjohnson: Resolved in [32cb9e0](https://github.com/RedHat-UX/red-hat-design-system/pull/2043/commits/32cb9e02c6140e83ad1bfa66338db391857e60f7).

elements/rh-disclosure/rh-disclosure-lightdom-shim.css
```diff
@@ -1,6 +1,7 @@
 rh-disclosure:not(:defined) {
-  /* stylelint-disable-next-line rhds/token-values */
-  border: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle, #c7c7c7);
+  --rh-color-border-subtle: var(--rh-color-gray-30, #c7c7c7);
```
bennypowers, line 2: this will override the context values for disclosures in noscript scenarios

elements/rh-disclosure/rh-disclosure.css
```diff
@@ -1,11 +1,18 @@
 :host {
-  /* stylelint-disable-next-line rhds/token-values */
-  border: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle, #c7c7c7);
+  --rh-color-border-subtle: var(--rh-color-gray-30, #c7c7c7);
```
bennypowers, line 2: this will override the context values for disclosures in noscript scenarios

elements/rh-disclosure/rh-disclosure.css
```diff
@@ -41,11 +47,11 @@ summary {
   }
 }
 
-.dark summary {
+.on.dark summary {
   &:hover,
   &:active,
   &:focus {
-    background-color: var(--rh-color-surface-dark-alt);
+    background-color: var(--rh-color-surface-dark-alt, #292929);
```
bennypowers, line 54: Because the `-alt` tokens fit awkwardly into the theming system, I'd like to see if we can move away from `*-dark-alt` and use relative colour in oklch instead. This is a larger issue than disclosure though, so perhaps we should spin this comment off into Diglett.

adamjohnson: Let's spin this off into DIglett.

elements/rh-disclosure/rh-disclosure-lightdom-shim.css
adamjohnson: Seems like we have two options here:

1) Remove these variable definitions (eg line 2 here and lines 2 & 10 in `rh-disclosure.css`) and tell people to always insert `<rh-disclosure>` inside a provider/`<rh-surface>`.
2) Leave these variable definitions so it "just works" in it's base form and, if/when these colors are incorrect, lean on the implementing dev to correct the border colors.

What's your preference? 

(Let me know if there's an option three I'm not seeing.)

bennypowers: @zeroedin's work on light-dark is the place to solve this problem. let's remove the override here, so that it's correct as intended on surface.

elements/rh-disclosure/test/rh-disclosure.spec.ts
bennypowers, line 1: we need some basic functionality tests here for open/close

bennypowers (changes requested):
I made some changes here:
- removed colour overrides
- simplified the esc trap (still not entirely sure why we need this)
- removed `:host:has`, which i think was anyways leftover from the first iteration of this element


Please check over my changes for correctness, and add docs and tests

elements/rh-disclosure/test/rh-disclosure.spec.ts
adamjohnson: Added in [bd11e2b](https://github.com/RedHat-UX/red-hat-design-system/pull/2043/commits/bd11e2b6a8a3621e2c12879cdefbb705669bbcaa).

elements/rh-disclosure/rh-disclosure-lightdom-shim.css
adamjohnson: Going with number 1, ✔️👍

marionnegp:
@coreyvickery, do you have capacity to do docs for `<rh-dialog>`? Looks like we'll have to update [what we have](https://ux.redhat.com/patterns/disclosure/) and move it out of the patterns section.

bennypowers (changes requested):
Still needs docs before we can merge

marionnegp:
@coreyvickery, do you have capacity to do docs for `<rh-disclosure>`? Looks like we'll have to update [what we have](https://ux.redhat.com/patterns/disclosure/) and move it out of the patterns section. If not, I can work on it.

Sorry if you're getting pinged twice on this. I had mistakenly written "dialog" instead of "disclosure."

coreyvickery:
@marionnegp I can, but I was under the impression that there was much more to do for `<rh-disclosure>` before the docs.

cc @bennypowers @markcaron

coreyvickery:
@adamjohnson It sounds like we're deprecating `--rh-color-surface-dark-alt` at some point, so for the dark scheme hover/focus/active states, you can use the value `--rh-color-surface-dark`. I think `--rh-color-surface-darker` is a bit too subtle.

coreyvickery:
Update 3/13: Docs are in progress.

adamjohnson:
Docs have been added!

@coreyvickery PTAL / double check the [docs in the DP](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/) to ensure that everything looks as it should.

Also, let me know if you want a different thumbnail image for disclosure on Elements > All Elements. TIA!

coreyvickery:
@adamjohnson Some quick content edits from Greg.

### Style page
- [ ] Under the `Hover` heading, change the text to `The entire trigger button surface changes color on hover.`
- [ ] Under the `Hover` heading, a table is missing under the images
- [ ] Under the `Focus and Active` heading, change the text to `A focus ring wraps around the entire trigger button in both focus and active states.`

### Accessibility page
- [ ] Under the `Keyboard interactions` heading, change the text to `The disclosure trigger button and content within the panel can be navigated with a keyboard.`
- [ ] In the table, the `Tab` and `Shift+Tab` rows have updated content
- [ ] Under the `Touch targets` heading, change the text to `The large disclosure trigger buttons make them easy to select.`

coreyvickery:
@adamjohnson Docs look great!

adamjohnson:
@bennypowers This PR now has the `color-palette` / `light-dark()` changes integrated. It also has docs. PTAL.

Next week if you'd give it the ol 👍, we can get it merged into Cubone and released shortly thereafter.

elements/rh-disclosure/docs/10-style.md
```diff
@@ -0,0 +1,120 @@
...
+    </ol>
+  </figcaption>
+</figure>
+
+## Color scheme
+
+### Light scheme
+
+<uxdot-example>
+  <img alt="Two disclosures. One closed, one open on a light color scheme."
```
bennypowers, line 30: Down the line we should inline all of `<rh-disclosure>`'s docs SVGs with `light-dark()` and fallback vars (and unique ids)

bennypowers (approved):

PR #2054, opened 2024-11-18
docs: dark mode

bennypowers:
## What I did

1. add dark mode toggle to masthead (iconography, ux TBD)
2. various hacks and workarounds in context, surface, switch, to make darkmode + hydration work
3. prevent fouc (flash of white background)

https://github.com/user-attachments/assets/f607a492-bf41-4931-96d5-3cfb2b11ce2a

## Testing Instructions

1. set throttling in network panel
2. set dark mode prefs in browser
3. load up the dp
4. toggle the switch in the masthead
5. toggle dark mode in OS / browser prefs

## Notes to Reviewers
icons and ux of the switch are TBD


bennypowers:
Theme and color palette are [two different things](https://ux.redhat.com/theming/color-palettes/). In RHDS, "theme" means [*customizing*](https://ux.redhat.com/theming/customizing) one or more color palettes. When we think about it that way, what need does the design system docs site have to customize the theme, outside of pages which provide examples of how to do so? Wouldn't that be confusing, a bait-and-switch?

zeroedin:
> ...let me describe what I think you mean, and respond to that. If I got it wrong, let me know.

I could have been clearer.

By theme switch I mean utilizing our theme approach alongside of [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) which is IMHO the basis for a switch of light and dark directed by user preference for the whole page.  This then is an additional layer on top of what we have already with custom themes + color-palettes.

We are maybe saying the same thing here or closer than we think.  To help demonstrate here is a fast and dirty code example https://codepen.io/zeroedin/pen/NPKGmWp of how I would expect this to work.

`prefers-color-scheme` is a top level page thing from my point of view, then surface/[color palette] is the next layer and that then is customized through the theming.  This isn't to say the switch can target specific surfaces/color-palettes but that it isn't it's primary focus.

bennypowers:
This PR already supports prefers-color-scheme (testing instructions point 5)

PR #2057, opened 2024-11-20
Feat/lightdom css controller

bennypowers:
## What I did

1. add new `LightdomCSSController` which auto-loads lightdom css
2. remove rh-table-lightdom.css from docs pages
3. add an `--_rhds-lightdom: rh-table` brand to the table lightdom stylesheet (required for the controller to work efficiently)

## Testing Instructions

1.

## Notes to Reviewers
The SSR story has yet TBD, it may be possible to automatically write lightdom sheets as styletags via the controller / element renderer

### DIFF
--- a/docs/_includes/layouts/pages/element.11ty.ts
+++ b/docs/_includes/layouts/pages/element.11ty.ts
@@ -71,9 +71,9 @@ export default class ElementsPage extends Renderer<Context> {
     } = doc;
     const content = fileExists ? await this.renderFile(filePath, ctx) : '';
     const stylesheets = [
-      '/assets/packages/@rhds/elements/elements/rh-table/rh-table-lightdom.css',
+      // '/assets/packages/@rhds/elements/elements/rh-table/rh-table-lightdom.css',
       '/styles/samp.css',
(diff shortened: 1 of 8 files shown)


zeroedin:
Doing this causes FOUC on rh-table at load.  I don't see this as a desired effect.

https://github.com/user-attachments/assets/fc182c1d-b5f7-4ee1-bab5-39c0e4e0d4aa

I'm questioning how loading the styles via javascript in this case would be helpful to the end user seems likely it would only reduce the need for an implementer to add a`<link href>` for for mandatory lightdom css components.  

It may also be I need more time to process how we divide up the lightdom.css and lighdom-shim.css files and which get loaded via this controller.

bennypowers:
yes, as mentioned in chat, I'm aware of the fouc, and as i wrote in the description i'm working on an ssr solution

lightdom css is part of the element definition. meaning that the element *does not work* if you don't have it's associated lightdom css. What this PR solves (as discussed at length in many places over several months) is 

1. the case where the user forgot to load the lightdom css - one of our most common support issues
2. the case where the user *did* load the lightdom css, but loaded it in the wrong place (i.e. in the document when the element is in a shadow root)

Hope that clarifies

PR #2060, opened 2024-11-20
[POC] feat(navigation-secondary): poc for details summary

zeroedin:
## What this POC is

- Adds `details > summary` as the key component for the hamburger menu of `navigation-secondary`
- Enables `navigation-secondary` to be used as a DSD fragment (or SSR'd) that can be hydrated on the client.
- Removes CLS from load as the initial load of the hamburger is closed due to usage of the `details > summary` standard mechanism.
- Enables navigation to function as a collapsed menu even under no JS scenarios, either through failure of proper load or user preference (not a condition to support but still works for).
- Shifts from `@media` queries to `@container` queries.  Likely improvements can be made to the implementation of a resize observer in lieu of using Screensize controller which only works with `matchMedia`
- A proving ground for ideas on how to better support `navigation-primary` in the future as a mission critical component.

##  What this POC is not

- This is a breaking change.  Not an incremental update to `navigation-secondary`.  
- Does not solve for every possible use/state, such as when used as standard component implementation and not the DSD artifact, if JS fails the details summary is not exposed therefore content ends up hidden for dropdowns.  However such use case is deemed a choice by implementation for this POC.  There maybe some workarounds we can do like a fallback slot for a link in case of NO JS scenarios, this hasn't been fully explored.
- Does not solve for the distribution of the DSD artifact.

### DIFF
--- a/elements/rh-navigation-secondary-dsd/README.md
+++ b/elements/rh-navigation-secondary-dsd/README.md
@@ -1,0 +1,11 @@
+# Navigation Secondary Dsd
+Add a description of the component here.
+
+## Usage
+Describe how best to use this web component along with best practices.
+
+```html
+<rh-navigation-secondary-dsd>
+
+</rh-navigation-secondary-dsd>
+```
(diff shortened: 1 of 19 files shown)


zeroedin:
Given the launch of #2171 right around the corner, which takes some of the concepts from this POC, I think we can close this PR now.

PR #2067, opened 2024-11-26
fix(footer): show content when js fails

### LINKED ISSUE #2058: [bug] `<rh-footer>` doesn't gracefully degrade
### Describe the bug

There are some `-lightdom.css` styles that hide content when the element is not defined. This creates a broken experience when JS doesn't load.

IMO, the Footer is an important piece of navigation and access to information. Therefore, the content within the light DOM should be styled in a progressively enhanced way—like a stacked list of links at the minimum.

Maybe there's a technical challenge I'm missing?

### Screenshots

image

image

### Steps to reproduce

1. Go to redhat.com or [rh-footer demo](https://ux.redhat.com/elements/footer/demo/)
2. Disabled JS
3. Refresh and see broken footer (hidden links and content)

### Expected behaviour

Content/links aren't hidden.

### Element HTML

_No response_

### Operating System (OS)

iOS

### Browser

Chrome

_No response_

bennypowers:
Closes #2058

## What I did

1. show footer content after 5 seconds if javascript fails
2. remove visually-hidden clip styles

## Testing Instructions

1. run dev server
1. disable js
2. view footer demo
3. after 5 seconds, footer content should appear


markcaron:
@bennypowers The functionality seems to work great! I wonder, though, if there's some way we can add a little more style to the lightDOM, so the headings and other things are legible. Not sure if there's any other `:not(:defined)` styles we can use for layout as well.

PR #2068, opened 2024-11-26
fix(accordion): hover colors

### LINKED ISSUE #2065: [bug] `<rh-accordion>` not showing hover state
### Describe the bug

`<rh-accordion>`'s hover state isn't showing a hover state.

### Screenshots

_No response_

### Steps to reproduce

View accordion demos on UX dot.

### Expected behaviour

On hover, an accordion background used to turn to `--rh-color-surface-lightest` (light theme) or `--rh-color-surface-dark-alt` (dark theme). ([Docs examples](https://ux.redhat.com/elements/accordion/style/#hover))

Should we consider not changing the background color but showing the red border on the left on hover?

### Element HTML

_No response_

### Operating System (OS)

macOS

### Browser

Chrome

_No response_

bennypowers:
Closes #2065  

## What I did

1. restore hover colors for accordion headings
2. remove unnecessary wrapping div

## Testing Instructions

1. view accordion demo in dev server on light and dark backgrounds


marionnegp (changes requested):
- [x] I can see the hover background color in the color context demo, but it's not showing up in any other demo.

- [ ] Did the div that was removed allow the red line to extend to the panel header when it's expanded?
Screenshot 2024-11-26 at 9 23 53 AM

- [ ] On hover, the accordion in dark theme shows black text.
Screenshot 2024-11-26 at 9 28 36 AM

bennypowers:
looks like maybe some ssr problems

bennypowers:
@marionnegp PTAL, thanks

marionnegp (changes requested):
The changes to previous comments look good! I noticed a couple other things that I might have missed last time.

- [ ] When any of the accordion panels are expanded, the last accordion also shows a dropshadow. There shouldn't be a dropshadow on a panel unless it's expanded.
Screenshot 2024-12-02 at 4 24 07 PM

- [ ] When you first land on the color context demo, the text in an expanded panel is black for the dark theme accordions. I'm able to resolve it by viewing the accordion in a light color palette and then clicking back to a dark color palette, so it might just be a demo error. Noting it in case this confuses UX dot users.
Screenshot 2024-12-02 at 4 25 05 PM

bennypowers:
let's do the drop shadow thing in a separate issue - there was confusion about this earlier in the quarter and i want to make sure we're all on the same page

with regards to the context/ssr issues, i added another workaround, so please take a look

adamjohnson (changes requested):
According to [the docs](https://ux.redhat.com/elements/accordion/style/#hover), the focus states still need some love:

- [x] Background color needs to be changed / applied on focus
- [x] Ensure blue outline on focus across browsers

Focus state in Microsoft Edge:
Accordion with a focused item in Edge showing a black outline

Focus state in Safari:
Example of accordion focus in safari, noting the second item is focused and it's confusing

Note that in Safari, users could easily confuse which item is currently focused.

marionnegp (approved):

zeroedin (changes requested):
The font weight  of the accordion panel headers looks different in Safari and Firefox then it does in Chrome.

Chrome:
image

Firefox:
image

Safari:
image

In fact manually changing Chrome's font weight to 500 and 700 does no visual change.

https://github.com/RedHat-UX/red-hat-design-system/blob/8ec2cc48ce5268cc9f7ae40af6eca3cc806b3252/elements/rh-accordion/rh-accordion-header.css#L128-L130

I believe this has to do with the fonts stack we are loading and Chrome's difference in how it renders weights that aren't loaded for the stack.  Where as I believe Safari and seemingly Firefox too will synthesize it if it is missing.. at least I believe I remember seeing that somewhere was the case for Safari although a quick search didn't return proof to that respect.

https://github.com/RedHat-UX/red-hat-design-system/blob/8ec2cc48ce5268cc9f7ae40af6eca3cc806b3252/docs/styles/fonts.css#L25-L39

We do not load a 700 weight for RedHatText in our [fonts.css](https://github.com/RedHat-UX/red-hat-design-system/blob/8ec2cc48ce5268cc9f7ae40af6eca3cc806b3252/docs/styles/fonts.css) file.  We do however load a 700 weight for RedHatDisplay.  

Are we sure we want to be using RedHatText for this header? 

https://github.com/RedHat-UX/red-hat-design-system/blob/8ec2cc48ce5268cc9f7ae40af6eca3cc806b3252/docs/styles/fonts.css#L17-L23

If so, we'll need to modify the font stack to accommodate the 700 weight for body text.  Also that said [tokens for body text](https://github.com/RedHat-UX/red-hat-design-system/blob/8ec2cc48ce5268cc9f7ae40af6eca3cc806b3252/docs/styles/fonts.css) only go to 500 so likely this isn't the font family we want here.

bennypowers:
yes, @zeroedin : in my experiments, I found that no amount of `host.requestUpdate('on')` was helpful to dislodge the (incorrect) ssr value, but unsetting then resetting the property did trigger a rerender of that template part. I also found that I had to adjust some of the timing of the provider lifecycle so that monkey-in-the-middle late upgrading providers still work as expected both in the non-ssr (dev server) and ssr (eleventy) scenarios.

i'm looking forward to lit-ssr supporting this stuff so we can jettison our custom code

zeroedin (approved):

adamjohnson (approved):
Links glow, tint modified

PR #2069, opened 2024-11-26
fix(tooltip): announce content to assistive technology

### LINKED ISSUE #2051: [bug] `rh-tooltip` should announce content to screen readers on focus
`rh-tooltip` currently only presents its contents visually upon hover and focus.

As demonstrated at [its docs demo page](https://ux.redhat.com/elements/tooltip/demo/), tooltips are keyboard-focusable, by default. And they do display their text content visually upon focus. However, this content is not currently read by screen readers.

Screencap of a focused tooltip

adamjohnson:
## What I did

1. Tooltip content is put in a `.visually-hidden` `<div>` with `aria-live` in the `<body>` so that it can be read by AT.
1. Added functionality where hitting ESC closes the `rh-tooltip`.
1. ~~Added an `aria-describedby`/`id` pair to elements in the lightdom of `<rh-tooltip>`~~
1. ~~Added Markup Guidance to the [Accessibility page](https://deploy-preview-2069--red-hat-design-system.netlify.app/elements/tooltip/accessibility) of `<rh-tooltip>` docs.~~

## Testing Instructions

1. Check out the [tooltip demo](https://deploy-preview-2069--red-hat-design-system.netlify.app/elements/tooltip/demo/) in the DP.
1. Test tooltip with various combinations of assistive tech
1. ~~Check out the Accessibility page, proofread the Markup Guidance, give feedback.~~

## Notes to Reviewers

~~I couldn't get `customElements.whenDefined()` working in place of `setTimeout()`. PTAL and maybe you will have better luck.~~

~~This PR changes the trigger `<rh-button>` elements in the demos to native `<button>` elements so that tooltip can be accessible.~~

Closes #2051.


elements/rh-tooltip/rh-tooltip.ts
```diff
@@ -100,6 +104,75 @@ export class RhTooltip extends LitElement {
...
+
+    const findFocusableElement = (node: Element): HTMLElement | null => {
+      if (node.matches(focusableSelector)) {
+        return node as HTMLElement;
+      }
+
+      if (node.tagName.toLowerCase().startsWith('rh-')) {
+        setTimeout(() => {
+          const shadowFocusable = node.shadowRoot?.querySelector(focusableSelector);
+          shadowFocusable?.setAttribute('aria-describedby', this.#tooltipContentID);
```
bennypowers, line 139: we shouldn't need to do this. 

When I tried the deploy preview, I found that:

1. Orca linux didn't read the tooltip content in any browser
2. VO Safari didn't read the tooltip content
3. VO Chrome read the tooltip content, but not as a part of the tooltip, i.e. in a separate announcement


I then added `aria-describedby` to the shadow span wrapping the invoker slot:
```html
<div id="container"
     style="${styleMap(styles)}"
     class="${classMap({ open,
                        'initialized': !!this.#initialized,
                         [on]: !!on,
                         [anchor]: !!anchor,
                         [alignment]: !!alignment })}">
  <div id="invoker"
       class="c"
       role="tooltip"
       aria-describedby="tooltip"
       aria-labelledby="tooltip">
    <slot id="invoker"></slot>
  </div>
  <div id="tooltip"
       class="c"
       aria-hidden="${String(!open) as 'true' | 'false'}">
    <slot name="content">${this.content}</slot>
  </div>
</div>
```



I found that:
1. orca on linux didn't read the tooltip
4. VO safari didn't read the tooltip
5. VO chrome read the tooltip

So in other words, using lightdom aria attributes didn't appear to me to have much positive effect, and using the shadow DOM attributes seems to work a bit better

bennypowers (changes requested):

elements/rh-tooltip/rh-tooltip.ts
adamjohnson: Now I realize my direction thus far was futile since reading `aria-describedby` / IDREF pairs from across the shadowdom / lightdom will not work without cross-root ARIA.

It's also clear that adding these ARIA + IDREF attributes in the shadowdom does little to nothing to aid accessibility.

So, to prioritize accessibility, I have changed the demos and added some docs around `<rh-tooltip>` accessibility.

elements/rh-tooltip/rh-tooltip.ts
```diff
@@ -57,10 +58,13 @@ export class RhTooltip extends LitElement {
 
   #initialized = false;
 
+  #tooltipContentID = getRandomId('tooltip-content');
+
   override connectedCallback(): void {
     super.connectedCallback();
     ENTER_EVENTS.forEach(evt => this.addEventListener(evt, this.show));
     EXIT_EVENTS.forEach(evt => this.addEventListener(evt, this.hide));
+    document.addEventListener('keydown', this.#onKeydown);
```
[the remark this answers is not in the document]
adamjohnson: Resolved in [3e12c51](https://github.com/RedHat-UX/red-hat-design-system/pull/2069/commits/3e12c51c87d423fb9e2fca47a1eb23998f1a2af5). Great suggestion, thanks!

adamjohnson:
Looks like this `aria-live` approach does work in Firefox with VO, but it does _not_ work in Safari or Edge. I added `aria-atomic="true"` hoping that would fix it. Alas, no joy.

When adding `aria-live` content to the lightdom, it gets announced. WDYT about adding the `<div aria-live="polite">` there?

bennypowers:
i'll try refactoring to using a single, global, visually hidden aria-live region appended to the document body

bennypowers:
orca ff works
orca epiphany doesnt' announce, i think live regions don't work at all in that configuration: https://codepen.io/bennyp/pen/MYgJVdR?editors=1001
orca doesn't appear to work with chromium at all, not just for this pr

zeroedin:
This comment/issue is not related to changes in this PR, but it looks like there is an issue with positioning calculation in the left [demo](https://deploy-preview-2069--red-hat-design-system.netlify.app/elements/tooltip/demo/left/) where the tooltip is both in a `dir="rtl"` and `style="text-align: right"` styled parent. My assumption is this is related to RTL support of floating-dom controller.  I will break this out into a new issue and track it separately.

elements/rh-tooltip/rh-tooltip.ts
```diff
@@ -43,6 +43,41 @@ export class RhTooltip extends LitElement {
...
+  }
+
+  private static initAnnouncer() {
+    document.body.append((this.announcer = Object.assign(document.createElement('div'), {
+      ariaLive: 'polite',
+      // apply `.visually-hidden` styles
+      style: /* css */`
+        position: fixed;
+        top: 0;
+        left: 0;
```
zeroedin, line 73:
```Suggestion
inset-block-start: 0;
inset-inline-start: 0;
```

adamjohnson:
Feedback from Greg says that this (mostly) works across various screen readers—which is much better than it not working at all.

> So all things considered I think we can approve this as best effort

I agree with this. PTAL and thanks everyone for your help and feedback.

hellogreg:
> Feedback from Greg says that this (mostly) works across various screen readers—which is much better than it not working at all.

Yep, we're at least getting _something_ from all the major browser/reader combos, even if there are some quirks with, say, Android TalkBack/Chrome and Windows Narrator/Edge.

bennypowers (approved):

zeroedin (approved):

PR #2077, opened 2024-12-05
docs(tokens): renaming crayon tokens to primitive tokens

markcaron:
Trying to get rid of the term "crayon" since it no longer makes sense (we don't use crayon names we use number values) and it's confusing when switching between "crayon" for color and "global" for everything.

## What I did

1. Renamed "Crayon tokens" and other uses of "crayon" to "primitive" on the Foundations/Color, Tokens/Overview, and Theming/Customizing pages

## Testing Instructions

1. Check the Foundations/Color page and make sure it makes sense along with...
2. The Tokens/Overview page and...
3. The Theming/Customizing

### DIFF
--- a/docs/foundations/color/index.md
+++ b/docs/foundations/color/index.md
@@ -77,12 +77,12 @@ Design tokens are how we communicate and translate our design decisions to code.
 
 To learn more, go to the [Tokens][tokens] section.
 
-### Crayon and semantic tokens
+### Primitive and semantic tokens
 
 Our design system includes two sets of tokens:
 
-- **Crayon tokens** - reference hard-coded values and offer no information about usage
-- **Semantic tokens** - reference crayon colors and define how a color should be used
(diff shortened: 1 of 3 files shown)


markcaron:
Waiting on an updated diagram/image on the Foundations' Color page, which says "Color token".
image

CC @marionnegp

markcaron:
@marionnegp we also need to update this graphic on the [Tokens/Overview page](https://deploy-preview-2077--red-hat-design-system.netlify.app/tokens/#introduction):
image

It references `brand-red-500` which doesn't exist. The other tokens should also reflect what's actually in the system.

marionnegp:
**For Foundations - Color page:**
overview-color-semantic-tokens

**For Tokens - Overview page:**

Introduction

design-tokens-intro

Why we need tokens

design-tokens-why-we-need-tokens

Naming tokens

design-tokens-naming

Aliases

design-tokens-aliases

PR #2078, opened 2024-12-06
feat(dialog): use native HTML dialog element

adamjohnson:
## What I did

1. This is a continuation of #1865. Opening this as a separate PR because of merge conflicts in 1865.
2. Please review the PR body and comments thoroughly from 1865 for changes
    * Highlights:
        * Now uses the native HTML `<dialog>` element.
        * Added the `accessible-label` property
        * Improved modal styles
        * Focus trapping
        * Updated a11y docs
        * rh-surface wrapping only `<dialog>` element

## Testing Instructions

1. Visit [the DP Demo](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/).
2. View the other demos:
    * [Color Context](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/color-context/)
    * [Events](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/events/)
    * [Lots of Content](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/lots-of-content)
    * [No headings](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/no-headings/)
    * [No header content](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/no-header-content/)
    * [Video](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/video-modal/)
3. Add various elements to modal header, body, footer
4. Test variants (`variant="small"` etc and `position="top"`)
5. Check a11y
6. Check for appropriate spacing

### Docs updates:

  - [x] Some docs on the Guidelines page are incorrect as they have the modal body [scrolling independently](http://localhost:8080/elements/dialog/style/#configuration) from the header and footer. This is not possible short of a variant.
  - [x] Guidelines for the [video player close button location](http://localhost:8080/elements/dialog/guidelines/#video-player-dialog) differ from actual implementation (update image)
  - [x] Guidelines for [Overflow content](http://localhost:8080/elements/dialog/guidelines/#overflow-content) differ from implementation (update text, image)
  - [x] Border radius of dialog in images

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/1242 and https://github.com/RedHat-UX/red-hat-design-system/issues/1238.


marionnegp (changes requested):
- The close button's target looks pretty large. Can we use `--rh-length-xl` for the width and height of the touch target?
Screenshot 2024-12-09 at 12 35 46 PM

- Can the slots use [this spacing](https://ux.redhat.com/elements/dialog/style/#space)?

- Is there a way to keep the close button fixed if the content has to scroll?
Screenshot 2024-12-09 at 3 04 59 PM

- This comment might need to go back to design for a future feature update, but the close button isn't visible in the video dialog demo. (I can make an issue if design needs to tackle dark theme to fix this.)
Screenshot 2024-12-09 at 3 05 40 PM

elements/rh-dialog/demo/no-headings.html
```diff
@@ -1,11 +1,12 @@
...
+<rh-dialog accessible-label="This dialog's accessible name" trigger="no-headings-trigger">
+  <p>This dialog uses the <code>accessible-label</code> attribute to define its <code>aria-label</code>.
+     If the dialog includes a heading, the heading automatically becomes its accessible name—as long as the
+     <code>accessible-label</code> attribute does not exist. If neither a heading nor the <code>accessible-label</code>
+     attribute are provided, the dialog's accessible name defaults to the text content of its trigger element.
+     Learn more about these attributes in the <a href="https://ux.redhat.com/elements/dialog/code/">rh-dialog docs</a>.
+  </p>
+  <rh-cta slot="footer">
     <a href="#">Call-to-action</a>
   </rh-cta>
```
bennypowers, line 11:
```suggestion
  <rh-cta slot="footer" href="#">Call-to-action</rh-cta>
```

elements/rh-dialog/demo/rh-dialog.html
```diff
@@ -6,7 +6,7 @@ <h2 slot="header">Modal dialog with a header</h2>
     aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
     fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit
     anim id est laborum.</p>
-  <rh-cta>
+  <rh-cta slot="footer">
     <a href="#bar">Learn more</a>
   </rh-cta>
```
bennypowers, line 11:
```suggestion
  <rh-cta slot="footer" href="#bar">Learn more</rh-cta>
```

elements/rh-dialog/docs/40-accessibility.md
```diff
@@ -63,8 +63,24 @@ Only the close button and any interactive elements are selectable.
...
+
+Users can optionally provide an `accessible-label` attribute which overrides the built-in `aria-labelledby` functionality:
+
+```html
+<rh-dialog accessible-label="Page Properties" trigger="first-modal-trigger">
+  ...
+</rh-dialog>
+```
+
+The `accessible-label` attribute assigns an `aria-label` to the `<dialog>` element.
```
bennypowers, line 81:
```suggestion
```

i don't think this information (the "how") is relevant to consumers. Rather, the relevant part is that applying `accessible-label` to rh-dialog makes the dialog more accessible to sr users.

same for the comment about `aria-labelledby` above. 

we are in the business of making accessible html elements with custom, public apis. the exact, granular details of what happens in the shadow root and in memory (ElementInternals) should not be of any concern to consumers of the design system.

elements/rh-dialog/docs/40-accessibility.md
```diff
@@ -63,8 +63,24 @@ Only the close button and any interactive elements are selectable.
 
 ### Backdrop
 
-A dialog will not close by users clicking or tapping the backdrop or outside of the container.
+A dialog will close by users clicking or tapping the backdrop or outside of the container.
+
+
+## Accessible labels
+
+Each dialog needs an accessible name. If a dialog has a heading tag in the `header` or default slot, this component will automatically apply an appropriate ID to the heading tag and an `aria-labelledby` attribute to the `<dialog>`.
```
bennypowers, line 71:
```suggestion
Each dialog needs an accessible name. If a dialog has a heading tag in the `header` or default slot, it will automatically be used as the accessible name of the dialog.
```

elements/rh-dialog/docs/40-accessibility.md
```diff
@@ -63,8 +63,24 @@ Only the close button and any interactive elements are selectable.
...
+
+Users can optionally provide an `accessible-label` attribute which overrides the built-in `aria-labelledby` functionality:
```
bennypowers, line 73:
```suggestion
If there is no slotted heading, users should provide an `accessible-label` attribute, the value of which will be used as the accessible name:
```

elements/rh-dialog/docs/40-accessibility.md
```diff
@@ -63,8 +63,24 @@ Only the close button and any interactive elements are selectable.
...
+
+```html
+<rh-dialog accessible-label="Page Properties" trigger="first-modal-trigger">
+  ...
+</rh-dialog>
+```
+
+The `accessible-label` attribute assigns an `aria-label` to the `<dialog>` element.
 
+If neither an `accessible-label` nor any headings exist, the `aria-label` on the `<dialog>` will default to the text of the dialog's trigger.
```
bennypowers, line 83:
```suggestion
If neither an `accessible-label` nor any headings exist, the accessible name of the dialog will fall back to the text of the dialog's trigger.
```

elements/rh-dialog/docs/40-accessibility.md
```diff
@@ -37,17 +37,17 @@ A dialog can be opened by pressing `Enter` when the dialog trigger has focus. Wh
 
 ## Focus order
 
+When a dialog opens, the close dialog button recieves focus by default. Users have the ability to tab through each focusable element in the dialog. When the last focusable element in the dialog is reached, focus returns to the close dialog button—effectively trapping focus inside the dialog.
```
bennypowers, line 40:
```suggestion
When a dialog opens, the close dialog button receives focus by default. Users have the ability to tab through each focusable element in the dialog. When the last focusable element in the dialog is reached, focus returns to the close dialog button—effectively trapping focus inside the dialog.
```

elements/rh-dialog/rh-dialog.css
```diff
@@ -74,121 +72,133 @@ section {
...
-  aspect-ratio: 1/1;
+[part='close-button']::part(button) {
+  padding: var(--rh-space-xl, 24px);
 }
 
-[part='close-button']:is(:hover, :focus-within, :focus-visible) svg:is(svg, :hover) {
-  fill: var(--rh-color-icon-secondary-on-light, #151515);
+@media (min-width: 1200px) {
+  [part='close-button']::part(button) {
+    padding: var(--rh-space-2xl, 32px);
```
bennypowers, line 129: should we use a margin here on the rh-button instead of padding the shadow part?

elements/rh-dialog/rh-dialog.css
```diff
@@ -74,121 +72,133 @@ section {
...
 :host([type='video']) [part='close-button'] {
-  top: var(--offset-top);
-  right: var(--offset-right);
-  padding: var(--rh-space-sm, 6px);
-  color: var(--rh-color-icon-secondary-on-dark, #ffffff);
+  inset-inline-end: var(--offset-right);
+  inset-block-start: var(--offset-top);
+}
+
+:host([type='video']) #close-button::part(button) {
```
bennypowers, line 159: maybe we should use shadow classes for this instead of attr selectors? wdyt?

elements/rh-dialog/rh-dialog.ts
bennypowers, line 1: changeset notes that `backdrop` part is deprecated, so please update jsdoc here

elements/rh-dialog/rh-dialog.ts
```diff
@@ -80,93 +83,100 @@ export class RhDialog extends LitElement {
    */
   @property({ reflect: true }) position?: 'top';
 
+  /**
+   * Use `accessible-label="My custom label"` to add an `aria-label` to the `<dialog>` element.
```
bennypowers, line 87:
```suggestion
   * Use `accessible-label="My custom label"` to specify the dialog's accessible name.
```

elements/rh-dialog/rh-dialog.ts
```diff
@@ -235,35 +239,126 @@ export class RhDialog extends LitElement {
...
-      if (closeOnOutsideClick && path.includes(overlay!) && !path.includes(dialog!)) {
+      if (closeOnOutsideClick && !path.includes(content!)) {
         event.preventDefault();
         this.cancel();
       }
     }
   }
 
-  @bound private onKeydown(event: KeyboardEvent) {
+  #trapFocus() {
```
bennypowers, line 262: why can't we do this with `.showModal()`?

adamjohnson: Thanks for that feedback! Indeed, those links you referenced are helpful and differ from what the APG says.

I was actually looking at the APG Guidelines for [Dialog (modal)](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/):

> Tab: If focus is on the last tabbable element inside the dialog, moves focus to the first tabbable element inside the dialog.
Shift + Tab: If focus is on the first tabbable element inside the dialog, moves focus to the last tabbable element inside the dialog.

That said, @hellogreg pointed me to a [massively helpful discussion](https://github.com/whatwg/html/issues/8339#issuecomment-1263476659) on this very topic.

tl;dr the way the browser has implemented "focus trapping" with a native HTML `<dialog>` element is indeed correct. The APG guidelines are intended for people not using a native `<dialog>` element.

So, after making the changes Marionne requested, I will update/remove this `#trapFocus()` code and resolve this conversation. 👏

.changeset/loud-coins-shake.md
```diff
@@ -0,0 +1,21 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-dialog>`: Dialog now uses the native HTML `<dialog>` element internally.
+
+Note: `#overlay` is now deprecated in favor of a public variable on the CSS psuedo-element `::backdrop`:
```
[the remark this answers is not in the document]
adamjohnson: The overlay `part` / `div` is still there and can be used. I took care to make sure this wasn't a major, breaking change. If users defined something on the `#overlay`, that will take precedence.

marionnegp (changes requested):
The changes form the last PR look great! 

- [ ] In the "Lots of Content" demo, the padding around the content is lost as you scroll, and it looks like the header sticks to the top, instead of scrolling with the rest of the content area.

Screenshot 2024-12-16 at 10 28 15 AM

adamjohnson:
Thanks for your feedback!

> * [ ]  In the "Lots of Content" demo, the padding around the content is lost as you scroll, and it looks like the header sticks to the top, instead of scrolling with the rest of the content area.

@marionnegp In the existing docs under Style > [Configuration](http://localhost:8080/elements/dialog/style/#configuration), the image denotes that the "header section" remains fixed. Some things to note:

  * The current implementation on uxdot has a [fixed header](https://ux.redhat.com/elements/dialog/demo/lots-of-content/)
  * As I noted in the PR body, having both a fixed header _and_ footer is not possible short of a variant + setting the height of the dialog to ~100% of the browser's height

Going forward, which option should we choose:

  1. Keep the fixed header, make sure header content retains padding
  1. Nix the fixed header, let header scroll out of view. The only sticky item is the close button.

elements/rh-dialog/docs/40-accessibility.md
adamjohnson: Resolved in [303f85f](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/303f85f107a053833d07585d07cff8bdd74f2cff).

elements/rh-dialog/rh-dialog.css
adamjohnson: Resolved in [c292dbe](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/c292dbeb9ee26c02f87c0f4830c2b17caa329922).

elements/rh-dialog/rh-dialog.ts
adamjohnson: Should be resolved in [0ff425f](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/0ff425f3cdd681cf06c163dbb60f117940200dbe).

marionnegp:
> Going forward, which option should we choose:
> 
> Keep the fixed header, make sure header content retains padding
> Nix the fixed header, let header scroll out of view. The only sticky item is the close button.

Ah, I should have checked those docs. I'd say to do the first, keep the header and retain its padding. Thank you!!

elements/rh-dialog/rh-dialog.css
```diff
@@ -1,194 +1,227 @@
...
+  margin-block-start: calc(-1 * var(--offset-top));
 }
 
-header {
+#header {
+  background-color: var(--rh-color-surface);
   position: sticky;
-  top: 0;
-  background-color: var(--rh-color-surface-lightest, #ffffff);
+  inset-block-start: calc(-1 * var(--offset-top));
```
zeroedin, line 91: When viewport height is small the content in the dialog modal is hard to read, maybe undo the `position: sticky` for vertically challenged viewports.

```Suggestion
  inset-block-start: calc(-1 * var(--offset-top));
  @media (max-height: 320px) {
    position: unset;
  }
```

zeroedin: You can also invert that rule starting with the position: unset, and applying sticky on`min-width: 320px`.  mobile first :)

adamjohnson: Good suggestion. I was not thinking about that.

Implemented this in [3b9e903](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/3b9e9033d8c7a04bc2f795cb308a058fa564f643). I didn't think `320px` was adequate, so I stepped up the media query a fair amount. So it should only happen if there's (hopefully) enough space.

If, at this point, it doesn't work for someone, they can override using shadow parts and go up or down in media query pixel width/height as needed.

marionnegp (changes requested):
The space between the header and the body gets lost once content is able to scroll. Can dialog retain the 16px (maybe as padding)?
Screenshot 2024-12-18 at 10 26 23 AM

marionnegp (approved):

elements/rh-dialog/rh-dialog.ts
```diff
@@ -80,93 +84,107 @@ export class RhDialog extends LitElement {
...
-                </header>
+      <div id="rhds-wrapper" class="${classMap({ mobile })}">
+        <!-- @deprecated: 👇 use public vars for the backdrop pseudo-element instead. -->
+        <div id="overlay"
+             part="overlay"
+             ?hidden=${!this.open}>
+        </div>
+        <rh-surface class=${classMap({ hasHeader, hasDescription, hasFooter })}
+                    ?hidden=${!this.open}
+                    color-palette="${ifDefined(this.type === 'video' ? undefined : 'lightest')}">
```
zeroedin, line 152:
```Suggestion
                    color-palette="lightest'"
```
As discussed in Slack, `undefined` for color-palette defaults to lightest.

zeroedin (approved):
Let's get triggering modals

elements/rh-dialog/rh-dialog.ts
adamjohnson: Looks like this attribute is required for dark color palettes.

If you open the demo locally (after removing line 152) and switch to a dark{er/est} color palette without the attribute, the close icon is invisible when focused and light gray when not focused.

elements/rh-dialog/demo/no-headings.html
```diff
@@ -1,13 +1,12 @@
...
-  <p>This modal doesn't have any headings. Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod
-    tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco
-    laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse
-    cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui
-    officia deserunt mollit anim id est laborum.</p>
-  <rh-cta>
-    <a href="#">Call-to-action</a>
-  </rh-cta>
+<rh-dialog accessible-label="This dialog's accessible name" trigger="no-headings-trigger">
+  <p>This dialog uses the <code>accessible-label</code> attribute to define its <code>aria-label</code>.
```
bennypowers, line 3:
```suggestion
  <p>This dialog uses the <code>accessible-label</code> attribute to define its accessible name for assistive technology.
```

elements/rh-dialog/rh-dialog.css
```diff
@@ -1,194 +1,254 @@
...
-  display: flex;
-  position: fixed;
-  height: 100%;
-  width: 100%;
-  top: 0;
-  left: 0;
-  align-items: center;
-  justify-content: center;
-  z-index: 500;
+.visually-hidden {
```
bennypowers, line 13: > [!NOTE]
> Not relevant to this PR, and please don't commit it to this branch, rather a separate PR.

I think it's time to move this and the previous rule into a shared stylesheet module, which would get adopted (via the static `styles` field) into every element.

.changeset/loud-coins-shake.md
```diff
@@ -0,0 +1,21 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-dialog>`: Dialog now uses the native HTML `<dialog>` element internally.
+
+Note: the `overlay` CSS shadow part is now deprecated in favor of the `--rh-dialog-backdrop-background-color` CSS custom property.
```
bennypowers, line 7: does the part still work? if not, this will need a major version bump

```suggestion
Note: the `overlay` CSS shadow part is now deprecated in favor of the `--rh-dialog-backdrop-background-color` CSS custom property. It will still work, but you can expect the `overlay` part to be removed in a future version
```

Also, please confirm @coreyvickery that this deprecation aligns with design expectations: namely that in future versions the only customizable aspect of the backdrop will be its color (including transparency) - and that the existing ability to customize any css property for the backdrop should be removed.

elements/rh-dialog/rh-dialog.css
```diff
@@ -1,194 +1,254 @@
...
+  inset-block-start: 0;
+  margin-block-end: var(--rh-space-lg, 16px);
+  position: relative;
+
+  &:before {
+    background-color: var(--rh-color-surface);
+    block-size: 100%;
+    content: '';
+    inline-size: 100%;
+    inset-block-start: calc(-1 * var(--offset-top));
```
bennypowers, line 98: `--offset-top` should either be `--_offset-top` or `--rh-dialog-offset-block-start`, depending on whether @coreyvickery intends for the offset to be user-customizable.

elements/rh-dialog/rh-dialog.css
```diff
@@ -1,194 +1,254 @@
...
+  &:after {
+    background-color: var(--rh-color-surface);
+    block-size: var(--rh-space-lg, 16px);
+    content: '';
+    inline-size: 100%;
+    position: absolute;
+    z-index: -1;
+  }
+
+  @media (min-height: 768px) {
```
bennypowers, line 112: please leave a comment with the name of the breakpoint token e.g. Please also use the new range syntax

```css
/* --rh-media-sm */
@media (height >= 768px) {
```

elements/rh-dialog/rh-dialog.css
```diff
@@ -1,194 +1,254 @@
...
+  }
 }
 
-header ::slotted(:is(h1,h2,h3,h4,h5,h6)[slot='header']) {
+#content ::slotted([slot='header']) {
+  margin-top: 0 !important;
+}
+
+#header ::slotted(:is(h1,h2,h3,h4,h5,h6)[slot='header']) {
+  font-family: var(--rh-font-family-heading, RedHatDisplay, 'Red Hat Display', 'Noto Sans Arabic', 'Noto Sans Hebrew', 'Noto Sans JP', 'Noto Sans KR', 'Noto Sans Malayalam', 'Noto Sans SC', 'Noto Sans TC', 'Noto Sans Thai', Helvetica, Arial, sans-serif);
```
bennypowers, line 157: does this need to use the new font stack? did we merge that yet?

elements/rh-dialog/rh-dialog.css
```diff
@@ -1,194 +1,254 @@
...
-  padding: 0;
-  text-align: left;
-  position: absolute;
-  cursor: pointer;
-  line-height: 24px;
-  padding-block: 0.375rem;
-  padding-inline: var(--rh-space-lg, 16px);
-  top: 0;
-  right: calc(var(--rh-space-xl, 24px) / -3);
+#body ::slotted(p) {
```
bennypowers, line 162: should this be
```suggestion
#body ::slotted(p:first-of-type) {
```

elements/rh-dialog/rh-dialog.ts
```diff
@@ -45,26 +46,29 @@ async function pauseYoutube(iframe: HTMLIFrameElement) {
...
+ * @cssprop {<color>} [--rh-dialog-overlay-background-color=transparent] @deprecated
+ *          Sets the background color for the `#overlay` `<div>`. Use `--rh-dialog-backdrop-background-color` instead.
  */
 @customElement('rh-dialog')
 export class RhDialog extends LitElement {
-  static readonly version = '{{version}}';
-
   static readonly styles = [styles];
 
   protected static closeOnOutsideClick = true;
```
bennypowers, line 74: i think we can remove this

elements/rh-dialog/rh-dialog.ts
```diff
@@ -80,93 +84,107 @@ export class RhDialog extends LitElement {
...
   #cancelling = false;
+  #lastTabbable: HTMLElement = this.closeButton;
 
   #slots = new SlotController(this, null, 'header', 'description', 'footer');
 
   connectedCallback() {
     super.connectedCallback();
-    this.addEventListener('keydown', this.onKeydown);
+    this.addEventListener('keydown', this.#onKeyDown);
     this.addEventListener('click', this.onClick);
```
bennypowers, line 127:
```suggestion
    this.addEventListener('click', this.#onClick);
```

elements/rh-dialog/rh-dialog.ts
```diff
@@ -80,93 +84,107 @@ export class RhDialog extends LitElement {
...
   }
 
+  disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('keydown', this.#onKeyDown);
```
bennypowers, line 132:
```suggestion
```
This should get cleaned up when the element is removed, since the listener is on the host.

elements/rh-dialog/rh-dialog.ts
```diff
@@ -235,35 +247,51 @@ export class RhDialog extends LitElement {
...
 
-  @bound private onTriggerClick(event: MouseEvent) {
+  @bound private async onTriggerClick(event: MouseEvent) {
     event.preventDefault();
     this.showModal();
+    await this.updateComplete;
+    this.closeButton?.focus();
   }
 
   @bound private onClick(event: MouseEvent) {
```
bennypowers, line 257:
```suggestion
  #onClick(event: MouseEvent) {
```

bennypowers (changes requested):
Last few changes then we'll be ready to rock

.changeset/loud-coins-shake.md
adamjohnson: > does the `overlay` part still work?

Current version of `overlay` [on uxdot](https://ux.redhat.com/elements/dialog/demo/) is the overlay/backdrop of the dialog. 

This PR's version keeps the `overlay` `<div>` and many of the styles (lines 146-149 in `rh-dialog.ts` and lines 35-44 in `rh-dialog.css`).

The only change is `overlay` is a transparent `<div>` overlaying the contents of the page. It's invisible to the user and customizable via the existing CSS part—should a user want to do that.

With this PR's upgrade to the native `<dialog>` element, the [native `::backdrop` property](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop) is being used alongside a customizable CSS variable.

elements/rh-dialog/rh-dialog.css
adamjohnson: Indeed, the following ("public") variables [currently exist on uxdot](https://github.com/RedHat-UX/red-hat-design-system/blob/main/elements/rh-dialog/rh-dialog.css#L145-L147) and are in production—albeit not documented:

```css
--offset
--offset-top
--offset-right
```

That leaves us with a decision:

  1. Do we change these variables as suggested?
      * A breaking change
  3. Or, do we leave these variables as is (and possibly document them)?

elements/rh-dialog/rh-dialog.css
adamjohnson: Screenshot of open dialog with margin only on the first paragraph

Just to confirm, this is how it should look? Seems like each paragraph should have some `margin-block-end` spacing. That said, I am frequently incorrect on these things. :-) 

cc @marionnegp

marionnegp: @adamjohnson, if it's going to follow what we have on redhat.com, body paragraphs should have a `margin-block-end` of 16px before another paragraph.

elements/rh-dialog/rh-dialog.css
adamjohnson: I like this change; however, [in `stylelintrc.yaml`](https://github.com/RedHat-UX/red-hat-design-system/blob/main/.stylelintrc.yml#L27), this syntax is being disallowed and will fail builds.

After making this change, the linter says:

```
stylelint: error
media-feature-range-notation - Expected "prefix" media feature range notation
```

Here's [the docs](https://stylelint.io/user-guide/rules/media-feature-range-notation/) on `media-feature-range-notation`. Looks like that would need to be changed to `context`, probably in a separate PR.

elements/rh-dialog/rh-dialog.css
adamjohnson: If I put:

```css
/* "Old" font stack: */
font-family: var(--rh-font-family-heading, RedHatDisplay, 'Red Hat Display', Helvetica, Arial, sans-serif);
```

And run `npm run lint`, I get an error telling me to use the new font stack:

```
elements/rh-dialog/rh-dialog.css
  157:20  ✖  Expected --rh-font-family-heading to equal RedHatDisplay, 'Red Hat Display', 'Noto Sans Arabic', 'Noto Sans Hebrew', 'Noto Sans JP', 'Noto Sans KR', 'Noto Sans Malayalam', 'Noto Sans SC', 'Noto Sans TC', 'Noto Sans Thai', Helvetica, Arial, sans-serif  rhds/token-values
```

Are you talking about [this tokens issue](https://github.com/RedHat-UX/red-hat-design-tokens/issues/113) and corresponding PR?

tl;dr Seems like keeping the new font stack is the way to go...? Let me know how you would like to proceed.

.changeset/loud-coins-shake.md
adamjohnson: Added Benny's note about removing `overlay` in a future version in [34d7598](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/34d7598e71ee7bbd5fe0563f2f155795f6eb3f36).

elements/rh-dialog/rh-dialog.ts
adamjohnson: Resolved in [a994deb](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/a994deb7cbc3b7c40822e614b2650f6503ae4691).

Alos removed other references to `@bound` in [4d1ecb5](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/4d1ecb59aa4c487e057d11f520e8041d78e67528).

elements/rh-dialog/rh-dialog.ts
adamjohnson: Resolved in [a994deb](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/a994deb7cbc3b7c40822e614b2650f6503ae4691).

elements/rh-dialog/rh-dialog.ts
adamjohnson: Removed in [1848cfa](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/1848cfa36cc8da88195452fcd0de494aaf140f44).

elements/rh-dialog/rh-dialog.css
adamjohnson: Resolved in [db066c1](https://github.com/RedHat-UX/red-hat-design-system/pull/2078/commits/db066c1321d332a00cb64a6c47f42210c30f59d4).

elements/rh-dialog/rh-dialog.css
bennypowers: we'll need to update the tokens package to 2.1.1, but let's do that in a separate PR

elements/rh-dialog/rh-dialog.css
bennypowers: I think we need a patch changeset but one that clarifies that these names were always supposed to be private. WDYT @zeroedin? cc @zhawkins @OneEightyFirst

elements/rh-dialog/rh-dialog.css
bennypowers: https://github.com/RedHat-UX/red-hat-design-system/pull/2106

elements/rh-dialog/rh-dialog.css
adamjohnson: If we wanted to be super conservative, the other option might be to do something like:

```css
#selector {
  /* @deprecated: --offset-top and will be removed in a future release */
  margin-block-start: var(--offset-top, var(--_offset-top)); 👈 
}
```

elements/rh-dialog/rh-dialog.css
adamjohnson: Could you expand on this how you'd recommend implementing this? 

- Are you suggesting using [CSS module scripts](https://web.dev/articles/css-module-scripts) (limited browser support)?
- Would this go in RHDS? eg: `elements/shared/styles/visually-hidden.css`
- Would it go in pfe-tools? eg `@patternfly/pfe-tools/shared/styles/visually-hidden.css`

Happy to get this going with some direction. TIA!

marionnegp:
I'll be making dialog docs updates in this branch/PR this week! Please don't merge before they're done.

.changeset/loud-coins-shake.md
adamjohnson: Paging Mr. @coreyvickery. 😇 

See Benny's questions about overlay/backdrop and `--offset` vars below.

marionnegp:
@coreyvickery, I updated the dialog docs to accommodate some of the changes made in this PR. Could you take a look at the [mockup](https://www.figma.com/proto/B37PNvwXAxVprTSaj11B9G/Element---Dialog?node-id=12001-570&t=W7q1tfkpU22F8npr&scaling=scale-down-width&content-scaling=fixed) and add any feedback?

coreyvickery:
@marionnegp Mock-up looks good.

marionnegp:
@coreyvickery @adamjohnson, feel free to take a look at the [updated docs in the DP](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/).

marionnegp:
@coreyvickery @adamjohnson, feel free to take a look at the [updated docs in the DP](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/).

.changeset/loud-coins-shake.md
adamjohnson: Since everything is backwards compatible and Benny's requested changes have been made, I'm going to resolve this convo. @coreyvickery if this depreciation does _not_ align with design expectations, feel free to unresolve.

elements/rh-dialog/rh-dialog.css
adamjohnson: We should chat about this during office hours sometime. I'm interested to hear how we'd go about implementing this and think it's a great thing to do.

For now, going to resolve this one.

elements/rh-dialog/docs/20-guidelines.md
```diff
@@ -1,137 +1,27 @@
...
 ### Video player dialog
 
-A video can replace a fixed-width dialog container, it should have the same 
-width and include a close button.
+A video dialog should have the same aspect ratio as the video and include a close button.
 
 <uxdot-example width-adjustment="1000px" variant="full" alignment="left" no-border>
-  <img src="../dialog-video-player.png"
-        alt="A dialog video player spanning a 12-column grid with a white close button"
+  <img src="../dialog-patterns-video-player.svg"
```
adamjohnson, line 19: @marionnegp This image is missing the close "X" in the upper right.

marionnegp: Done, thanks for catching that!

coreyvickery:
@marionnegp Let me know if you still need my eyes on this.

marionnegp:
@coreyvickery, I think we're good. Thank you!

bennypowers (approved):
lotsa green tests, man!

markcaron:
@adamjohnson it doesn't seem like Dialog's dark theme was taken care of in this PR ([see demo](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/demo/color-context/)). May need a new one.

markcaron:
@adamjohnson, although, our [Dialog docs](https://deploy-preview-2078--red-hat-design-system.netlify.app/elements/dialog/style/#theme) still say it's available only in Light. However, there is a [Dark version](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/RHDS-component-library---v2.1.2?node-id=65-150&p=f&t=wtsVjAssrltUGFl8-0) of the component.

CC: @coreyvickery @marionnegp

adamjohnson:
Yup, I specifically kept it as light only in this PR due to the existing docs.

As mentioned, it shouldn't be too tough to add functionality for dialog to use a dark color scheme—but we would have to update the docs accordingly.

PR #2087, opened 2024-12-12
docs(card): demo of cards in flex container

### LINKED ISSUE #2044: `<rh-card>`: when in parent flex container
Card collapses when in a parent flex container, we should investigate this and understand why and document that for other users

https://codepen.io/zeroedin/pen/poMxXmB

Thanks @zhawkins for reporting

bennypowers:
Closes #2044 

## What I did

1. add demo of card in flex container

## Testing Instructions

1. tbd

### DIFF
--- a/elements/rh-card/demo/in-flex-container.html
+++ b/elements/rh-card/demo/in-flex-container.html
@@ -1,0 +1,104 @@
+<div class="flex">
+  <rh-card>
+    <svg slot="header" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 522 191.4">
+      <defs>
+        <clipPath id="clippath">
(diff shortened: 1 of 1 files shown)


zeroedin:
I think we just need to apply a width to the host

```css
:host {
  width: 100%;
}
```

zeroedin:
went ahead and added this change as well as nested the CSS if you feel like it's too much feel free to revert and just add the width to 100% to the host and call it a day.

bennypowers:
since this changes the card element css (and ts) it needs a changeset
also maybe this is a breaking change?

zeroedin:
I need to revisit this. This PR got buried in other priority work. Need to investigate whether other changes to rh-card in the mean time haven't already resolved this issue.

PR #2088, opened 2024-12-12
docs(icon): remove sets demo

bennypowers:
Users should instead see https://ux.redhat.com/foundations/typography

## What I did

1. remove the icon sets demo


zeroedin (commented):
We should add an alert at the top of the `rh-icon` demo page rerouting people to the iconography page.  There has been confusion about where to find the icons with the users I've worked with, this would be helpful.

```html
<rh-alert state="info">
  <h2 slot="header">Looking for the icons?</h2>
  <p>All the icons you might need are right at your fingertips! Head over to our <a href="/foundations/iconography/">iconongraphy</a> foundations page to explore the full collection.</p>
</rh-alert>
```
or something to this effect

image

zeroedin:
@bennypowers I went ahead and dropped a commit adding that alert with the link to the iconography page for your review code wise.  @marionnegp what are your thoughts on this?  Let me know if you want the language changed.

marionnegp (changes requested):

docs/_includes/layouts/pages/element.11ty.ts
```diff
@@ -900,16 +900,24 @@ export default class ElementsPage extends Renderer<Context> {
...
+      iconAlert,
+      entries.map(([filename, config]) => this.#renderPlayground(
+        filename,
+        config,
+        ctx,
+        common,
+        entries
+            .filter(([, config]) => config.inline === filename)
+            .map(([s]) => s)
+      )),
```
marionnegp, line 919: Removing the sets demo is a good thought! I'm seeing an extra comma after the alert, but that's it.

marionnegp (changes requested):

marionnegp (approved):

PR #679, opened 2022-12-16
feat(audio-player): add rh-audio-player

nikkimk:
### rh-menu
(based on [Audio-player XD](https://xd.adobe.com/view/149bfa72-9616-4435-abf8-d3790a17ef29-41c8/grid))
- [Menu](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/menu/demo)

### rh-audio-player
(based on [Audio-player XD](https://xd.adobe.com/view/149bfa72-9616-4435-abf8-d3790a17ef29-41c8/grid))
- [Audio player](https://deploy-preview-679--red-hat-design-system.netlify.app/components/audio-player/demo)
- [Customizations](https://deploy-preview-679--red-hat-design-system.netlify.app/components/audio-player/demo/customization)
- [Prevent concurrent playback](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/audio-player/demo/prevent-concurrent-playback.html)
- [Detailed transcript](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/audio-player/demo/detailed-transcript.html)
- [Language localization](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/audio-player/demo/language-localization.html)
- [Right-to-left](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/audio-player/demo/right-to-left.html)
- [Docs](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/audio-player) based on [spec Doc](https://docs.google.com/document/d/1TcDrLGN5PbYem54SUz62bMNJt-vYhBQJI7jOsWItN1o/edit)

Closes #637 and #1072

PR #2095, opened 2024-12-16
fix(audio-player): use rh-icon instead of custom svg for scrubbers

### LINKED ISSUE #2049: [feat] update icons `<rh-audio-player>`
### Description

The rewind and fast forward buttons are currently custom SVGs, but they can be updated to `<rh-icon>`. The rewind button should use `<rh-icon set="ui" icon="undo"></rh-icon>`. The fast forward can use `<rh-icon set="ui" icon="redo"></rh-icon>`. 

### Suggested solution

- [ ] Design done (update icons in the Figma library)
- [ ] Update in web component
- [ ] Update docs images

### Screenshots

_No response_

### Example API

_No response_

### Additional context

_No response_

zeroedin:
## What I did

1.  Replaced custom svg's with `rh-icon` for scrubbers

Closes #2049

### TODO:

- [ ] Docs images as listed in issue

## Testing Instructions

1. [View deploy preview](https://deploy-preview-2095--red-hat-design-system.netlify.app/elements/audio-player/demo/)

## Notes to Reviewers

@marionnegp I added the rh-icon with the same size as the play button (24px) as that looked closest to the previous state. If it should be a different size let me know.  I'll leave PR in draft and if you want to make the changes to the docs images on this same branch go right ahead.


marionnegp (changes requested):
The scrubber icon size looks good to me!

- [ ] Can the disabled scrubber buttons be `--rh-color-gray-30` (light theme) and `--rh-color-gray-50` (dark theme) for now? (We'll eventually add disabled color tokens.)

The rest of these can be updated in a separate PR if you'd prefer:

- [ ] Speed controls don't seem to be using the right color in dark theme.
Screenshot 2024-12-17 at 12 17 32 PM

- [x] Play button in light theme player should be in a `darkest` circle with a white play icon:
Screenshot 2024-12-17 at 12 20 53 PM

- [ ] Play button in dark theme player should be in a white circle with a `darkest` play icon:
Screenshot 2024-12-17 at 12 22 23 PM

zeroedin:
> * Speed controls don't seem to be using the right color in dark theme.

This looks like it opens up a bigger issue on audio player.  Will need to dig deeper tomorrow to see what is going on, Color context from the outside leaks into audio player and it has an internal surface which doesn't set to match the parent context:

Example of issue:
```html
<rh-context-demo color-palette="darkest"> <!--sets context to dark-->
  <rh-audio-player> <!-- dark background get set -->
    <!-- shadowroot-->
       <rh-surface> <!-- internal context is set to light (note no color palette) -->
          Text / Buttons <!-- all think they are on a light background -->
   </rh-audio-player>
 </rh-context-demo>
 ```

**Comment by zeroedin** [comment]
@bennypowers I can't believe the resolution was as simple as removing the `rh-surface` from the shadow root. However, I couldn't find a reason for it. Can you double-check me here? I may be overlooking something.

**Comment by zeroedin** [comment]
> Play button in light theme player should be in a darkest circle with a white play icon:

@marionnegp, I'll look at these changes with the circle not being present on the play icon now. The circle was likely part of the custom SVG previously.

**Comment by zeroedin** [comment]
> The circle was likely part of the custom SVG previously.

This was not a true statement 😆 , that said I'm now adjusting the variables used directly based on the context.

```css
#play,
#full-play {
  border-radius: 50%;
  background-color: var(--rh-color-surface-darkest, #151515);
  color:
    var(--rh-audio-player-icon-background-color,
      var(--rh-color-text-primary-on-dark, #ffffff));
}

.dark {
  #play,
  #full-play {
    background-color: var(--rh-color-surface-lightest, #ffffff);
    color:
      var(--rh-audio-player-icon-background-color,
        var(--rh-color-text-primary-on-light, #151515));
  }
}
```

zeroedin:
image

Another bug found, the highlight in the text being read in the transcript also appears to be off. 

Does not appear to be new / introduced in this PR either, [also found on ux.redhat.com](https://ux.redhat.com/elements/audio-player/demo/mini/)

image

zeroedin:
> I think we've talked about this before somewhere. Déjà vu?

It was in `rh-video-embed.` it looks like the difference is we need to use an `rh-button variant="play"` which will get us the adjusted variant we are looking for. Looking to see if we can reuse that here.

marionnegp:
> Another bug found, the highlight in the text being read in the transcript also appears to be off.

The docs show images of the [highlight](https://ux.redhat.com/elements/audio-player/features/#highlighting) for light theme, and that blue looks most like `--rh-color-blue-20`. @coreyvickery, did you specify a highlight color for light and dark themes?

Looking at Nikki's original [audio player PR](https://github.com/RedHat-UX/red-hat-design-system/pull/679) and the [transcript demo](https://deploy-preview-679--red-hat-design-system.netlify.app/elements/audio-player/demo/detailed-transcript/), it might have been using the default browser highlight. Maybe this was a bug introduced when we made changes to the theming?

zeroedin:
> Maybe this was a bug introduced when we made changes to the theming?

I believe that is the case, yes.

elements/rh-audio-player/rh-audio-player.css
```diff
@@ -444,11 +443,30 @@ rh-menu > button:focus {
...
+  /* danger! overwriting an rh-button internal var, do what I say not as I do. */
+  --_play-background-color: var(--rh-color-surface-darkest, #151515);
+}
+
+.dark {
+  #play::part(button),
+  #full-play::part(button) {
+    /* danger! overwriting an rh-button internal var, do what I say not as I do. */
+    --_play-background-color: var(--rh-color-surface-lightest, #ffffff);
+  }
```
zeroedin, line 469: I don't like doing this, but I must override an internal CSS variable in the rh-button to match the Figma styles. @bennypowers, as you suggested, we can break this out into a separate issue. Maybe go with the broken styles while still using the `rh-button variant="play"` in this PR. @marionnegp, what are your thoughts?

zeroedin: After a discussion with @marionnegp we are going to back out of the usage of `rh-button variant=play` for now and reimplement the styles needed to adjust the icon that is already in rh-button so we can release this PR still as a patch. There was another issue with this approach, and that is that the `rh-button variant=play` doesn't support swapping its icon to a pause button easily, even if we can just use the part to override the background colors necessary for audio-player.

zeroedin:
Given changes in theming between the opening of this PR and its becoming stale, I recommend closing this PR in favor of opening a newly cut branch of the current state main to deal with this issue and reimplement the changes instead of dealing with the conflicts that are bound to happen here.  Leaving it open till a new PR is cut that references this one.

zeroedin:
Close in favor of: https://github.com/RedHat-UX/red-hat-design-system/pull/2469

PR #2096, opened 2024-12-18
fix(tile): float footer arrow when footer is empty

bennypowers:
Addresses https://github.com/orgs/RedHat-UX/discussions/2086

This PR is mostly a straw proposal intended to provoke discussion on what the best solution to the problem is. If we end up merging it, because it suits our purposes, that's great too, but there's no pressure that this specific solution has to be the one that's adopted.

## What I did

1. collapse the footer whitespace when empty. i.e. position the footer arrow absolutely in that case.

## Testing Instructions

1. See [Demo of tile grid without footer content](https://deploy-preview-2096--red-hat-design-system.netlify.app/elements/tile/demo/without-footer-content/)


zeroedin (commented):
I dig it!

bennypowers:
I feel like I should specify that I took one possible approach to solving this problem here in this PR, and it might not be the best approach. There could be other approaches, and we could discuss those if we want.

markcaron:
@marionnegp can you look at this? I think we need design sign off

marionnegp:
I think the arrow gets too close to the body text. I can see if constraining the width of the body slot (and maybe header slot) will help without limiting the amount of space too much.
Screenshot 2025-03-20 at 9 52 42 AM

bennypowers:
we might be able to do it with a clip path.. iirc there's something for flowing text around stuff

marionnegp:
> we might be able to do it with a clip path.. iirc there's something for flowing text around stuff

That might preferable to adding a max-width for the tile content. :sweat_smile:

zeroedin (approved):
Link Gap, Troubled Margins

marionnegp (approved):

bennypowers:
Should have a patch changeset

.changeset/hip-lamps-rule.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tile>`: Fixed whitespace when footer slot is empty.
```
bennypowers, line 5:
```suggestion
`<rh-tile>`: improved layout when footer slot is empty.
```

markcaron: Sounds good

PR #2106, opened 2025-01-05
chore: tokens 2.1.1

bennypowers:
## What I did

1. bump tokens version
2. lint css files

## Testing Instructions

1. check i18n text in demos (e.g. cta rtl demos)


zeroedin (commented):
I think we should add the additional required CSS in the RTL demos as suggested in the [tokens repo changeset](https://github.com/RedHat-UX/red-hat-design-tokens/pull/139/files#diff-cf4a04795e2b6d5af32afa66403a53ca428986b048ad901ce8e7f26b6ccdc16f) 

```
[lang=" he"] {
  font-family: var(--rh-font-family-lang-he: "Noto Sans Hebrew");
}
```

We might also mention this explicitly in the changelog for RHDS @zhawkins, I believe you have mentioned that when we make changes to the dependencies, sometimes those changes feel as if they get lost in RHDS's changelog, do you have any opinions on how that would get called out?

adamjohnson (changes requested):
These changes look good.

It does look like there are still references to `Noto *` in a few other places like:

  * `docs/get-started/index.md` - `<style>` tag
  * `elements/rh-card/demo/color-context.html` - `<style>` tag
  * `elements/rh-card/demo/sticky-pattern.html` - `<style>` tag
  * `elements/rh-card/demo/title-pattern.html` - `<style>` tag
  * `elements/rh-card/demo/variants.html` - `<style>` tag
  * Should the [Typography Foundations](https://deploy-preview-2106--red-hat-design-system.netlify.app/foundations/typography/#international) table list the other languages?

Thanks for your work on this.

bennypowers:
> I think we should add the additional required CSS in the RTL demos as suggested in the [tokens repo changeset](https://github.com/RedHat-UX/red-hat-design-tokens/pull/139/files#diff-cf4a04795e2b6d5af32afa66403a53ca428986b048ad901ce8e7f26b6ccdc16f)
> 
> ```
> [lang=" he"] {
>   font-family: var(--rh-font-family-lang-he: "Noto Sans Hebrew");
> }
> ```
> 
> We might also mention this explicitly in the changelog for RHDS @zhawkins, I believe you have mentioned that when we make changes to the dependencies, sometimes those changes feel as if they get lost in RHDS's changelog, do you have any opinions on how that would get called out?

If this is required, it's RHDS v3

zhawkins:
Thanks for the work on this y'all!

> We might also mention this explicitly in the changelog for RHDS @zhawkins, I believe you have mentioned that when we make changes to the dependencies, sometimes those changes feel as if they get lost in RHDS's changelog, do you have any opinions on how that would get called out?

🤔 I'm not really sure, you could use a star emoji? 😁

zeroedin:
> If this is required, it's RHDS v3

Maybe required was too strongly worded.  If the failover to `Helvetica, Arial, sans-serif` is enough to cover the correct character rendering for other languages then it wouldn't be required per say.  I think what I'm after is should "we" be required to show a complete example of what "should" be done in our demos. Noto Hebrew will not load otherwise in the RTL demos we use Hebrew text.

bennypowers:
> > If this is required, it's RHDS v3
> 
> Maybe required was too strongly worded. If the failover to `Helvetica, Arial, sans-serif` is enough to cover the correct character rendering for other languages then it wouldn't be required per say. I think what I'm after is should "we" be required to show a complete example of what "should" be done in our demos. Noto Hebrew will not load otherwise in the RTL demos that use Hebrew text.

This should be a separate PR. I think it should be part of a base style sheet, and that stylesheet should be shared (i.e. adopted) into each element's shadow root as well. This is a separate issue and will require discussion. The first iteration should be quite minimal - lang, h1-6, p, a, etc. - no classes.

zeroedin:
> This should be a separate PR. I think it should be part of a base style sheet, and that stylesheet should be shared (i.e. adopted) into each element's shadow root as well. This is a separate issue and will require discussion. The first iteration should be quite minimal - lang, h1-6, p, a, etc. - no classes.

Sounds good to me

docs/foundations/typography/index.md
```diff
@@ -71,7 +71,7 @@ accommodate mixed-language content. For more technical details, go to the
 <rh-table>
 
 | Language | Font family                  |
-|----------|------------------------------|
+| -------- | ---------------------------- |
 | Chinese  | Noto Sans Simplified Chinese |
 | Japanese | Noto Sans Japanese           |
 | Korean   | Noto Sans Korean             |
```
adamjohnson, line 77: Should we list the other languages here? Or nuke the table and have people refer to the [font tokens](https://deploy-preview-2106--red-hat-design-system.netlify.app/tokens/font/) page?

adamjohnson (approved):
Languages get total makeover

PR #2116, opened 2025-01-10
fix(pagination): Colors, right to left demo, and logical properties

adamjohnson:
## What I did

1. Add fallback colors to fix incorrect colors from undefined color token variables
1. Fix background color being incorrectly applied to `<a>` tags in the `open` variant
1. Change `width`/`height`/`top`/`bottom` etc to use logical properties
1. Fix Right to Left demo layout

## Testing Instructions

1. First, view the [rh-pagination demo on uxdot](https://ux.redhat.com/elements/pagination/demo/).
1. Note the lack of proper colors
    * Active page borders
    * No border around stepper `<input>`
    * Link color incorrect 
1. Now view the rh-pagination demo from this DP.
    * Colors are correct as specified in [the docs](https://ux.redhat.com/elements/pagination/style/#variants)
1. Compare the [open variant demo on uxdot](https://ux.redhat.com/elements/pagination/demo/open/) vs the open variant on this DP. Make sure it matches specs.
1. Look at the [RTL demo on uxdot](https://ux.redhat.com/elements/pagination/demo/right-to-left/) (it's broken!). Now compare that to [this DP's RTL demo](#).

## Notes to Reviewers

Ensure that this is the proper way to handle adding fallback colors to things like `--rh-color-interactive-primary-default`.

Note: the [color context demo on uxdot](https://ux.redhat.com/elements/pagination/demo/color-context/) works because the component is wrapped in a `<rh-surface>`. These fallbacks are in place for when the component is not wrapped in a `<rh-surface>` component.

### DIFF
--- a/elements/rh-pagination/rh-pagination-lightdom.css
+++ b/elements/rh-pagination/rh-pagination-lightdom.css
@@ -121,10 +121,6 @@ rh-pagination:is([overflow='start'], [overflow='both']) li:first-child:after {
  * OPEN VARIANT
  *****************************************************************************/
 
-rh-pagination[variant|='open'] {
-  --_list-a-bg-color: transparent;
-}
-
 rh-pagination[variant|='open'] li > a:hover:after,
 rh-pagination[variant|='open'] li > a:focus:before,
 rh-pagination[variant|='open'] li > a:focus:after,
@@ -134,7 +130,10 @@ rh-pagination[variant|='open'] li > a[aria-current]:after {
   inset-block-end: -1px;
 }
 
+rh-pagination[variant|='open'] li > a {
+  --_list-a-bg-color: transparent;
+}
+
 rh-pagination[variant|='open'] li > a[aria-current] {
-  background-color: var(--_list-a-bg-color);
   border-color: transparent; /* NOTE: Keep active/hover borders in-line */
 }
--- a/elements/rh-pagination/rh-pagination-lightdom.css
+++ b/elements/rh-pagination/rh-pagination-lightdom.css
@@ -22,11 +22,11 @@ rh-pagination li > a {
   display: grid;
   font-family: var(--rh-font-family-heading, RedHatDisplay, 'Red Hat Display', 'Noto Sans Arabic', 'Noto Sans Hebrew', 'Noto Sans JP', 'Noto Sans KR', 'Noto Sans Malayalam', 'Noto Sans SC', 'Noto Sans TC', 'Noto Sans Thai', Helvetica, Arial, sans-serif);
   font-size: var(--rh-font-size-body-text-xl, 1.25rem);
-  height: var(--_link-size);
+  block-size: var(--_link-size);
   place-content: center;
   position: relative;
   text-decoration: none;
-  width: var(--_link-size);
+  inline-size: var(--_link-size);
 }
 
 rh-pagination[size='sm'] li > a {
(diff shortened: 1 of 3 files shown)


elements/rh-pagination/rh-pagination.css
```diff
@@ -20,17 +20,23 @@
   gap: var(--rh-space-2xl, 32px) calc(var(--rh-space-xs, 4px) / 2);
   flex-wrap: wrap;
 
-  --rh-pagination-accent-color: var(--rh-color-interactive-primary-default);
+  /* stylelint-disable-next-line */
```
bennypowers, line 23: why do we have to disable this rule? it's there for a reason. perhaps there's a bug lower down in the stack we need to fix? Try putting it in a surface.

bennypowers (changes requested):
really like the logical properties! nice work!

Let's find a more comprehensive solution for the computed theme properties, one that doesn't require us to disable lint rules.

elements/rh-pagination/rh-pagination.css
adamjohnson: Thanks for the review.

This is the same issue we're seeing in [this comment](https://github.com/RedHat-UX/red-hat-design-system/pull/2043#discussion_r1916745541) on #2043.

If pagination is **not** wrapped in a surface, the tokens are undefined; thus, all the colors are incorrect. See the note in this PR's body under "Notes to reviewers".

elements/rh-pagination/rh-pagination-lightdom.css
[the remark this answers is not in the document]
adamjohnson: Resolved in [c4ee58d](https://github.com/RedHat-UX/red-hat-design-system/pull/2116/commits/c4ee58de1e22e699ceb8c47d2b0bc90ca310a0bc). PTAL and make sure this is a `minor`.

elements/rh-pagination/rh-pagination.css
adamjohnson: Fixed in [b418e4e](https://github.com/RedHat-UX/red-hat-design-system/pull/2116/commits/b418e4edb57c78824d44bb73d5c552b135ecdf5d), [5d77624](https://github.com/RedHat-UX/red-hat-design-system/pull/2116/commits/5d77624fafafeedf923ef5e6e6c044761b64622d), and [a7e352d](https://github.com/RedHat-UX/red-hat-design-system/pull/2116/commits/a7e352d41262edd7e73d7cdb43c32e5a6662469e).

.changeset/dry-icons-love.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": minor
```
bennypowers, line 2:
```suggestion
"@rhds/elements": patch
```

.changeset/dry-icons-love.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+`<rh-pagination>`: fix pagination colors by wrapping elements in `<rh-surface>`, use logical properties
```
bennypowers, line 5:
```suggestion
`<rh-pagination>`: improves layouts for non-left-to-right languages
```

The addition to surface isn't germane to the changelog, since it's a docs issue. However the use of logical properties is a user-facing improvement, so it's helpful to describe here what the benefit of the change is.

bennypowers (changes requested):
The docs pages must specify that rh-surface is required. It _may_ be also indicated that this requirement could be relaxed in the future.

adamjohnson:
@bennypowers See Pagination > [Code](https://deploy-preview-2116--red-hat-design-system.netlify.app/elements/pagination/code/#displaying-correct-pagination-colors) in the DP for words around the requirement of `<rh-surface>`. PTAL, TY!

bennypowers (approved):

PR #2119, opened 2025-01-13
docs: load demo deps early on

### LINKED ISSUE #2097: [bug][docs] `<rh-button>` second, empty button appearing in WebKit
### Describe the bug

At the docs site, a second, empty button typically appears next to `<rh-button>` in Mac and iOS Safari. This can be seen at the [button demo page](https://ux.redhat.com/elements/button/demo/) and for other elements where we're using `rh-button` (e.g., tooltip).

Word on the street is that this is an SSR-related issue.

### Screenshots

Double button at docs site

### Steps to reproduce

1. Go to the [`rh-button` demo page](https://ux.redhat.com/elements/button/demo/) in Safari.
2. Observe the buttons.

### Expected behaviour

Only the real button should appear, not the impostor.

### Operating System (OS)

macOS and iOS

### Browser

Safari 18.2

bennypowers:
Closes #2097

## What I did

1. see discussion thread in #2097: basically by loading button in the same inline module as the lit-html hydration support module

## Testing Instructions

1. check out the pr preview demos

## Notes to Reviewers
it could be that webkit executes each static import within each inline module in source order (good!) but not necessarily each inline module in html-source order (potentially confusing) such that if we load up the hydration support module in the head, and rh-button in the body, rh-button might get loaded first, and then only afterwards hydration support (which patches lit-html's render faculty), leading to double-render.


zeroedin:
Likely related, on the DP in Safari, the planned tag is loading twice in the side nav.
image

Also, the `uxdot-header` has two shadowroot containers

image
resulting in rendering as:
image

I at first thought this was cached, but clearing the cache did not resolve.

bennypowers:
@zeroedin PTAL, I tried to boil things down as much as possible. it looks good to me on the DP

PR #2138, opened 2025-01-22
perf!: convert `color-palette` to CSS via `light-dark()`

### LINKED ISSUE #2151: `<rh-accordion>`: color context demo bug/misrepresentation
In the [color-context demo](https://github.com/RedHat-UX/red-hat-design-system/blob/main/elements/rh-accordion/demo/color-context.html) for accordion, we are placing a `color-palette` attribute on the `rh-accordion-header` however it is not a color-context provider so it should not have this attribute. 

The sibling `rh-accordion-panel`, however, is a color context provider. But I believe this is a mistake. It is my understanding of the design that we wouldn't want a specific panel to be different than the rest of the accordion, much less the header that controls it. I relate this to that same decision in tabs where the panel isn't a provider and shouldn't be separated in color from the tabs that control it.

@coreyvickery @marionnegp thoughts?

Here is the demo code:

https://github.com/RedHat-UX/red-hat-design-system/blob/d1182b0da15c17e64cad80681d7897af64bd5d09/elements/rh-accordion/demo/color-context.html#L11-L17

This came up when I worked with the context PR and corrected it there. I meant to write this issue up. @Nouveau caught that API change and reminded me I needed to bring this forward as a possible bug/misrepresentation.

Here is the corrected demo code:

https://github.com/RedHat-UX/red-hat-design-system/blob/a95f9d811152e167e271e48e8c4c1be5bd056069/elements/rh-accordion/demo/color-context.html#L9-L17

And the commit that changes the accordion panel to no longer be a context provider on that PR:

https://github.com/RedHat-UX/red-hat-design-system/pull/2138/commits/28d5253143550907e3de2faccf0a63af5992fd2d

If it is true that the accordion panel should not be a provider, we should fix that in advance of the context PR changes.

zeroedin:
## What I did

This is an exploration of using [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) as the basis for our context color switching. By utilizing the [light-dark()](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark) color function, we can reduce the overhead of additional Javascript that is not currently providing any more functionality than what these CSS properties and the cascade can do.

Blocked by https://github.com/RedHat-UX/red-hat-design-tokens/pull/160

## Testing Instructions

1. [View the Deploy preview](https://deploy-preview-2138--red-hat-design-system.netlify.app/)
1. [Read the revised theming docs](https://deploy-preview-2138--red-hat-design-system.netlify.app/theming/)

### View context demos
1. [`<rh-accordion>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/accordion/demo/color-context/)
2. [`<rh-announcement>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/announcement/demo/color-context/)
1. [`<rh-alert>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/alert/demo/color-context/)
1. [`<rh-audio-player>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/audio-player/demo/color-context/)
1. [`<rh-avatar>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/avatar/demo/color-context/)
1. [`<rh-back-to-top>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/back-to-top/demo/color-context/)
1. [`<rh-badge>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/badge/demo/color-context/)
1. [`<rh-blockquote>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/blockquote/demo/color-context/)
1. [`<rh-breadcrumb>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/breadcrumb/demo/color-context/)
1. [`<rh-button>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/button/demo/color-context/)
1. [`<rh-card>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/card/demo/color-context/)
1. [`<rh-code>`-block](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/code-block/demo/color-context/)
1. [`<rh-cta>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/call-to-action/demo/color-context/)
1. [`<rh-dialog>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/dialog/demo/color-context/)
1. [`<rh-footer>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/footer/demo/)
1. [`<rh-health-index>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/health-index/demo/color-context/)
1. [`<rh-menu>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/menu/demo/color-context/)
1. [`<rh-navigation-secondary>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/navigation-secondary/demo/dark-variant/) (use devtools to remove color-palette)
1. [`<rh-pagination>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/pagination/demo/color-context/)
1. [`<rh-site-status>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/site-status/demo/color-context/)
1. [`<rh-spinner>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/spinner/demo/color-context/)
1. [`<rh-stat>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/statistic/demo/color-context/)
1. [`<rh-subnav>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/subnavigation/demo/color-context/)
1. [`<rh-surface>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/surface/demo/color-palettes/)
1. [`<rh-switch>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/switch/demo/color-context/)
1. [`<rh-table>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/table/demo/color-context/)
1. [`<rh-tabs>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/tabs/demo/color-context/)
1. [`<rh-tag>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/tag/demo/color-context/)
1. [`<rh-tile>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/tile/demo/color-context/)
1. [`<rh-tooltip>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/tooltip/demo/color-context/)
1. [`<rh-video-embed>`](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/video-embed/demo/color-context/)

Closes #2151 

## Notes to Reviewers

Some components had their CSS modified to accommodate this approach. Avatar and Breadcrumbs used the `.dark` consumer class to swap images. I took a couple of different approaches in this PR. 

In Avatar, the default SVG was swapped in the render() based on context; in this PR, it now styles the SVG internals using `light-dark()` and just loads a single SVG in the render(), no swapping necessary. 

Breadcrumbs used data URI SVG and swapped the URI based on the `.dark` class. Unfortunately, we can not use `light-dark()` to swap a background image, so instead, I used the data URI SVG as a mask and then used `light-dark()` to style the background color the mask cuts out. Same result, different approach.

~~Other than that, the addition of the `color-scheme.css` is necessary. My assumption is that maybe this file could be served by the tokens repo as part of global.css or as a separate file.~~

Please think about these changes and future-forward feature sets that will be here soon (i.e., style queries). Am I missing something? Is there potential where this doesn't work? 

In this PR, I personally did not find a current use case that it doesn't work, only places we needed to take a slightly different approach mentioned above.


bennypowers:
Really impressive!

- Shouldn't we use `only` for when color-palette is applied, and `light dark` when it's absent? This could be in a shared stylesheet 
- Instead of putting the `light-dark()` switch in every element, at evey callsite, it should be in the shared style sheet exported from tokens, element sheets would continue to reference the computed values. It could be that the only changes necessary in element sheets should be selectors (i.e. `.on.dark`)
- We might just be able to get away with making this a minor or even patch release, due to previous design choices for the apis like insisting on private `on`

zeroedin:
> * Shouldn't we use `only`

Yes looks like we should.  

> Instead of putting the light-dark() switch in every element, at evey callsite, it should be in the shared style sheet exported from tokens, element sheets would continue to reference the computed values. 

Yes need to go through the changes I've made here with a fine tooth comb and ensure the changes in the element sheets can't be replaced with the computed value already present in `color-scheme.css`.  I went through pretty quickly mainly looking for 'gotchas' that would disprove the idea.

> * We might just be able to get away with making this a minor or even patch release

That was my take away with the changes made, that there wasn't anything fundamentally changing for the user.

zeroedin:
> * We might just be able to get away with making this a minor

Thinking about this more, what if someone has written a 3rd party component that relies on `rh-surface` to provide the context? Because `rh-surface` in this PR no longer provides this through an event could that be considered a breaking change for external usage? Is there a need to deprecate the JS portion and still have it available?

bennypowers:
> gotchas

`dark-alt`, `saturated`, and `base` are your gotchas. of the three, `dark-alt` is the only one in current use. that's why it needs to be removed and replaced with colour transforms in oklch. The existence of `dark-alt` screws with the maths of the rest of the theming system. cc @marionnegp @coreyvickery

bennypowers:
> > * We might just be able to get away with making this a minor
> 
> Thinking about this more, what if someone has written a 3rd party component that relies on `rh-surface` to provide the context? Because `rh-surface` in this PR no longer provides this through an event could that be considered a breaking change for external usage? Is there a need to deprecate the JS portion and still have it available?

I'm preemptively invalidating that user's abuse of the context system.

zeroedin:
> I'm preemptively invalidating that user's abuse of the context system.

I'm even just talking about this:
```
<rh-surface>
  <random-element> <!-- is a @colorContextConsumer() -->
</rh-surface>
```

If we say surface no longer provides the means to satiate `@colorContextConsumer()` and that they need to update their component to use `light-dark()` internally instead. While it's not breaking internally to RHDS it may lead to 3rd party components no longer working without the upgrade warning of a breaking, yes? no?.

zeroedin:
> `dark-alt`, `saturated`, and `base` are your gotchas

These 3 things are not "gotchas" from my position. 

If anything, `saturated` and ' base` might really be themes; I don't think they are context (light, dark). Also, we haven't even implemented those anywhere; it's been a fragment of an idea from inception from my point of view. Definitely, it's something to clarify between us and design, and I look forward to those conversations.

`dark-alt` as a "surface" color is a strange bird indeed as it is currently not tied to "provider" [color-palette] attr. But, AFAIK, it is usually directly applied in specific instances rather than using a `light-dark` computed value. I think that is likely 'ok'. Do we want a `color-palette="dark-alt"`? Maybe not, but we could do that if we really find it necessary; just not sure it is at this time.

bennypowers:
> > I'm preemptively invalidating that user's abuse of the context system.
> 
> I'm even just talking about this:
> 
> ```
> <rh-surface>
>   <random-element> 
> </rh-surface>
> ```
> 
> If we say surface no longer provides the means to satiate `@colorContextConsumer()` and that they need to update their component to use `light-dark()` internally instead. While it's not breaking internally to RHDS it may lead to 3rd party components no longer working without the upgrade warning of a breaking, yes? no?.

could be, but i sincerely doubt this is a real case in the wild

> If anything, `saturated` and `base` might really be themes;

100%. they should be removed from the context system and treated as themes

> `dark-alt` ...  if we really find it necessary;

the reasons we find it necessary are for contrast in specific scenarios: a better solution is colour transforms in oklch

zeroedin:
> a better solution is colour transforms in oklch

Yeah, if you are just suggesting the removal of `dark-alt` and converting `dark` with a transform for those cases, I'm all for that as well.

marionnegp:
> Do we want a color-palette="dark-alt"? Maybe not, but we could do that if we really find it necessary; just not sure it is at this time.

This might be a dumb question, but if we don't add it as a color palette, does this prevent us from using it as a surface token in an element?

> If anything, saturated and base might really be themes;

>> 100%. they should be removed from the context system and treated as themes

Agreed. We may have a need for more than just light and dark theme in the future, but right now, designers are focusing on getting at least those two themes working for stuff like docs.redhat.com. Better to tackle any saturated or base theming later.

> Thinking about this more, what if someone has written a 3rd party component that relies on rh-surface to provide the context? Because rh-surface in this PR no longer provides this through an event could that be considered a breaking change for external usage? Is there a need to deprecate the JS portion and still have it available?

>> I'm preemptively invalidating that user's abuse of the context system.

Should `rh-surface` be used only to give RHDS elements and patterns context ideally?

elements/rh-badge/rh-badge.css
```diff
@@ -12,19 +12,25 @@ span {
   font-weight: 700;
   line-height: var(--rh-line-height-body-text, 1.5);
   padding-inline: var(--_badge-padding, var(--rh-space-md, 8px));
-}
+  color:
+    light-dark(var(--rh-color-text-primary-on-dark, #ffffff),
+      var(--rh-color-text-primary-on-light, #151515));
```
bennypowers, line 17:
```suggestion
  color: var(--rh-color-text-primary);
```

docs/styles/color-scheme.css
bennypowers, line 1: this file should
1. live in the tokens repo
2. be generated through style dictionary build there
3. have two flavours: `:host([color-palette=whatev])` for shadow and `[color-palette=whatev]` for light (not super sure about the lightdom one...)

bennypowers (changes requested):
Pending review from the team, this is blocked on the requisite changes to tokens.

bennypowers:
This PR even improves how native controls (fieldset, input, etc) look inside color-palette elements. see the tabs demo for an example

KamiQuasi (changes requested):
`theming/customizing.md` page - Could we make the active tab on the code sample be the CSS? There is a small cognitive disconnect on that page between what is actually being customized to do the theming because it never actually says to do so with CSS changes but instead references tokens, elements, and other design system jargon that may some may not understand equate to CSS values (not that they shouldn't, but if you got dropped into that page it may cause some confusion).

docs/theming/developers.md
```diff
@@ -44,13 +44,108 @@ and elements. In other words, theming is the developer’s process of
 orchestrating design tokens with elements, particularly by way of themeable 
 container elements.
 
+## How theming works
+
```
KamiQuasi, line 48: I'd love to see something along these lines with the corresponding CSS alongside it (expanded out): 
```
As an example, here is the CSS of a theme we'll call "bordeaux". We set some initial base color scheme values for the theme, and then use those to set the appropriate design system values...
```

KamiQuasi (changes requested):
This suggestion is like the tldr; for a developer. They can dig into it more, further down the page.

bennypowers:
Thanks for the tips, @KamiQuasi. I made some changes PTAL and LMK if they do the trick.

elements/rh-tile/rh-tile.css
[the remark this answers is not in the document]
bennypowers: Will need to check the figma

elements/rh-accordion/rh-accordion-header.css
```diff
@@ -14,7 +14,9 @@
   padding: 0;
   margin: 0;
   color: var(--rh-color-text-primary);
-  background-color: var(--rh-color-surface);
```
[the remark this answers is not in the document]
bennypowers: fixed this. when a palette is set, the accordion children get the surface color, otherwise, they get lightest/darkest depending on the scheme

elements/rh-button/rh-button.css
```diff
@@ -7,6 +7,15 @@
   display: none !important;
 }
 
+[part='icon'] {
+  display: none;
+  pointer-events: none;
+}
+
+rh-icon {
+  color: currentcolor;
```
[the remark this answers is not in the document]
bennypowers: fixed

elements/rh-button/rh-button.css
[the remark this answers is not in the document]
bennypowers: fixed

elements/rh-health-index/rh-health-index.css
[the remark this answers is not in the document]
bennypowers: should be fixed

elements/rh-table/rh-table.css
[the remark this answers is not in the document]
bennypowers: i _think_ this is fixed

elements/rh-tile/rh-tile.css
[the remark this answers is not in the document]
bennypowers: we could open up palettes with 
```css
#outer.palette {
  background-color: var(--rh-tile-background-color, var(--rh-color-surface));
}
```

but i do believe the design intent is *only* lightest and darkest. cc @marionnegp

elements/rh-video-embed/rh-video-embed.css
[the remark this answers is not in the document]
bennypowers: having fixed button this should work now too

docs/theming/color-palettes.md
```diff
@@ -35,74 +35,46 @@ easier and to improve customers' digital experiences. Authors and developers who
 adopt the color palette system will produce accessible, branded experiences with
 less effort and greater cross-property consistency.
 
-## What are color palettes
+## Color palettes
```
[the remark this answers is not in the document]
bennypowers: I also think I this whole section should be rearranged 

Right now it Starts with colour palettes, then moves on to colour schemes, then theming.

It would make more sense to start with colour scheme, i.e. dark or light, because that's what most developers would recognize. After that, we can move on to explaining what a colour palette is, because that's new jargon in our design system. Then we can finish off with themeing

marionnegp (changes requested):
@bennypowers, I'm still working through all of the color context demos but wanted to drop this in before the weekend.

`<rh-button>`
- [x] I'm not seeing hover states for any of the buttons. This is affecting how focus and active states too.

`<rh-audio-player>`
- [x]  The dark theme's track color should be `--rh-color-gray-50`.
- [ ] The light theme's track color should be `--rh-color-gray-30`.
- [ ] The border around the contextual menu is missing for both light and dark themes. It also looks like it's missing a drop shadow. @coreyvickery, which box shadow token should be used here? 
Screenshot 2025-03-07 at 2 47 58 PM
- [ ] The play button's background color isn't showing for light or dark.

`rh-card`
- [ ] The default CTA's arrow doesn't show up in either light or dark cards.
Screenshot 2025-03-07 at 3 46 10 PM

`rh-code-block`
- [ ] Looks like padding is missing around the FABs.
Screenshot 2025-03-07 at 3 58 23 PM

`rh-cta`
- [ ] Focus states in light and dark themes are showing different surface and text colors than [what the docs show](https://deploy-preview-2138--red-hat-design-system.netlify.app/elements/call-to-action/style/#focus).
Screenshot 2025-03-07 at 4 03 02 PM
- [ ] Active states in light and dark aren't working.

bennypowers:
> `rh-cta` Focus states in light and dark themes are showing different surface and text 
Can you open a new issue for this?

bennypowers:
@coreyvickery can you take a look at the "Inline color palettes" section here? from chat:

> since the color-palette attribute already gives us the ability to do this (and actually doesn't do anything other than that), do you think we can rework that section in light of the rest of the page content?

marionnegp (changes requested):
rh-accordion
- [x] Is it possible for accordion's header panel to match color palette of its parent container? I'm thinking about how rh-footer's surface color doesn't match rh-accordion's in the mobile footer.

rh-footer
- [x] I can't see links in the accordions of the mobile version.
Screenshot 2025-03-10 at 10 42 00 PM

rh-health-index
- [ ] The border is missing from the lg and xl sizes in both light and dark mode.
- [x] Border colors of the highlighted letter grades don't all line up with what's in Figma.

rh-navigation-secondary
- [ ] text in logo slot shouldn't be underlined
- [ ] links in secondary nav's dropdown shouldn't be underlined

rh-pagination
- [ ] open variant's buttons should not have a background color. This is working correctly for the arrow buttons

rh-site-status
- [ ] Partial system outage: The exclamation point in the icon looks like it has some light gray pixels that are making it hard to see the exclamation point.
Screenshot 2025-03-10 at 9 36 32 PM

- How can I view the other two site statuses? (The status major demo doesn't use the right icon or text based on what I'm seeing in Figma, so I want to make sure I'm looking at the right thing.)

rh-subnav
- [ ] bottom border color on hover should be `--rh-color-gray-50`.
- [ ] focus ring should encompass the whole subnav item.
- [ ] focus state should also show the gray bottom border.

rh-table
- [ ] surface color for rh-table should change with the color palette (see [current table demo](https://ux.redhat.com/elements/table/demo/color-context/)).

rh-tabs
- [ ] box variant isn't showing all borders and tab bg colors

rh-tag
- [ ] light theme tags should be using lighter border colors.

rh-tile
- [ ] light theme tile is showing up in dark color palettes

rh-tooltip
- [ ] this is not a color issue, but the  tooltip isn't lining up with the trigger button.
Screenshot 2025-03-10 at 10 19 57 PM

rh-video-embed
- [ ] Triangle icon in play button should be white.

bennypowers:
@marionnegp I made most of those changes, but have some notes for your review:

> rh-accordion
> * Is it possible for accordion's header panel to match color palette of its parent container? I'm thinking about how rh-footer's surface color doesn't match rh-accordion's in the mobile footer.

I think what you mean here is that you always want the accordion’s headers and panels to match - meaning that if an accordion has a color palette, then all it’s headers and panels have the same color-palette and can’t have a different one. This also solves the problem with rh-footer links

> rh-site-status:
> * Partial system outage: The exclamation point in the icon looks like it has some light gray pixels that are making it hard to see the exclamation point.
> * How can I view the other two site statuses? (The status major demo doesn't use the right icon or text based on what I'm seeing in Figma, so I want to make sure I'm looking at the right thing.)


I’m not exactly sure what you mean here, I couldn’t find any demo with the words “Partial system outage” in it. What are your steps to reproduce?

> rh-subnav
> * bottom border color on hover should be --rh-color-gray-50.

If we used border tokens instead, we'd get dark mode for free

> rh-table
> * surface color for rh-table should change with the color palette (see current table demo).

Could be i'm misunderstanding but it seems like we're making an exception for table. Shouldn't table work the same as every other element? Take for example card and tile, which use `--rh-color-surface-lightest` for their background colours when rendering in light color scheme, and `--rh-color-surface-darkest` when rendering in dark color scheme; but if the user wishes, they can proactively set a color palette on a card or tile, which causes them to use the appropriate surface color for the backround.

> rh-tabs
> * box variant isn't showing all borders and tab bg colors

I don't understand exactly what the issues are, and the figma doesn't elaborate. we'll have to go over this in more detail.

> rh-tooltip
> * this is not a color issue, but the tooltip isn't lining up with the trigger button.

Unrelated to this PR, so we'll address it elsewhere

docs/theming/developers.md
```diff
@@ -44,13 +44,108 @@ and elements. In other words, theming is the developer’s process of
...
+### Color Scheme Providers
```
marionnegp, line 49: As a heads up, our headings and page titles should be using sentence case. I think this is something we'll have to check for across UX dot, but I thought they could be fixed for the Theming section while it's being worked on anyway.

docs/theming/color-palettes.md
```diff
@@ -188,6 +159,44 @@ it can also set its own color palette.
...
+}
+```
+
+Avoid customizing element CSS directly
+
+</uxdot-best-practice>
+
+</rh-card>
+
+See the [customizing](./customizing/) page for a more detailed example.
```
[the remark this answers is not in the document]
bennypowers: good catch. fixed.

docs/theming/developers.md
bennypowers: fixed

elements/rh-tabs/rh-tab.css
```diff
@@ -125,11 +125,11 @@ slot[name='icon'] {
 
   /* When button is active */
   &.active {
-    --_button-text-color: var(--rh-tabs-link-color, var(--rh-color-text-primary));
+    color: var(--rh-tabs-link-color, var(--rh-color-text-primary));
```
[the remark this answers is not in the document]
bennypowers: thanks, fixed.

marionnegp:
> rh-cta Focus states in light and dark themes are showing different surface and text
> 
> > @marionnegp Can you open a new issue for this? as for the rest they were a combination of build failing thus not updating the DP to show latest changes, and some genuine bugs. thanks for the QA

The color issue I was seeing with CTA's focus state is gone now! I was seeing  text `--rh-color-text-primary-on-light` over our darker red hover color, which made it hard to read.

uxdot/uxdot-color-scheme-picker.css
```diff
@@ -0,0 +1,82 @@
...
+
+  #button-group {
+    display: flex;
+
+    --rh-icon-size: var(--rh-size-icon-01);
+  }
+
+  label {
+    position: relative;
+    background-color: none;
```
markcaron, line 26: Not a valid property

image

markcaron: `transparent` works

markcaron (changes requested):
Can we get the borders between the `<label>` elements down to `1px` total?

Will likely require some overlapping magic.

Currently:
image

Thinner:
image

marionnegp (changes requested):
- [x] Looks like the copy buttons on the tokens pages are being highlighted with `--rh-color-interactive-primary-hover`, but can we use `--rh-color-blue-20` for now? Maybe we need to add a token like `--rh-color-interactive-highlight-hover` or something.
Screenshot 2025-03-13 at 1 46 18 PM

.changeset/floppy-deer-cry.md
```diff
@@ -0,0 +1,20 @@
+---
+"@rhds/elements": major
```
zeroedin, line 2: with this marked as a major does that mean we are going to cubone as 3.0?

.changeset/loose-rooms-stand.md
zeroedin, line 1: Nicely worded changeset 👍

docs/_includes/layouts/pages/basic.njk
```diff
@@ -24,9 +24,7 @@ layout: layouts/base.njk
 {% include 'partials/component/masthead.njk' %}
 {% include 'partials/component/sidenav.njk' %}
 
-<rh-surface id="main"
-            role="main"
-            color-palette="lightest">
+<rh-surface id="main" role="main">
```
zeroedin, line 27: We can remove the `<rh-surface>` here and from the layouts and just use `<main>`.  We aren't hard setting a color-palette in the surface so the user preference / setting the scheme switcher is enough.

docs/_includes/layouts/pages/tokens.njk
```diff
@@ -61,9 +61,7 @@ layout: layouts/base.njk
 {% include 'partials/component/masthead.njk' %}
 {% include 'partials/component/sidenav.njk' %}
 
-<rh-surface id="main"
-            role="main"
-            color-palette="lightest">
+<rh-surface id="main" role="main">
```
zeroedin, line 64: same as previous, remove rh-surface in preference of <main>

docs/elements/demo.html
```diff
@@ -36,9 +36,9 @@
     <script async src="https://ga.jspm.io/npm:es-module-shims@1.8.1/dist/es-module-shims.js"></script>
   </head>
   <body unresolved>{% if demo.filePath %}
-    <main>
+    <rh-surface role="main">
```
zeroedin, line 39: I don't think this surface is necessary here. I believe all the necessary CSS is already provided by `/styles/demo/styles.css`

docs/index.html
```diff
@@ -152,9 +162,7 @@
 {% include 'partials/component/masthead.njk' %}
 {% include 'partials/component/sidenav.njk' %}
 
-<rh-surface id="main"
-            role="main"
-            color-palette="lightest">
+<rh-surface id="main" role="main">
```
zeroedin, line 165: same as previous, remove rh-surface in preference of `<main>`

docs/styles/tokens-pages.css
zeroedin, line 1: I realize outside scope of this PR but this file should be moved under `styles/pages/tokens.css`

docs/theming/color-palettes.md
```diff
@@ -17,92 +17,90 @@ order: 2
...
+
+## Color palettes
+<a id="what-are-color-palettes"></a>
+<a id="how-color-palettes-work"></a>
+
+<abbr title="red hat design system">RHDS</abbr> defines six color palettes
+ranging from `lightest` to `darkest`, and those are the two palettes you will
+use the most for the majority of your projects.
+
+Understanding color palettes, color schemes, and theming work together is
```
zeroedin, line 74:
```Suggestion
Understanding how color palettes, color schemes, and theming work together is
```

docs/theming/customizing.md
```diff
@@ -72,15 +90,23 @@ and to both the `light` and `dark` backgrounds.
...
+To create a custom theme, designers and developers need only set the values for 
+the relevant color properties in the design system in their page's <abbr 
+  title="Cascading style sheets">CSS</abbr>. For example, let's create a custom 
+_Bordeaux_ theme. We'll write a CSS rule which selects for the `theme-bordeaux` 
+class, and in that rule we'll set the values of <abbr title="red hat design 
+  system">RHDS</abbr> theming tokens to our custom theme colors.
+In the following example, you can see how we first define some custom theme 
+colours as CSS custom properties beginning with `--bordeaux-`, then we use those 
+values in our custom theme by applying them to RHDS theming tokens, i.e. tokens 
+that look like `--rh-color-*-on-(light|dark)`.
```
zeroedin, line 102: Should we add an alert here noting the `-on-light|dark` tokens are deprecated and to use the `light-dark()` function as seen below?

zeroedin: Also see lower on the page the `Careful` alert noting the `-on-light|dark` being used to calculate the ones we are now actively overwriting with `light-dark()`

docs/theming/developers.md
```diff
@@ -61,21 +156,6 @@ a `<rh-surface>` may be used as a full-width container and provide the
...
-  * `lightest`
-  * `lighter`
-  * `light`
-  * `dark`
-  * `darker`
-  * `darkest`
-
 ### Theming whole pages
 
 To theme an entire page, you may replace the `<main>` element with an
```
zeroedin, line 161: You can do this without adding a `<rh-surface>` if you set your own `color-scheme` rule and include the CSS from `@rhds/tokens/css/color-scheme.css`.  It might win us some points for offering both solutions here.

docs/styles/styles.css
zeroedin, line 1: We should apply a `::selection {  background-color / color }` rule with a `light-dark()` switch. Dark mode selection is hard to read right now.

uxdot/uxdot-color-scheme-picker.css
bennypowers: thanks, fixed

docs/theming/customizing.md
bennypowers: Good catch. fixed.

docs/_plugins/lit-ssr/worker.ts
```diff
@@ -58,7 +73,28 @@ class RHDSSSRableRenderer extends LitElementRenderer {
...
+      }
+      yield '</style>';
+    }
+    // Render template
+
+    yield* renderValue(
+      // @ts-expect-error: if upstream can do it, so can we
+      this.element.render(),
+      renderInfo,
+    );
```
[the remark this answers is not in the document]
bennypowers: Thank you

docs/theming/developers.md
bennypowers: b1768dfc

docs/index.html
bennypowers: 81b1e9f0

docs/_includes/layouts/pages/basic.njk
bennypowers: 81b1e9f0

bennypowers:
i colocated docs graphics and inlined them where feasible - having them next to the content helps us update them to use light-dark mode colors

zeroedin:
> i colocated docs graphics and inlined them where feasible - having them next to the content helps us update them to use light-dark mode colors

Awesome I've been wanting to do that just so we can better track resources and where they are used.

docs/foundations/typography/usage/type-usage-best-practice-3-do.svg
```diff
@@ -3,6 +3,7 @@
...
         fill="#0066CC" />
   <g clip-path="url(#clip0_623_5192)">
```
bennypowers, line 9: it's kinda annoying, but we have to sanitize and namespace all the IDs, now that these are inlined

KamiQuasi (approved):
LGTM!

adamjohnson (commented):
Couple minor theme > scheme changes. Otherwise, I dig it. 👏

adamjohnson:
One more thing: I noticed that if you select the "dark" radio button toggle at the top (the middle toggle), then refresh the page or go to a new page, there's a flash of a light color scheme before it changes over to a dark color scheme. This is because JS is booting up and adding `color-scheme: dark` to the `<body>`.

This _could_ be a terrible idea, but what do ya'll think about inlining a render-blocking `<script>` tag in the `<head>` of uxdot that attempts to `localStorage.getItem('RHDS-color-scheme');` and set the color scheme on the `<body>` before the page renders? It'd probably be 5-10 lines of JS.

docs/theming/color-palettes/index.md
```diff
@@ -0,0 +1,242 @@
...
+               src="../patterns/card-child-consumers.html"
+               full-height="">
+</uxdot-pattern>
+
+<rh-card class="pullquote-card right">
+  <rh-blockquote>Color palettes lets authors write more HTML, simpler CSS, and
+    less JavaScript</rh-blockquote>
+</rh-card>
+
+To summarize: color providers can adopt one six palettes, three dark and
```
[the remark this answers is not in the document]
bennypowers: thanks, good looking out

coreyvickery:
@zeroedin One tiny theme switcher edit.

- [ ] Remove light blue border on hover

bennypowers (approved):
luminosity gradients: to market!

markcaron (approved):
Let's Get This Merged!

marionnegp (approved):
`light-dark`! Gather the masses.

PR #2158, opened 2025-02-09
fix(accordion): large size

### LINKED ISSUE #2037: [bug] `<rh-accordion>`: large attribute styles
The styles sizes guide says the following:

> There are two available sizes and the only difference is the title text size. You can use the Small size on large breakpoints, but not the Large size on small breakpoints due to the potential of long title text wrapping to more than two lines.

However it appears when applying the `large` style the title text size does not change.  Instead the only change that appears to happen is in the panel padding.

bennypowers:
Closes #2037

## What I did

1. apply large font size on non-small screens

## Testing Instructions

1. check out the [`large` demo](https://deploy-preview-2158--red-hat-design-system.netlify.app/elements/accordion/demo/large/) demo
2. compare with [basic demo](https://deploy-preview-2158--red-hat-design-system.netlify.app/elements/accordion/demo/)

## Notes to Reviewers

I kinda fudged the font sizes here. @coreyvickery @marionnegp design guidance is much appreciated here.


coreyvickery:
@bennypowers

* Text size for `large` size panel heading is `--rh-font-size-body-text-lg` and interior body text is `--rh-font-size-body-text-md`
* Text size for `small` size panel heading and interior body text is `--rh-font-size-body-text-md`

@marionnegp Do we need to document this info somewhere on ux dot?

bennypowers:
@coreyvickery PTAL i think this is what you meant

elements/rh-accordion/rh-accordion.css
```diff
@@ -5,6 +5,16 @@
   &.expanded {
     box-shadow: var(--rh-box-shadow-sm, 0 2px 4px 0 rgba(21, 21, 21, 0.2));
   }
+
+  &.large {
+    --_header-font-size: var(--rh-font-size-body-text-md, 1rem);
+    --_panel-font-size: var(--rh-font-size-body-text-md, 1rem);
+
+    /* --rh-media-sm */
+    @media (min-width: 768px) {
```
zeroedin, line 14: I wonder if this should  be a @container query with the component itself the container. Accordions aren't always full-width to the document so the container query would allow accordions placed in a smaller fraction of a page to adjust their size to properly based on their own width rather than the page, but when they are full-width still adjust appropriately.

marionnegp (changes requested):
I'm still seeing the `large` accordion's heading at `--rh-font-size-body-text-md`.

bennypowers:
I think the container query broke it

zeroedin:
> I'm still seeing the `large` accordion's heading at `--rh-font-size-body-text-md`.

@marionnegp this should be fixed now for you to review.

PR #2171, opened 2025-02-17
feat(navigation-primary): add navigation primary

### LINKED ISSUE #375: [feat] `<rh-navigation>` ready to migrate
### Tag Name

rh-navigation

### RHDS Only

- [X] The component will be **RHDS only**, delete the PFE component once published

### Todos

- [ ] On what we should do with pfe-navigation right meow.

### LINKED ISSUE #2268: [bug] rh-icon microns should be a range
`<rh-icon set="microns">` should use a range between 8px-12px by specification.  However it was hardcode to always be 12px.

```
.microns svg {
    width: 12px;
}
```

Suggested fix:

```
.microns svg {
  min-width: 8px;
  max-width: 12px;
  width: var(--rh-icon-size, 12px);
}
```

Reference missing token values for this range as well https://github.com/RedHat-UX/red-hat-design-tokens/issues/163

zeroedin:
## What I did

1.  Added `<rh-navigation-primary>`

Closes #375 
Closes #2268 

## TODO

- [x] Tests
- [x] Docs / JSDoc
- [x] Changeset
- [x] Reverse the idea of `standalone` instead,  a `group` option? Putting it on the hamburger items instead. I feel the idea of a standalone link in the slots="links" is a bit confusing as standalone has more to do with how the dropdowns work at different breakpoints and specifically inside the hamburger or not.
- [x] Add overflow scroll when details open content panels are taller than viewport (mobile or squashed vertically browser)
- [ ] Better code examples of actual panel content

## Testing Instructions

1. [`<rh-navigation-primary>`](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/)  demo
2. [Color Context](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/color-context/)  demo
3. [Not Defined](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/not-defined/)  demo
4.  [Right to Left](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/right-to-left/)  demo
5. [Translation](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/translation/)  demo
6. [Link as top-level](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/links-as-top-level/)  demo

## Notes to Reviewers

Review the code and look for anything glaringly out of sorts.  Is the documentation good enough?  Design review for any spacing issues. Look for anything off with colors in context demos.  Accessibility review: How does this sound in screen readers? We still have the issue with a single list item for the event slot. Maybe a fix for that can be followed up with a patch release.

There are some non standard icon sizes used reference https://github.com/RedHat-UX/red-hat-design-tokens/issues/163


bennypowers:
Thanks for pushing this branch, @zeroedin. I'm interested in learning more about the issues you had with rh-disclosure, because it would be good to recycle that as much as possible. Perhaps there's room for `rh-menu`, `rh-dropdown`, etc, (implementing the required aria patterns) with a common controller between them?

zeroedin:
> Thanks for pushing this branch, @zeroedin. I'm interested in learning more about the issues you had with rh-disclosure, because it would be good to recycle that as much as possible. Perhaps there's room for `rh-menu`, `rh-dropdown`, etc, (implementing the required aria patterns) with a common controller between them?

I felt like the changes that were to be requested were beyond the scope of what a simple disclosure would do, and the styles needed were also quite different. I also needed an element that handled dropdowns and slotted links, so I settled on `rh-navigation-item` rather than `rh-navgiation-dropdown` as not all elements in the menu are disclosure/dropdown but shared styles between. This does not mean we can't take a much deeper look at this and look at how, if possible, to return to that idea and try again or apply a controller as you suggested.

elements/rh-navigation-item/rh-navigation-item.ts
```diff
@@ -0,0 +1,157 @@
...
+    const [dropdownMenu] = this.#slots.getSlotted<HTMLElement>('');
+    this.#highlight = dropdownMenu.querySelector('[aria-current="page"]') ? true : false;
+    this.requestUpdate();
+  }
+
+  public close() {
+    this._details.open = false;
+  }
+
+  public open() {
```
zeroedin, line 148: Should there be a declarative way to open a navigation item? 

I want to lean towards no as an answer, and this may be where this details summary implementation splits from an `rh-disclosure,` which can be set to `open` on page load by default.

I'm unsure of a use case where you would want the navigation to auto-load with a dropdown open declaratively. I think that might confuse a user when they do not trigger the interaction, as well it would "hide" content below it, unlike a disclosure, which pushes it down. 

I can see this behavior for jump links nav or an accordion/disclosure where you would potentially follow a #hash link to that section of the page, and you would want the disclosure open, but hiding page content with an open nav on load might be where my opinion splits.

bennypowers:
Because this is more of a site-system-pattern than it is a design system element (i.e. it's not like rh-cta, it's more like rh-footer), our number of users is limited. Therefore we should, on release, commit to not changing the shadow DOM content for some period of time, say 2 major versions. This would require designers to understand the technical limitations on changing primary navigation, and it would also require implementing users to stay up to date if we do make any shadow DOM changes (e.g. in case of showstopping accessibility bugs).

elements/rh-navigation-primary/demo/rh-navigation-primary.html
```diff
@@ -0,0 +1,273 @@
+<rh-navigation-primary role="navigation">
+  <rh-navigation-item variant="dropdown">
+    <span slot="summary">AI</span>
```
adamjohnson, line 3: On mobile + tablet viewports where the disclosures are stacked like so:

Main nav dropdown items stacked in the open menu on mobile viewports

If you tab into an expanded details/summary (eg: "Link 1" in "AI Content") and hit Escape, shouldn't it just close the currently open details/summary? Not the entire open mobile site menu?

FWIW rh-disclosure will only close the currently opened, nested disclosure when hitting escape. Not the whole enchilada. See the [nested demo](https://deploy-preview-2043--red-hat-design-system.netlify.app/elements/disclosure/demo/nested-disclosures/) there.

adamjohnson (changes requested):
Some quick a11y review on a Friday afternoon. Impressive work here!

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,426 @@
...
+          </div>
+          <details id="hamburger" @toggle="${this.#hamburgerToggle}">
+            <summary>
+              <rh-icon icon="menu-bars" set="ui"></rh-icon>
+              <slot name="summary" id="summary">Menu</slot>
+              <rh-icon icon="caret-down" set="microns" aria-labelledby="summary"></rh-icon>
+            </summary>
+            <div id="details-content" role="list">
+              <slot></slot>
+            </div>
```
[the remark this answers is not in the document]
zeroedin: WHAT?  oh boy... that'll be interesting

elements/rh-navigation-primary/demo/rh-navigation-primary.html
zeroedin: Good catch. For compact viewports, yes, it shouldn't close the hamburger. I think I just need to add the skip option which doesn't close the hamburger on the close method.  

```
case 'Escape': {
   this.close(true);
   break;
}
```

elements/rh-navigation-primary/demo/rh-navigation-primary.html
```diff
@@ -0,0 +1,273 @@
...
+          </rh-avatar>
+          <div>Login: jdoe@redhat.com</div>
+          <div>Account number: 0123456</div>
+          <div>jdoe+alias@redhat.com</div>
+          <div>
+            <select>
+              <option>English</option>
+              <option>French</option>
+              <option>German</option>
+            </select>
```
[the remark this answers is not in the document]
zeroedin: Yeah the content here was just added ad hoc to see what the nav panels looked like with content.  Page owners will own that slotted content and its styles BYOHTML/CSS style.  We should definitely clean them up before launch to production as more complete examples.

elements/rh-navigation-primary/demo/rh-navigation-primary.html
```diff
@@ -0,0 +1,273 @@
...
+    </a>
+  </rh-navigation-item>
+
+  <rh-navigation-item slot="links" standalone><a href="#docs">Developers</a></rh-navigation-item>
+  <rh-navigation-item slot="links" standalone><a href="#support">Docs</a></rh-navigation-item>
+  <rh-navigation-item slot="links" standalone><a href="#console">Support</a></rh-navigation-item>
+
+  <rh-navigation-item slot="dropdowns" hide="sm" variant="dropdown" standalone>
+    <rh-icon slot="icon" icon="search" set="ui"></rh-icon>
+    <span slot="summary">Search</span>
```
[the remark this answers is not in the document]
zeroedin: Nice, will fix.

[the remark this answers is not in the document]
zeroedin: visually hidden instead of display none, while the icons and text are slotted I was display none the slot which isn't ideal as even if you added a label to the icon you end up with a double announcement, so now I just target the slotted summary text and visually hide it instead so its still used as accessible label text for the summary.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,426 @@
...
+            <slot name="logo">
+              <a href="/">
+                <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 613 145" height="30" width="125">
+                  <defs>
+                    <style>
+                      .cls-1{fill:var(--rh-color-brand-red, #ee0000);}
+                      .cls-2{fill:var(--rh-color-text-primary, #151515);}
+                    </style>
+                  </defs>
+                  <title>RedHat-Logo</title>
```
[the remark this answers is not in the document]
zeroedin: Interesting, wonder where that copy pasta came from.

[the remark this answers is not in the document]
zeroedin: Might come out of Figma like that now that I think about it.  Likely uses the name of the layer.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,426 @@
...
+            && this.#openSecondaryDropdowns.size === 0) {
+          this.#closeOverlay();
+        }
+      }
+    }
+  }
+
+  #onKeydown(event: KeyboardEvent) {
+    switch (event.key) {
+      case 'Escape': {
```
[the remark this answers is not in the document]
zeroedin: Will likely have to solve this exactly how we solved for nested disclosures.

elements/rh-navigation-item/README.md
bennypowers, line 1: needs some prose

.changeset/perfect-llamas-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-avatar>` improved layout compatibility for plain variant avatars
```
bennypowers, line 5:
```suggestion
`<rh-avatar>`: improved layout compatibility for plain variant avatars
```

elements/rh-navigation-primary/README.md
bennypowers, line 1: needs some more prose here

bennypowers (changes requested):
some initial impressions

elements/rh-navigation-primary/rh-navigation-primary.ts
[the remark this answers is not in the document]
adamjohnson: Fixed by [7701230](https://github.com/RedHat-UX/red-hat-design-system/pull/2171/commits/7701230968da9752566bc9d416f576aed348b085). 👏

elements/rh-navigation-primary/demo/rh-navigation-primary.html
adamjohnson: Probably still some work to do on this. Couple things:

(tested in Safari on mobile viewports)

1. Tab to open the mobile menu. Menu is open, hit escape. Backdrop dissapears, but menu stays open.
2. Open the mobile menu, then open a submenu (eg: "Learn"). Hit esacape. "Learn" closes. Hit escape again, main menu doesn't close.
3. Hitting escape in a submenu closes the backdrop. Backdrop should stay open if the mobile menu is open and disappear when everything is closed.

zeroedin: @adamjohnson Give this another run and let me know if I need to tweak/missed anything I believe I've now covered all bases. 🤞

zeroedin: Please take a look again, I think I have it squared away now.

elements/rh-navigation-primary/README.md
zeroedin: I'm leaving this open, I did some minor additions, but it still needs more.

elements/rh-navigation-primary/demo/rh-navigation-primary.html
adamjohnson: Nailed it. Nice job. 🚀

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,449 @@
...
+
+/**
+ * Navigation Primary
+ * @slot - Place element content here
+ */
+@customElement('rh-navigation-primary')
+export class RhNavigationPrimary extends LitElement {
+  static readonly styles = [styles];
+
+  #internals = InternalsController.of(this, { role: 'navigation' });
```
adamjohnson, line 56: This is being semantically marked as a `<nav>`, while the next element inside the shadowdom is also a `<nav>` element. I'd probably remove this and go with the `<nav id="container">` so there's just one `<nav>` element.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,449 @@
...
+                </svg>
+              </a>
+            </slot>
+          </div>
+          <details id="hamburger" @toggle="${this.#hamburgerToggle}">
+            <summary @click="${this.#hamburgerClick}">
+              <rh-icon icon="menu-bars" set="ui"></rh-icon>
+              <slot name="summary" id="summary">Menu</slot>
+              <rh-icon icon="caret-down" set="microns" aria-labelledby="summary"></rh-icon>
```
adamjohnson, line 170: From 992px to 1199px, the accessible label, "Menu", is repeated 2x to assistive technology. This is due to the aria-labelledby on the caret-down.

Have to re-work this label to jive with the "Menu" word visibility.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,449 @@
...
+    const rtl = this.#dir.dir === 'rtl';
+    const { on = 'light' } = this;
+    const classes = { compact: this.compact, rtl, on: true, [on]: !!on };
+    return html`
+      <nav id="container" class="${classMap(classes)}">
+        <div id="bar">
+          <div id="logo">
+            <slot name="logo">
+              <a href="/">
+                <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 613 145" height="30" width="125">
```
adamjohnson, line 151: Should this get minified?

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,449 @@
...
+              <rh-icon icon="menu-bars" set="ui"></rh-icon>
+              <slot name="summary" id="summary">Menu</slot>
+              <rh-icon icon="caret-down" set="microns" aria-labelledby="summary"></rh-icon>
+            </summary>
+            <div id="details-content" role="list">
+              <slot></slot>
+            </div>
+          </details>
+          <div id="secondary">
+            <div id="event" role="list"><slot name="event"></slot></div>
```
adamjohnson, line 177: re: `role="list"`:

Can there be multiple events?

Right now there's only one event that is being inserted into a list. If there can be more than one event, this can stay.

Otherwise, perhaps `role="list"` should be removed.

adamjohnson: Also, is there a benefit to using `role="list"` for these versus just `<ol>` or `<ul>`?

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,449 @@
...
+              </a>
+            </slot>
+          </div>
+          <details id="hamburger" @toggle="${this.#hamburgerToggle}">
+            <summary @click="${this.#hamburgerClick}">
+              <rh-icon icon="menu-bars" set="ui"></rh-icon>
+              <slot name="summary" id="summary">Menu</slot>
+              <rh-icon icon="caret-down" set="microns" aria-labelledby="summary"></rh-icon>
+            </summary>
+            <div id="details-content" role="list">
```
adamjohnson, line 172: Looks like aXe is complaining about these `role="list"` attributes:

> Certain ARIA roles must contain particular children.

[aXe docs](https://dequeuniversity.com/rules/axe/4.10/aria-required-children?application=AxeEdge) for this error.

elements/rh-navigation-primary/rh-navigation-primary.ts
adamjohnson: Also, I'd recommend adding `aria-label="Main navigation"` to whichever `<nav>` element you decide to keep. There will likely be several `<nav>` elements on a page and labeling them individually will help AT.

zeroedin: I need to remove the internal `<nav>` that should just be a div. What we want to do is shift the [role] from a not-defined state (attr on the component in lightdom) then remove that apply the role via internals on hydration.  This is what we do in the secondary nav as well.

The available accessible-label property takes care of that second comment.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: AFAIK we don't do any minification to our HTML on build.  I believe we attempted that but ran into reasons why not to.   I'm not sure minifying the coupleof  lines of SVG here provides any value.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: I wonder if this is a false positive given element internals is responsible for the `rh-navigation-primary-item` role of list-item which is what is complaining about isn't available.   Question, did you test in the DP which is SSR'd?  I'd be curious about the results in local dev environment `:8000` when its not DSD'd.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: /cc @bennypowers

elements/rh-navigation-primary/rh-navigation-primary-item.ts
```diff
@@ -0,0 +1,135 @@
...
+  @state()
+  private _open = false;
+
+  @property() summary?: string;
+
+  @property() variant?: 'link' | 'dropdown' = 'link';
+
+  @property({ reflect: true }) hide?: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl' = undefined;
+
+  @property({ reflect: true, type: Boolean }) standalone = false;
```
bennypowers, line 58: make this a private context thing, instead, so that if it's in the context of a hamburger menu or whatever, then that triggers the internal changes

hellogreg:
This all is really looking great! Here are some quick thoughts after Mac testing. I'm checking Windows this afternoon, too.

### Some things I love

- Panels collapsing when focus moves past the last item in them (exception mentioned below)
- The Devs/Docs/Support link underlines on focus/hover.
- The matching pill-shaped focus indicators. (All the the focus indicators look nice.)

### General notes
_Some of these may be by design, but thought I'd mention anyway._

- "For You" should be sentence case.
- Pressing shift-tab when on the first item within an expanded panel returns focus back to the toggle button (good!) but collapses the panel (is that by design?).
- Expanded summary widens to panel width when open. Is that also by design?
- On narrower windows (with the hamburger menu), if you expand the menu and then tab past all its items (to the Summit link), the menu panel remains expanded. This is unlike the other disclosures, when the page is wider.

### Safari

- Summary icon doesn't change from collapsed to expanded when top level buttons (e.g., "AI") are toggled.
collapsed summary icon on expanded summary in safari

hellogreg:
> I'm checking Windows this afternoon, too.

No surprises in initial tests with Windows JAWS (Chrome), NVDA (Chrome and Firefox), or Narrator (Edge).

Oh, one thing I forgot to mention in the previous message: every screen reader says the Summit link is in a list, albeit one with only one item.

hellogreg:
On iOS (didn't check Android yet), if the menu overflows beyond the bottom of the screen and there isn't enough page content beneath it, the page didn't scroll to show the full contents of the menu. [Video to demonstrate.](https://drive.google.com/file/d/1_EtU56JXTOcTLyEb5d-Gt7pA9_8cu2Q3/view?usp=drive_link)

zeroedin:
> On iOS (didn't check Android yet), if the menu overflows beyond the bottom of the screen and there isn't enough page content beneath it, the page didn't scroll to show the full contents of the menu. [Video to demonstrate.](https://drive.google.com/file/d/1_EtU56JXTOcTLyEb5d-Gt7pA9_8cu2Q3/view?usp=drive_link)
> 
> Edit: This was _not_ an issue on my Android test phone.

Yup this is still in my todo list in the PR description. But hey glad you caught it as well cause I almost forgot about it and just put it in the todo the other day to remind myself its not complete there.

elements/rh-navigation-primary/rh-navigation-primary-item.ts
zeroedin: Let's chat about this during office hours. I made an attempt to consume the context and then re-provide it with a modified value, but it ended up being more difficult than anticipated.

Psuedo code representative of my approach:
```
<rh-nav>
  # defaults hamburger false
  # context { compact: true, hamburger: false }
  <rh-nav-hamburger>
    # sets hamburger: true
    # context { compact: true, hamburger: true } because its ancestor provider is the rh-nav-hamburger
    <rh-nav-item>
    <rh-nav-item>
    <rh-nav-item>
    <rh-nav-item>
  </rh-nav-hamburger>
  # context { compact: true, hamburger: false}  because its ancestor provider is the rh-nav
  <rh-nav-item>
  <rh-nav-item>
  <rh-nav-item>
</rh-nav>
```

I referenced this: https://lit.dev/playground/#sample=examples/context-consume-provide

This works with the state being hydrated on the client but doesn't work in SSR. Unfortunately, that's where I'd prefer this state to be available.

marionnegp (changes requested):
I'm going to do a more thorough review, but I wanted to drop these here before I forgot.

For desktop: 
- [ ] inline padding for nav items in `#hamburger` should be `--rh-space-lg`
- [ ] Spacing between the logo and the nav items should be `--rh-space-2xl` for screens  >992px
- [ ] I think I'm seeing the mobile styles when the nav items are expanded
Screenshot 2025-03-21 at 4 17 40 PM
- For the vertical dividing lines between the `links` slots, how annoying would it be to use `gray-60` instead of the `border-subtle` color for dark mode?

elements/rh-navigation-primary/rh-navigation-primary-item.ts
bennypowers: don't put the context on the nav, keep it only on the hambuger


```html
<rh-nav>
  # defaults hamburger false
  # @provider({ context: rhNavContext }) // { compact: true }
  <rh-nav-item>
    # @consumer({ context: rhNavContext })
  <rh-nav-hamburger>
    # @provider({ context: rhNavHamburgerContext })
    # true (or some other truthy value, maybe the hamburger state
    <rh-nav-item>
      # @consumer({ context: rhNavContext })
      # @consumer({ context: rhNavHamburgerContext })
  </rh-nav-hamburger>
</rh-nav>
```

elements/rh-navigation-primary/rh-navigation-primary-item.css
```diff
@@ -0,0 +1,460 @@
...
+      border-end-end-radius: calc(var(--rh-border-radius-default, 3px) * 2);
+      border-end-start-radius: calc(var(--rh-border-radius-default, 3px) * 2);
+    }
+  }
+
+  &.standalone {
+    details {
+      display: contents;
+
+      & > * {
```
bennypowers, line 179: this and the following rules are violating no-descending specificity.

Nesting is great but in this file it's working against readability, please rework this sheet so that it reads top-to-bottom

elements/rh-navigation-primary/rh-navigation-primary-item.ts
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * Color palette
+   */
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
+
+  render() {
+    const { hide = '', variant = '', standalone } = this;
+    const { compact = true } = this.ctx ?? {};
+    return html`
+      <div id="container" part="container" class="${classMap({
```
bennypowers, line 82:
```suggestion
      <div id="container" class="${classMap({
```

Do we need this part? I think `part="container"` is an antipattern - it exposes implementation details

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,442 @@
...
+      this.#upgradeAccessibility();
+      this.#internals.ariaLabel = this.accessibleLabel;
+    }
+  }
+
+  render() {
+    const rtl = this.#dir.dir === 'rtl';
+    const classes = { compact: this.compact, rtl };
+    return html`
+      <div id="container" class="${classMap(classes)}">
```
bennypowers, line 139:
```suggestion
    return html`
      <div id="container" class="${classMap({ compact: this.compact, rtl })}">
```

bennypowers (changes requested):

elements/rh-navigation-primary/rh-navigation-primary-item.ts
zeroedin: There is a usage in lightdom. There was an issue trying to set the padding-inline-end given a complex :nth-last-child rule. Could likely use a CSS  var instead of opening up all the things on the container. Styling the first and last items of slotted items are always a PITA internal to the component.

elements/rh-navigation-primary/rh-navigation-primary-item.ts
zeroedin: When I attempted that, I got a conflict at the `context`  level. IIRC was a typescript issue. I couldn't find a good example of an element consuming two different contexts, so took the approach of trying to modify the context along the way. I'll give this another shot and add the error here if I run into it again.

zeroedin: Figured the typescript error out: 

For both the provide and consume decorator I need to pass the imported context as the type not as the value.
```
  @provide({ context: hamburgerContext }) private ctx = this.#makeContext();

  @consume({ context: hamburgerContext, subscribe: true })
  @property({ attribute: false })
  private hamburgerCtx?: RhNavigationPrimaryHamburgerContext;
```

elements/rh-navigation-primary/rh-navigation-primary-item.css
zeroedin: I agree with you this file needs to be revisited; with the number of changes to the design, it has gotten unwieldy in a lot of ways.

docs/_data/repoStatus.ts
```diff
@@ -36,6 +36,8 @@ export interface RepoStatusRecord {
   name: string;
   /** Appears on the header of the docs page */
   description?: string;
+  /** Appears above the header on the docs page */
+  preamble?: string;
```
bennypowers, line 40: let's just put this in the docs file itself, rather than add config here

elements/rh-avatar/rh-avatar.css
```diff
@@ -50,7 +50,7 @@
 }
 
 :host([layout='block']) #container,
-#container.mobile {
```
bennypowers, line 53: is this related to nav1?

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,470 @@
...
+} from './rh-navigation-primary-item.js';
+import './rh-navigation-primary-overlay.js';
+
+import '@rhds/elements/rh-icon/rh-icon.js';
+import '@rhds/elements/rh-surface/rh-surface.js';
+
+import styles from './rh-navigation-primary.css';
+
+
+export class RhNavigationPrimaryHamburgerClickEvent extends Event {
```
bennypowers: don't think we need this, or the click listener

bennypowers (commented):
removed custom events

elements/rh-navigation-primary/rh-navigation-primary-item.ts
```diff
@@ -0,0 +1,130 @@
...
+import { themable } from '@rhds/elements/lib/themable.js';
+
+import '@rhds/elements/rh-icon/rh-icon.js';
+
+import './rh-navigation-primary-item-menu.js';
+
+import styles from './rh-navigation-primary-item.css';
+
+@customElement('rh-navigation-primary-item')
+@colorPalettes
```
bennypowers, line 26: is it necessary for this to have a color-palette attr?

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -0,0 +1,457 @@
...
+                </svg>
+              </a>
+            </slot>
+          </div>
+          <details id="hamburger" @toggle="${this.#hamburgerToggle}">
+            <summary>
+              <rh-icon icon="menu-bars" set="ui"></rh-icon>
+              <div id="summary"><slot name="summary">Menu</slot></div>
```
bennypowers, line 167:
```suggestion
              <div id="summary"><slot name="mobile-toggle-label">Menu</slot></div>
```

bennypowers: let's call this by what it does rather that what it is in the shadow root right now

docs/_data/repoStatus.ts
markcaron: @bennypowers the goal was to get it above the `Overview` which is added automatically.

elements/rh-avatar/rh-avatar.css
zeroedin: It's a fix for avatar that is needed in nav. I can move it to its own PR if that is helpful.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: I was thinking about doing this as well. While it is correct for the element, it doesn't feel right for the implementor.

elements/rh-navigation-primary/rh-navigation-primary-item.ts
zeroedin: Maybe not anymore.  Might be an orphan now. I will investigate.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: Also, I wonder if it couldn't just be a property on the component.  Not sure there will be a need to change anything that would need a slotted level of concern.  🤔

elements/rh-navigation-primary/rh-navigation-primary-item.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  @consume({ context, subscribe: true })
+  @state()
+  private compact?: boolean;
+
+
+  /**
+   * Color palette
+   */
+  @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
```
bennypowers, line 65: keep themable, but remove this

zeroedin: Ok, yeah, that makes sense, given this is a lightdom element. A primary nav item may be the first to instantiate on the page and provide the theme styles.  Good catch.

elements/rh-navigation-primary/demo/rh-navigation-primary.html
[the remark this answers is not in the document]
adamjohnson: Heads up: this is solved on the main primary nav demo, but `<label>`'s are needed on the other demos (like not defined, translation, etc).

elements/rh-navigation-primary/rh-navigation-primary.ts
adamjohnson: Looks to be happening in both the DP and `:8000`. On `:8000` there are actually more occurances.

I agree with you though, this appears to be a false positive based on the a11y tree and output from VO. It's a bummer because this axe error is likely to be flagged by other devs even though it appears to be a false positive.

zeroedin:
> Should this hamburger icon be white on dark schemes?

Yes, based on the figma file.  Are you seeing it as something else.  As far as I can tell it is set to `color: var(--rh-color-text-primary)` which would be white on dark scheme. 

> If you go to the Not Defined demo and disable JS, the "My Red Hat" icon disappears. 👇

`rh-avatar` is not SSR'd in 11ty. So, while the primary nav javascript isn't loaded on that page, turning off JS globally stops avatar JS from loading. Avatar here is a slotted element. We could also slot a rh-icon for unauthed experiences as well.  

Should we SSR avatar? Maybe... That said, this will likely be implemented as an integration with Primer "account dropdown" or some other way in a Drupal template / authed experience that won't be equal to what we see here in the demo directly.

elements/rh-navigation-primary/rh-navigation-primary.ts
adamjohnson: So, @hellogreg and I took another look at this today and found something interesting.

If we add `role="listitem"` to each `<rh-navigation-primary-item>` element, the axe error disappears. 

I realize we're already setting `role="listitem"` via the Internals Controller. Unfortunately, that doesn't appease axe. 

What do you think about adding `role="listitem"` to these elements @zeroedin?

<hr>

You can test this in your console via:

```js
const getPrimaryItem = document.querySelectorAll('rh-navigation-primary-item');
for (item of getPrimaryItem) {
    item.setAttribute('role', 'listitem');
}
```

markcaron (commented):
@zeroedin I noticed that the caret doesn't flip in Safari in [the demo](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/). Not sure if that's a known issue or not.

image

hellogreg:
Doing screen reader tests today. It's playing well with all Mac browsers and VoiceOver. Same for iOS Safari and VoiceOver.

Next up: Linux Firefox/Orca.

hellogreg:
Noticed something with keyboard nav in all browsers. If you...

1. Put focus on the search button.
2. Press space/enter to expand it.
3. Tab to text input.
4. Tab again to the Submit button.
5. Press shift tab once...

Focus goes all the way back to the original search button (not the text input) and the panel is collapsed. Apologies if this is something tracked above and I missed it!

[Video clip](https://drive.google.com/file/d/1tOycSGp2Ns6B5wJkVlOur8h-EJDV9esc/view?usp=sharing)

hellogreg:
Windows screen reader checks complete. No surpirses with NVDA (with Chrome/Firefox), JAWS (Chrome), or Narrator (Edge).

Android is all that's left.

hellogreg:
Android Talkback worked well, too, except that the expanded panels remain expanded even when focus moves past them. In the following shots, focus has been moved to My Red Hat and then to the "Link outside" link, but the nav panel remains expanded.

Android shot 1


Android shot 2

elements/rh-navigation-primary/rh-navigation-primary-item-menu.ts
```diff
@@ -0,0 +1,53 @@
...
+      }
+      const nav = root.host.closest('rh-navigation-primary');
+      this.compact = nav ? nav.offsetWidth < 1200 : true; /* failover to true */
+    }
+  }
+
+  render() {
+    const { compact = true } = this;
+    return html`
+      <div id="container" class="${classMap({ compact, dehydrated: !this.#hydrated })}">
```
adamjohnson, line 42: We talked about this earlier but wanted to mark it down here: looks like on the [base demo](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/) on desktop, the dropdowns are expanding full width instead of being constrained to `max-width: 1440px;`.

Screenshot of main navigation with an expanded dropdown without a max width on the dropdown contents

Looks like the `.compact` class logic needs some love.

zeroedin: @adamjohnson pushed code earlier to resolve, can you take a peak make sure you are seeing it on your end and resolve this issue as a confirmation?

zeroedin:
> @zeroedin I noticed that the caret doesn't flip in Safari in [the demo](https://deploy-preview-2171--red-hat-design-system.netlify.app/elements/navigation-primary/demo/). Not sure if that's a known issue or not.
> 
> image

this has been resolved.

elements/rh-navigation-primary/rh-navigation-primary-item-menu.ts
adamjohnson: Can confirm this is resolved. Nice.

marionnegp (changes requested):
@zeroedin, still checking through some of the other parts of the nav, but I wanted to drop this in here to avoid holding up the PR. It's looking really good. I think some of the things I found were due to design changes happening as you were building.

**`#bar` & `#logo`**
- [ ] Dark mode box-shadow should be `box-shadow: 0px 4px 6px 1px rgba(0, 0, 0, 0.60);` (The color and opacity are already correct.)
- [x] Nav bar height is 78px in the demo but 83px in mockup. I'm not sure what's causing the difference
- [x] The logo's `margin-inline-start` should stay at `--rh-space-2xl` until the 576px breakpoint. `#secondary`'s `margin-inline-end` should stay at `--rh-space-2xl` until the same breakpoint.
- [x] Because of padding on the `AI` nav item, there's 40px of space between the RH logo and `AI`, but the mockup is showing 32px. I think `#logo` just needs its `margin-inline-end` to change to `--rh-space-lg`.

**Developers | Docs | Support (`links` slot)**
- [x] Developers, Docs, and Support should use `--rh-font-size-body-text-sm`.
- [ ] The underline on hover should be dashed.
- [ ] Underline color should be `gray-50` in light mode and `gray-40` for dark mode. (Will likely need to figure out a better token for this one.)
- [ ] Can the focus ring (for both focus and active states) be offset from link by 2px to look more like this: 
Screenshot 2025-04-03 at 9 02 35 PM 
(This is to align with some focus indicator guidelines that's being worked on.)
- [ ] For both light and dark mode, the background color shouldn't change in the focus and active states.
- [ ] The space between the link and the divider bar should be `--rh-space-lg`. 
- [ ] Divider color should be `gray-20` in light mode and `gray-60` in dark mode. (I know these don't align with our current border tokens. :-/ )
- [ ] The focus ring should appear in the active state too.
- [ ] This doesn't matter for the web component, but `Developers` should be `Console`.
- [ ] I think these links should disappear at 1440px. They currently disappears at 1439px.

**Summit logo (`event` slot)**
- [ ] For both light and dark mode, the background color shouldn't change in the focus and active states.
- [ ] Can the focus ring (for both focus and active states) be offset from link by 2px to look more like this:
Screenshot 2025-04-03 at 9 42 14 PM
- [ ] The focus ring should appear in the active state too.
- [ ] I think the logo's supposed to be visible until the screen hits 576px.

**`#hamburger`**
- [ ] At the 320px breakpoint, there should be 4px (`--rh-space-xs`) between the hamburger button and the account button.

**Search, For you, and account dropdowns (`dropdowns` slot)**
- [ ] Should use `--rh-font-size-body-text-sm`.
- [ ] Caret icon should be `gray-40` in light mode. (This is different from our subtle border color token. I can't remember why it was chosen.)
- [ ] Caret icon size should be 10px.
- [ ] On hover and focus, the border width should be 1px. (Should change to 2px only for active/`open` state.)

marionnegp (changes requested):
**dropdowns in `#hamburger`**
- [x] Caret icon size should be 10px. (By the time I submit this, this edit and the next one might have been fixed...)
- [x] Caret icon should be `gray-40` in light mode.
- [ ] On mobile, `menu bars` icon size should be 18px and the color should be `--rh-color-icon-secondary`.
- [ ] On mobile, "Menu" text should use `--rh-font-size-body-text-sm`
- [ ] Can the hamburger menu have the same hover, active, and focus styles as the similar-looking buttons in the `dropdowns` slot? The active/open state just needs to use `--rh-color-surface-lightest` in light mode and `surface-darker` in dark mode. @KrystalMoore16, I don't think this was specced out in the Figma file.
- Waiting to hear back about whether we can use the hat only logo at very small screen sizes.

markcaron (approved):
@zeroedin Safari caret issue looks resolved! Fantastic work!

docs/_includes/layouts/pages/element.11ty.ts
```diff
@@ -167,7 +167,8 @@ export default class ElementsPage extends Renderer<Context> {
 
   async #renderOverviewPage(content: string, ctx: Context) {
     const description = ctx.doc.docsPage.description ?? ctx.doc.description ?? '';
-    return html`${!ctx.doc.planned ? '' : html`
```
bennypowers, line 170: instead of editing this file, please make the preamble markdown content in the element docs, and remove the `preamble` flat from `ElementDocsPageBasicData`, as well as the preamble text from `repoStatus`.

That content can be a persistent toasted alert, or an inline alert, on the overview page.

elements/rh-avatar/rh-avatar.css
bennypowers: 's fine

elements/rh-navigation-primary/rh-navigation-primary.ts
bennypowers: just add `role = 'listitem'` as an undecorated class field in rh-navigation-primary-item.ts and let's be done with it

docs/_data/repoStatus.ts
bennypowers: I'd rather not maintain more members in repoStatus, and I certainly don't want to have to trace the threads to find docs content in repoStatus.ts

Can this be a persistent alert, or an inline alert after the heading? that's attention grabbing, consistent, and doesn't require unrelated changes to our docs framework.

bennypowers (changes requested):

docs/_includes/layouts/pages/element.11ty.ts
bennypowers: et voila: [491f78b](https://github.com/RedHat-UX/red-hat-design-system/pull/2171/commits/491f78b770555c3db1501f458b23f307d405c14f)

image

bennypowers (approved):

marionnegp (changes requested):
Checked out the last pieces of the nav and found just a couple things:

**dropdown menu**
- [x] Based on the mockup, I don't think the dropdown menus in `#hamburger` need the rounded bottom corners. Looks a little funky at large screen sizes.

**Other demos**
- [x] It looks like a link as top level item in the `#hamburger` section should turn blue, show the dashed underline, and have no bottom border.

bennypowers:
## drop-shadow
To toggle the drop shadow size depending on color scheme, we'll use a style query on chrome, and fall back to the device preference if style queries are unsupported. we'll do this in a follow up patch

marionnegp (changes requested):
Last thing! I forgot to check the focus and active states for the link as top level item. Can the bottom gradient border be removed for both those states?

marionnegp (approved):
Woohoo! Awesome work.

PR #2182, opened 2025-02-24
feat(footer): href attribute for social links

bennypowers:
## What I did

1. allow users to specify a social link in the footer by adding an href attribute to the social link custom element, instead of slotted in a link element.
2. Analytics attributes on the host _should_ still work, so long as the latest click-to-eddl is loaded 🤞 
3. users must also include an `accessible-label` attribute
4. existing markup will still work
5. Even when using the slotted links, JavaScript is still necessary. (see the `updated` method in rh-footer-social link)

## Testing Instructions

1. see the modified rh-footer main demo
2. see the old rh-footer demo, now renamed to slotted social links

### DIFF
--- a/.changeset/forty-colts-tie.md
+++ b/.changeset/forty-colts-tie.md
@@ -1,0 +1,26 @@
+---
+"@rhds/elements": minor
+---
+`<rh-footer>`: social link element can now take an `href` attribute instead of a slotted link. The previous behaviour will still work.
+
+Before:
+
+```html
+
+<rh-footer-social-link slot="social-links"
+                       icon="linkedin"><a href="https://www.linkedin.com/company/red-hat"
+                                          data-analytics-region="social-links-exit"
+                                          data-analytics-category="Footer|social-links"
+                                          data-analytics-text="LinkedIn">LinkedIn</a></rh-footer-social-link>
+```
+
+After:
+```html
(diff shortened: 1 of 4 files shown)


bennypowers:
@heyMP @zhawkins WDYT about this? similar to what we did with CTA, without back-breaking changes, but this appears to be even more accessible, without requiring any more javascript than we already do, so... win win?

zhawkins:
Similar to rh-cta, I still prefer the link-within-the-slot approach, but I don't see any reason to not have the option. Deff feels more react-y this way. I imagine that's what you're going for 😉

bennypowers:
i'm not sure what react has to do with it

the point of react is to represent batched sets of dom updates in an in-memory tree, and apply them to the real dom at intervals in batches. this along with a compiled dsl which is confusingly similar to html

what this PR is about is saying that "a social link in the footer has an href" and meaning that, because what we're doing here is making custom HTML elements for red hat properties.

zeroedin (approved):
LGTM

As we discussed during office hours, this change looks good to me. The backward compatibility is nice for sites with the previous implementation, and there are no requirements / breaking changes for now. It also allows for easier overrides of the icons used when necessary, such as in the use case for the Reddit icon, which wasn't available at the time of this PR.

PR #2186, opened 2025-02-27
docs: fix context demo height

bennypowers:
Ensures that the context background colour fills the page height

## What I did

1. add max-content height to the provider in context demo element

## Testing Instructions

1. try the [tile context demo](https://deploy-preview-2186--red-hat-design-system.netlify.app/elements/tile/demo/color-context/)


lib/elements/rh-context-demo/rh-context-demo.css
```diff
@@ -19,4 +19,5 @@ rh-surface,
   padding: var(--rh-context-demo-padding, var(--rh-space-xl, 24px));
   position: absolute;
   inset: 0;
+  height: max-content;
```
bennypowers, line 22:
```suggestion
  min-height: max-content;
```

bennypowers:
Closes #2202

PR #2201, opened 2025-03-07
docs(vpn-links): adding a lock icon for VPN links

markcaron:
## What I did

1. Added a lock icon on VPN-based links

## Testing Instructions

1. Change any link's `href` to `https://url.corp.redhat.com/` or add `.requires-vpn` class to the `<a>`
2. Does it look ok?

### DIFF
--- a/docs/styles/styles.css
+++ b/docs/styles/styles.css
@@ -49,6 +49,19 @@
       color: var(--rh-color-interactive-primary-visited-default);
       &:hover { color: var(--rh-color-interactive-primary-visited-hover); }
     }
+
+    &[href^="https://url.corp.redhat.com/"]::after,
+    &.requires-vpn::after {
+      display: inline-block;
+      content: "";
+      margin-left: var(--rh-space-xs);
+      width: var(--rh-space-lg);
+      height: var(--rh-space-lg);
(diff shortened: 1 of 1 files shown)


bennypowers:
Closes #1179

docs/styles/styles.css
bennypowers, line 1: FYI this wont select for eg `<rh-cta href="http://url.corp.redhat.com...">`

hellogreg (approved):
locks good to me

(so long as you consider providing this info advisory and optional for non-visual users)

docs/styles/styles.css
markcaron: Should be good now.

markcaron:
> (so long as you consider providing this info advisory and optional for non-visual users)

@hellogreg since we're selecting off the URL, we'd have to be sure to provide an `aria-label` for screen readers as well (or some other, more friendly approach).

markcaron:
@hellogreg and @adamjohnson, before I merge this, I discovered a link (so far) in the docs that will require the `aria-label` to inform assistive tech users of the VPN requirement. I wanted to go ahead and take care of it:

- [Accessibility Foundations](https://ux.redhat.com/accessibility/#red-hat)

What do you think about the following format?

```
<a href="https://url.corp.redhat.com/accessibility-policy" aria-label="Accessibility policy (link requires VPN access)">accessibility policy</a>
```

In particular, needing to hand code the HTML `<a>` and appending ` (link requires VPN access)` to every VPN-based link.

PR #2215, opened 2025-03-17
docs(grid): update Foundations > Grid page

marionnegp:
## What I did

1. Updated the Foundations > Grid page with content for the new rh.com grid

## Testing Instructions

1. View Grid page in DP


marionnegp:
@coreyvickery, when you review this, check out how the TOC appears above the red text and let me know if this is ok. We might want to consider a new template if we're going to use this format on more docs pages.

coreyvickery:
@marionnegp Yes, let's get rid of the table of contents.

coreyvickery:
Thanks!

coreyvickery (approved):
Great job!

PR #2217, opened 2025-03-20
fix(accordion)!: remove bordered attr

### LINKED ISSUE #2036: [bug] `<rh-accordion>`: always applies bordered attribute
`<rh-accordion>`: always applies bordered attribute, however that attribute does not seemingly do anything either in code or styles.  Does this need deprecated?  Should it be applying a style?  There does seem to be mention of it in the style/guidelines.

bennypowers:
Fixes #2036

## What I did

1. remove `bordered` from accordion. it anyways doesn't do anything


zeroedin (approved):
LGTM Gets Time Managed

PR #2223, opened 2025-03-20
docs: differentiate "Using color alone" sections on multiple pages

hellogreg:
## What I did

Wrote separate content for these pages' "Using color alone" sections:
- [Foundations > Color > Accessibility](https://deploy-preview-2223--red-hat-design-system.netlify.app/foundations/color/accessibility/#using-color-alone)
- [Accessibility > Design](https://deploy-preview-2223--red-hat-design-system.netlify.app/accessibility/design/#using-color-alone)

They shared the same verbiage before. Now...
- The `Foundations...` color page has a little more info--including a quote from WCAG on how this should not dissuade us from using color (just relying solely upon it).
- The `A11y...` page has a sentence about who is affected by reliance upon color alone.
- The latter page also now cross-links to the former.

## Testing Instructions

1. Check the text on both pages, making sure it's accurate and reads well.


markcaron (approved):
Light Generates Tints Magically

PR #2233, opened 2025-03-24
docs: destroy import map cache on error

bennypowers:
Hopefully this will solve intermittent errors when generating the import map for element pages


adamjohnson (approved):
Let's generate that map!

PR #2249, opened 2025-03-31
fix(navigation-secondary): panel scrollbar visible only when overflowed

zeroedin:
## What I did

1. The fixed static `scroll` rule has been replaced with `auto`. For users with scroll bars set to always be visible, this renders when the dropdown panels are not overflowing, which is incorrect.

## Testing Instructions

1. View deploy preview
2. Turn on scroll bars always visible in OS Settings (Mac)
3. Open dropdown
4. Reduce the viewport to less than the height of the dropdown panel.
5. Should be no scroll bar visible until overflowed.


zhawkins (approved):

adamjohnson (approved):
Less Guides, Tidy Menus

Issue #2208, opened 2025-03-10
[docs] Add reasoning for no `#000000` backgrounds in theming


hellogreg:
Since it's come up a few times, maybe we should write a sentence or two on our site about why we don't offer a pure black background in our darkest theme.

Even though white text on a black background passes WCAG conformance (there's no maximum contrast threshold), that doesn't make it accessible--as it can be tough to read.

From Benny:
> Full black has limited use in RHDS, but for use in backgrounds, ...

From Mark:
>  I can see this info on Foundations > Color. And, cross link from the Tokens page where needed.

PR #2253, opened 2025-04-01
docs(foundations): adding usage of black color content

markcaron:
## What I did

1. Added content under Foundations > Color > Usage about using black.

## Testing Instructions

1. Checkout the new "Use of black" section, and make sure it reads well.

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/2208

### DIFF
--- a/docs/foundations/color/usage/index.md
+++ b/docs/foundations/color/usage/index.md
@@ -48,7 +48,9 @@ standards website][brandstandards] if you have brand questions.
 
 Red is our primary brand color. Red is also a strong color, so **use it as an accent**, not to fill large areas.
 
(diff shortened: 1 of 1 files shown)


markcaron:
@bennypowers we might want to wire up the Tokens page to add a CTA that points to this new "Use of black" section under the White/Black tokens and the Gray tokens.

hellogreg (approved):
logos, graphics too murky 🖤

coreyvickery:
@markcaron Seems fine to me.

This is separate, but the image below the `Brand red` heading looks smaller than it should be. Can you look into this?

coreyvickery (approved):
@markcaron Looks good.

PR #2279, opened 2025-04-07
docs(navigation-primary): removing old content from Style and Guidelines docs

markcaron:
## What I did

1. Removed all the old content from Style and Guidelines on Nav Primary's docs
2. Removed the old the `overview.png` so the new `overview.svg` shines

## Testing Instructions

1. Double check the new Overview image works and the Style and Guidelines pages read ok.


bennypowers (approved):
Lose - gettit? - that manual

PR #2280, opened 2025-04-08
fix: always load dependencies when updating

bennypowers:
## What I did

1. replace `getUpdateComplete` calls with `scheduleUpdate` calls in cta and `@themable`

## Testing Instructions

1. run dev server, open cta demo
2. observe that arrow appears

## Notes to Reviewers

`getUpdateComplete()` is only called when actually accessing `el.updateComplete`, and that doesn't happen during the normal lifecycle. therefore we override `scheduleUpdate` instead

### DIFF
--- a/.changeset/deep-jobs-stay.md
+++ b/.changeset/deep-jobs-stay.md
@@ -1,0 +1,4 @@
+---
+"@rhds/elements": patch
+---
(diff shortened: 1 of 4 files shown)


zeroedin (approved):
Let's Go Themeable Maintenence

PR #2282, opened 2025-04-08
fix(button): always use the current text color for icons

bennypowers:
ensures legibility in all color schemes

## What I did

1. set the icon color for buttons which have one to `currentcolor`: i.e. the color-scheme dependent text color for a given button

## Testing

https://deploy-preview-2282--red-hat-design-system.netlify.app/elements/button/demo/color-context/

image
image


zeroedin (approved):
🚀

Issue #2295, opened 2025-04-09
[docs][bug] `<rh-tile>` Demos TOC has layout issues


markcaron:
### Description

The TOC on `<rh-tile>` Demos is

1. Way too long (and won't scroll), making the last links impossible to reach
2. Squishes the arrows and rags the left-alignment on longer link text

Note: On #2 it seems that the `display: inline-flex;` is allowing the left "column" to shrink as needed, which shrinks the SVG.

### Screenshots

**Horizontal**
<img width="1211" alt="Image" src="https://github.com/user-attachments/assets/82531496-193a-4e41-9d89-1fb6b6f89729" />

**Vertical**
<img width="465" alt="Image" src="https://github.com/user-attachments/assets/8c512750-3740-48ce-8fd9-f9a0632675f2" />

### Steps to reproduce

1. Go to [Tile's Demo page](https://ux.redhat.com/elements/tile/demos/)
2. See wonky TOC

### Expected behaviour

1. All of the demo links in the vertical TOC should be reachable
2. The arrow sizes and left-alignment should be consistent

### Operating System (OS)

macOS

### Browser

Chrome

PR #2296, opened 2025-04-09
docs(uxdot-toc): fix TOC arrow size and alignment

markcaron:
## What I did

- [x] Fix arrow size and alignment in demos TOC

## Testing Instructions

1. Go to Tile's demo in DP
2. Compare to live version

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/2295


markcaron:
Will tackle fixing vertical TOC overflow on long demo TOCs, like Tiles, via reducing the number of items. It seems that the main source of this issue is the size of the list—no list should be that long.

bennypowers (approved):

Issue #2300, opened 2025-04-09
[bug] `<rh-tile>` with empty `body` and `footer` slots adds extra spacing


markcaron:
### Describe the bug

When a Tile has both `body` and `footer` slots empty, an `::after` pseudo-element in the `body` slot adds unnecessary spacing between the headline and the arrow.

This issue came from the [empty `footer` slot PR in Cubone](https://github.com/RedHat-UX/red-hat-design-system/pull/2096).

### Screenshots

<img width="918" alt="Image" src="https://github.com/user-attachments/assets/79d38743-27fe-4073-b526-cdfc6a2397a8" />

### Steps to reproduce

1. Go to ["Without Footer Content" demo for Tile](https://ux.redhat.com/elements/tile/demo/without-footer-content/)
2. Remove `body` (default) slot content
3. See extra spacing

### Expected behaviour

The extra spacing shouldn't exist.

### Element HTML

```HTML
<rh-tile>
  <h2 slot="headline"><a href="#top">Link</a></h2>
</rh-tile>
```

### Operating System (OS)

iOS

### Browser

Chrome

bennypowers:
Check on #2294

markcaron:
Can we test locally (and not via SSR) so we can get this patch out? And then revisit the SSR part after?

Seems like adding the SSR hints to the demo would be more of a docs fix/update.

PR #2302, opened 2025-04-09
fix(tile): empty body + footer slots

markcaron:
## What I did

1. Fix extra spacing on Tiles with empty `body` and `footer` slots

## Testing Instructions

1. Check [Tile's Without Footer Content demo](https://deploy-preview-2302--red-hat-design-system.netlify.app/elements/tile/demo/without-footer-content/)
2. Ensure spacing is correct on both empty `footer` Tiles and empty `footer` and `body` Tiles

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/2300


bennypowers:
https://github.com/patternfly/patternfly-elements/pull/2903 should help with `isEmpty`

markcaron:
@bennypowers there seems to be a regression in the previous `::after` fix. The arrow is too close to the body content again.

image

markcaron:
@bennypowers I don't think `${classMap({ empty: this.#slots.isEmpty() })}` works as well as `${classMap({ empty: !hasSlottedFooter })}` for some reason.

bennypowers:
try again after merging #2307

markcaron:
Can we test locally (and not via SSR) so we can get this patch out? And then revisit the SSR part after?

Seems like adding the SSR hints to the demo would be more of a docs fix/update.

zeroedin:
We believe that the SSR hints aren't working correctly here `ssr-hint-has-slotted-default` when not present should cause an `empty` class on the deafult slot:

```
<slot id="body" class="${classMap({ empty: this.#slots.isEmpty() })}"></slot>
```

See [without-footer-content.html](https://deploy-preview-2302--red-hat-design-system.netlify.app/elements/tile/demo/without-footer-content/), no hints are added and the auto hinting gets added as `ssr-hint-has-slotted-default="true"`.  The side effect to that is no empty class on the default slot.

Screenshot 2025-04-14 at 4 32 09 PM

This will be broken out as a seperate issue.

zeroedin (approved):
Logical Generative Thinking Machine

Issue #2278, opened 2025-04-07
[bug] `<rh-jump-links>` design fixes


marionnegp:
### Describe the bug

- [ ] In hover and focus states, the text should use `--rh-color-text-primary`.
- [ ] The space between the jump links and the page content should be `--rh-space-2xl` for the demos.
- [ ] The red border for the active state should use `--rh-border-width-lg`.
- [ ] In the nested jump links demo, I'm not seeing the thick gray border appear for the nested "Heading 3" link in its hover or focus states.
- [ ] The nested jump links demo doesn't show the jump links turning into a disclosure for mobile breakpoints. Is this a bug in the element or just the demo?
- [ ] When jump links collapse into a disclosure, the red border for the open disclosure should be removed.
- [ ] Need to add the vertical sticky variant inside a container. (It's at the bottom of the vertical jump links pages in [mockups](https://www.figma.com/design/UYlbwsOPXfnbJ9rjJxJBaP/rh-jump-links?node-id=2144-3434&m=dev).) If this _has_ to be done in Diglett instead of in a patch release earlier, just let me know, and I can add a "Planned" tag next to this info in the docs.
- [ ] Can long jump link text wrap? I had chatted with @adamjohnson about having the element's user add a `width`/`max-width` to the jump links' container to make it wrap, but it doesn't currently work that way in the demo.
- [ ] Add dark scheme

![Image](https://github.com/user-attachments/assets/e009b58f-cdfd-4614-aec9-29bd2eb9b1a0)

Note: There are questions around the design direction for horizontal jump links. If there are changes, I'll update this issue once I've been able to chat with @coreyvickery.

### Steps to reproduce

Check the [jump links demos](https://ux.redhat.com/elements/jump-links/demos/) from the Cubone release.

### Expected behaviour

See description of the bug.

### Element HTML

```HTML

```

### Operating System (OS)

macOS

marionnegp:
@bennypowers, I'm seeing the jump links move into a disclosure on small screens only in the responsive demo. Are the jump links supposed to switch to the disclosure when the screen gets small for all demos?

bennypowers:
@ArathyKumar PTAL at this

ArathyKumar:
> [@ArathyKumar](https://github.com/ArathyKumar) PTAL at this

Sure @bennypowers 
Thank you!

PR #2308, opened 2025-04-11
fix(jump-links): design issues

ArathyKumar:
## What I did

Resolved the issues mentioned in  https://github.com/RedHat-UX/red-hat-design-system/issues/2278 

Describe the bug

- [x] In hover and focus states, the text should use --rh-color-text-primary.
- [x] The space between the jump links and the page content should be --rh-space-2xl for the demos.
- [x] The red border for the active state should use --rh-border-width-lg.
- [x] In the nested jump links demo, I'm not seeing the thick gray border appear for the nested "Heading 3" link in its hover or focus states.
- [x] The nested jump links demo doesn't show the jump links turning into a disclosure for mobile breakpoints. Is this a bug in the element or just the demo? - _`This is not a bug. The demo just needs some responsive CSS to support that behavior on mobile.`_
- [x] When jump links collapse into a disclosure, the red border for the open disclosure should be removed.
- [x] Need to add the vertical sticky variant inside a container. (It's at the bottom of the vertical jump links pages in [mockups](https://www.figma.com/design/UYlbwsOPXfnbJ9rjJxJBaP/rh-jump-links?node-id=2144-3434&m=dev).) If this has to be done in Diglett instead of in a patch release earlier, just let me know, and I can add a "Planned" tag next to this info in the docs. - `Will be resolved as part of the ticket` _https://github.com/RedHat-UX/red-hat-design-system/issues/2401_
- [x] Can long jump link text wrap? I had chatted with @adamjohnson about having the element's user add a width/max-width to the jump links' container to make it wrap, but it doesn't currently work that way in the demo.

## Testing Instructions

1. Verify all the above mentioned issues are resolved.


elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -76,6 +77,10 @@ export class RhDisclosure extends LitElement {
   @query('summary') private summaryEl!: HTMLElement;
 
   render() {
+    const hasJumpLinks: RhJumpLinks | null = this.querySelector('rh-jump-links');
```
bennypowers, line 80: avoid side effects in render

this also should not be exposed on the host element: users should not have the final say on whether or not there's a jump links element here.

rather, set up a mutation observer and toggle some private state.

bennypowers (changes requested):

elements/rh-disclosure/rh-disclosure.ts
ArathyKumar: Hey @bennypowers 

Added the changes. 
Could you please check if it looks good

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -76,11 +78,12 @@ export class RhDisclosure extends LitElement {
   @query('details') private detailsEl!: HTMLDetailsElement;
   @query('summary') private summaryEl!: HTMLElement;
 
+  connectedCallback() {
+    super.connectedCallback();
+    this.#mutationsCallback();
```
bennypowers, line 83: this will fail in SSR because of the querySelector call

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -127,6 +130,13 @@ export class RhDisclosure extends LitElement {
     this.open = false;
     this.summaryEl.focus();
   }
+
+  #mutationsCallback(): void {
+    const hasJumpLinks: RhJumpLinks | null = this.querySelector('rh-jump-links');
+    if (hasJumpLinks && ! this.hasAttribute('has-jump-links')) {
+      this.setAttribute('has-jump-links', '');
```
bennypowers, line 137: let's use a shadow class here instead of a host attr

bennypowers (changes requested):

elements/rh-disclosure/rh-disclosure.ts
bennypowers: put this firstUpdated

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -122,6 +140,13 @@ export class RhDisclosure extends LitElement {
     this.open = false;
     this.summaryEl.focus();
   }
+
+  #mutationsCallback(): void {
+    const hasJumpLinks: RhJumpLinks | null = this.querySelector('rh-jump-links');
+    if (hasJumpLinks && !this.classList.contains('has-jump-links')) {
+      this.classList.add('has-jump-links');
```
bennypowers, line 147:
```html
<rh-disclosure class="has-jump-links">
```
this is no good because it exposes internal state as a public api


we want something like this
```html
<rh-disclosure>
  #shadow-root
  <details  class='has-jump-links'>...
</rh-disclosure>
```

if you need to style the host based on whether or not the shadow root has an element with that class, I think this might work, but check across browsers
```css
:host:has(.has-jump-links) {
 ...
```

elements/rh-disclosure/rh-disclosure.ts
bennypowers: even if we avoid crashing the ssr process

we'll still get a FOUC between when the disclosure DSD template renders on screen and when the mutation observer fires, adding the shadow class

therefore, try to use events

we might use context for this, because context runs in ssr. HOWEVER, context is *only* for passing information from ancestor elements to descendants. In our case, we want to pass information from the children (jump links) up to the parent (disclosure)


So what we might do is fire an event from jump links, and catch it in disclosure


```ts
// jump links
connectedCallback() {
  super.connectedCallback();
  this.dispatchEvents(new Event('jump-links-connect')); // might need to bubble
}
```


```ts
// disclosure
constructor() {
  super();
  this.addEventListener('jump-links-connect', e => {
    this.#hasJumpLinks = true
    this.requestUpdate();
    event.stopPropagation();
  });
}

connectedCallback() {
  super.connectedCallback();
  this.#mo.observe(...); // handle if jump links are removed
}
```

bennypowers: disadvantages to this approach, even if it works
- events are always public, and this whole thing is necessarily private - i.e. we don't want anyone else listening for `jump-links-connect`. docs are not a foolproof solution
- this may cause SSR rendering mismatch problems. we'll need to ensure that the state of the template is consistend between server and client render


```ts
connectedCallback() {
  if (!isServer && this.shadowRoot) // ssr happened, we have dsd template
    this.#hasJumpLinks = this.shadowRoot.querySelector('.has-jump-links');
  super.connectedCallback();
}
```
⬆️ this may not end up being a problem though, if there are no ssr client hydration errors

bennypowers: @adamjohnson @zeroedin @marionnegp can it ever happen that a page is rendered as mobile-first with the jump-links-containing-disclosure open by default? If not, then @ArathyKumar points out we can avoid ssr shenanigans by just querying for jump linnks on toggle. BUT if a mobile-first page can have the disclosure containing the jump links open by default, then we have to render all of this server side as well

adamjohnson: I vote KISS. If users want to see the jump links, they can open the disclosure.

Users are looking for content first. Tapping a jump link would be a secondary/tertiary action.

elements/rh-disclosure/rh-disclosure.ts
adamjohnson:
```css
:host:has(.has-jump-links) {}
```
> ...but check across browsers

Check Firefox IIRC. Probably good to go in Chrome.

elements/rh-disclosure/rh-disclosure.ts
zeroedin: So we are saying the `<rh-disclosure open>` would/should never be set by default when with a child of `<rh-jump-links>`? Or are we saying the open state shouldn't be explicitly driven by the viewport size when used in the jump links pattern?

Likely I don't fully understand the problem here.  

My interpretation is: 

We are saying the user follows a #link to a page that contains jump links, and on load, that page scrolls to that anchor. While doing so, the parent rh-disclosure should maybe open in a mobile viewport if it is closed by default.

What happens to the rh-disclosure at larger viewports in this scenario? I'm not sure we should open the disclosure in the story, but we should allow the user to do so if they want to see the list, maintaining the default state set.

If I'm missing something, let me know.

bennypowers: The feature we're implementing is special jumps links only styling for the disclosure element. This style (a red vertical line) is not something users are meant to be able to opt in or out of, and it's a special case within the disclosure styles intended *only* for the pattern were discussing. (@coreyvickery @marionnegp please confirm or correct me on that)

In this exceptional case, the page has two identical jump links on it:

1. One jump links is hidden at mobile page widths
2. The other is located inside a disclosure. That disclosure is hidden at non-mobile page widths.

We're discussing the state of the disclosure, the one that's only visible in narrow widths. If we accept that the dislosure defaults to closed in all cases, that masks it much easier to implement this feature. 

Because the special styles on the disclosure depend entirely on the presence or absence of the jump links element inside of it, we need to check the dom before applying those styles.

But, since the special styles are only visible when the disclosure is open, if we say that the disclosure can only be opened by a user action, then we can wait until the DOM is there before deciding whether or not to render those styles.

If, however, we decide that that disclosure can be set to open, even on the server side, before we know whether or not it's mobile or any of that, so then we have to get much more creative about discovering whether or not there's a jump links inside or not.

elements/rh-disclosure/rh-disclosure.ts
bennypowers: Screenshot From 2025-05-07 15-16-07
Left to right: Webkit, Firefox, Chromium

bennypowers: `:host:has` is not on interop 2025, but maybe 2026:

https://github.com/web-platform-tests/interop/issues/791

ArathyKumar: Hey @bennypowers 

How can pseudo-elements help address this issue?

We need a selector or attribute on the rh-disclosure component to determine whether it contains an rh-jumplinks child, so that the appropriate CSS can be applied. 

```
:host(.mobile.has-jump-links) details[open]:before {
  border-inline-start-color: transparent;
}
```

Can pseudo-elements be used to achieve this?"

elements/rh-disclosure/rh-disclosure.ts
zeroedin: >  If we accept that the dislosure defaults to closed in all cases, that masks it much easier to implement this feature.

I think this is likely the best way to to go.

elements/rh-disclosure/rh-disclosure.ts
ArathyKumar: Got it. I will try these two approaches.

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -141,11 +136,19 @@ export class RhDisclosure extends LitElement {
     this.summaryEl.focus();
   }
 
-  #mutationsCallback(): void {
-    const hasJumpLinks: RhJumpLinks | null = this.querySelector('rh-jump-links');
-    if (hasJumpLinks && !this.classList.contains('has-jump-links')) {
-      this.classList.add('has-jump-links');
-    }
+  #handleJumpLinkChildren(event: Event) {
+    const sheet: CSSStyleSheet = new CSSStyleSheet();
```
bennypowers, line 140: do this at the top level at the module, so that we're only parsing the css string once, but adopt the same stylesheet object here in the callback, so we don't have to recreate it each time.

bennypowers (changes requested):

bennypowers:
@ArathyKumar see 4a56626. This is what i mean by "parsing the stylesheet once and applying it many times"

That commit also adds some private state and a mutation observer. we can and should remove the `.mobile` selector on host, because that's not our concern as element authors.

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -72,8 +80,20 @@ export class RhDisclosure extends LitElement {
    */
   @property({ reflect: true }) summary?: string;
 
+  @state() private hasJumpLinks = false;
```
bennypowers, line 83: You can use this with classMap to toggle a `hasJumpLinks` class on the details element in the shadow root. With the mutation observer toggling that state depending on the presence of jump links, if the jump links is removed, the styles will be too.

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -72,13 +81,31 @@ export class RhDisclosure extends LitElement {
...
+
   @query('details') private detailsEl!: HTMLDetailsElement;
   @query('summary') private summaryEl!: HTMLElement;
 
+  #mo = new MutationObserver(() => this.#mutationsCallback());
```
bennypowers, line 89: IIRC, we can remove the mutation observer, and just calculate `hasJumpLinks` when the disclosure opens

```ts
@observes('open')
protected _openChanged() {
  // calculate hasJumpLinks here
}
```

elements/rh-jump-links/demo/responsive.html
```diff
@@ -5,7 +5,7 @@
     <rh-jump-links aria-labelledby="jump-links-title">
       <rh-jump-link href="#heading-1"
                     active>Heading 1</rh-jump-link>
-      <rh-jump-link href="#heading-2">Heading 2</rh-jump-link>
+      <rh-jump-link href="#heading-2">Heading 2 </rh-jump-link>
```
bennypowers, line 8: revert whitespace change

elements/rh-jump-links/demo/rh-jump-links.html
```diff
@@ -21,7 +21,7 @@ <h2 id="sections">Sections</h2>
   #jump-links-demo-container {
     display: grid;
     grid-template-columns: max-content auto;
-    gap: var(--rh-space-lg);
+    gap: var(--rh-space-2xl, 32px);
```
bennypowers, line 24:
```suggestion
    gap: var(--rh-space-2xl);
```

elements/rh-jump-links/rh-jump-link.css
```diff
@@ -2,13 +2,18 @@
   display: block;
   position: relative;
   width: max-content;
+
+  --rh-jump-link-max-width: var(--rh-length-7xl, 128px);
```
bennypowers, line 6:
```suggestion
```

elements/rh-jump-links/rh-jump-link.css
```diff
@@ -2,13 +2,18 @@
...
 }
 
 a {
   color: var(--rh-color-interactive-secondary-default);
   text-decoration: none;
   display: inline-block;
   box-sizing: border-box;
+  max-width: var(--rh-jump-link-max-width);
```
bennypowers, line 14:
```suggestion
  max-width: var(--rh-jump-link-max-width, var(--rh-length-7xl, 128px));
```

bennypowers (changes requested):

elements/rh-disclosure/rh-disclosure.ts
ArathyKumar: Added the changes.

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -16,6 +16,15 @@ export class DisclosureToggleEvent extends Event {
   }
 }
 
+const hasJumpLinksStyleSheet = new CSSStyleSheet();
+if (!isServer) {
+  hasJumpLinksStyleSheet.replaceSync(/* css */`
+    details[open]:before {
```
bennypowers, line 22:
```suggestion
    details.has-jump-links[open]:before {
```

elements/rh-disclosure/rh-disclosure.ts
ArathyKumar: I have added this code 

> connectedCallback() {
    super.connectedCallback();
    if (!isServer) {
      this.hasJumpLinks = !!this.querySelector('rh-jump-links');
    }
  }

in the connectedCallback() as well as the **red border flashes for a fraction of a second** upon opening the modal.

elements/rh-jump-links/rh-jump-link.css
```diff
@@ -9,6 +9,9 @@ a {
   text-decoration: none;
   display: inline-block;
   box-sizing: border-box;
+  max-width: var(--rh-jump-link-max-width, var(--rh-length-7xl, 128px));
```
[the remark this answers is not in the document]
coreyvickery: @adamjohnson Do we have to add one at all or could we let the consumer set that?

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -1,13 +1,14 @@
-import { LitElement, html } from 'lit';
+import { LitElement, html, isServer, type PropertyValues } from 'lit';
```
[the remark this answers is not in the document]
ArathyKumar: Added the changes

elements/rh-disclosure/rh-disclosure.ts
bennypowers: i removed the connectedCallback and didn't see any flash. by default, `@observes` runs in the update callback, which is before dom changes are propagated, so the state will be set prior along with the detail's element's open attribute.

bennypowers:
@coreyvickery I added in some default css for *headings slotted into the summary*. They'll be inline by default. however NB that many users will still need to unset their page css, if they for example set headings to be display: flex. See disclosure's nested-jump-links demo for more info

bennypowers (approved):

.stylelintrc.yml
```diff
@@ -68,6 +68,7 @@ rules:
     - allowed:
       - --rh-icon-size
       - --rh-navigation-active-item-color
+      - --rh-jump-link-max-width
```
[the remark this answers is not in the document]
bennypowers: https://github.com/RedHat-UX/red-hat-design-system/wiki/JSDoc#css-custom-properties

bennypowers:
Please address remaining comments, then we can merge

elements/rh-jump-links/rh-jump-link.css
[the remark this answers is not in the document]
markcaron: @coreyvickery the consumer can add one through the `--rh-jump-link-max-width` var.

@ArathyKumar we should consider renaming it to `--rh-jump-links-max-width` (with an 's' on links) so it's consistent with the element name.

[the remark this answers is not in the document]
markcaron: @coreyvickery, what should be the default max-width of Jump Links?

marionnegp:
For a single jump link item in the horizontal variant, I think we can use 200px as a max width. This is the same max width as a single `rh-tab`.

.stylelintrc.yml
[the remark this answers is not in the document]
markcaron: @adamjohnson can you give this a re-review?

elements/rh-jump-links/rh-jump-link.css
```diff
@@ -22,6 +25,7 @@ a {
       calc(var(--rh-space-lg, 16px) + var(--_padding-inline-start, 0px))
       var(--rh-space-lg, 16px);
     border-inline-start: var(--_bdr);
+    max-width: var(--rh-jump-link-max-width, calc(var(--rh-length-md, 8px) * 25));
```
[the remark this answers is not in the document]
marionnegp: I think it's ok to leave the max-width undefined for horizontal jump links. I _think_ that's the case for horizontal `rh-tabs`, so that behavior would be consistent with another RH element.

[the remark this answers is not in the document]
markcaron: @adamjohnson @marionnegp can you give this one more re-review?

1. I set the default `max-width` on `vertical` to `320px` (using some `calc()` based on the current `md` spacer size—figuring that'd be a good system-based number to utilize instead of a hard `px` value)
2. I removed the `max-width` property from the default `a` (where it was), so that `horizontal` `jump-link` elements have no `max-width` (causing the breakage Adam mentioned above).

elements/rh-jump-links/rh-jump-link.ts
```diff
@@ -16,6 +16,8 @@ import style from './rh-jump-link.css';
 import { rhJumpLinksOrientationContext } from './context.js';
 
 /**
+ * @cssprop  {<length>} [--rh-jump-link-max-width=200px|320px]
+ *           max-width for each jump link
```
adamjohnson: If we're not going to use this var for horizontal jump links, it'd be good to note that here.

```suggestion
 * @cssprop  {<length>} [--rh-jump-link-max-width=320px]
 *           max-width for each vertical jump link
```

adamjohnson (approved):
Links gain trendy makeover

PR #2332, opened 2025-04-18
fix(tabs): corrected hidden icons

zeroedin:
## What I did

1. Corrected hidden icons on tabs. 
2. Removed SlotController

## Testing Instructions

1. View the[ icons demo](https://deploy-preview-2332--red-hat-design-system.netlify.app/elements/tabs/demo/icons-and-text/)

## Notes to Reviewers

This is likely a SlotController issue but in review of the code I just didn't feel great about the modification of the display rule on slot and by removing that and moving the margin from the icon to the gap rule on the button flex already gets the same effect if a icon is slotted or not.

### DIFF
--- a/.changeset/eighty-results-bow.md
+++ b/.changeset/eighty-results-bow.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tabs>`: corrected icons from being incorrectly hidden 
+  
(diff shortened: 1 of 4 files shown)


marionnegp (approved):

bennypowers (approved):
Let's consider adding an `icon`/`icon-set` attr to rh-tab, to pair with the slot. That would release 3.1.0, so we can follow up in a later PR

PR #2331, opened 2025-04-18
fix(tabs): correct vertical tabs layout and background styles

zeroedin:
## What I did

1. Corrected vertical tab styles
2. Corrected background styles inherited from parent `color-scheme` and set `color-palette`

Closes #2320 

## Testing Instructions

1. Check all [DP demos](https://deploy-preview-2331--red-hat-design-system.netlify.app/elements/tabs/demos/), ensure correct rendering of tabs

PR #2334, opened 2025-04-18
fix: oklch token fallbacks

zeroedin:
## What I did

1. Corrected missing token fallbacks in oklch usages.  

## Testing Instructions

1.

## Notes to Reviewers

This shows up when global.css isn't loaded.  Components should still provide these fallbacks.

Note: I purposely didn't fix the one in tabs in this PR. That is being fixed #2331 as part of the vertical tabs correction.


marionnegp:
@coreyvickery, missed this one review. Do you mind taking a look to make sure the colors are right?

bennypowers (approved):

Issue #2314, opened 2025-04-14
[bug] SSR hinting doesn't seem to be working correctly


markcaron:
We believe that the SSR hints aren't working correctly here `ssr-hint-has-slotted-default` when not present should cause an `empty` class on the deafult slot:

```
<slot id="body" class="${classMap({ empty: this.#slots.isEmpty() })}"></slot>
```

See [without-footer-content.html](https://deploy-preview-2302--red-hat-design-system.netlify.app/elements/tile/demo/without-footer-content/), no hints have been manually added and the auto hinting gets added as `ssr-hint-has-slotted-default="true"`.  The side effect to that is no empty class on the default slot.

![Screenshot 2025-04-14 at 4 32 09 PM](https://github.com/user-attachments/assets/d95a42e8-a3ff-4f31-bdbb-0c04b604bbe5)

This will be broken out as a seperate issue.

_Originally posted by @zeroedin in https://github.com/RedHat-UX/red-hat-design-system/issues/2302#issuecomment-2802921133_

bennypowers:
i'll fix this when i get back

PR #2336, opened 2025-04-20
docs: correctly generate default slot ssr hints

bennypowers:
## What I did

1. fix the docs transformer which injects ssr hint attributes

## Testing Instructions

1. View the demos for [rh-tile](/elements/tile/demo/without-footer-content)
2. check that the appropriate ssr hints are applied to each host element

## Notes to Reviewers

This is one part of a two-part fix for #2314


bennypowers:
see also https://github.com/patternfly/patternfly-elements/pull/2906

zeroedin (approved):
After bringing over the controller update from Patternfly this looks good!

PR #2340, opened 2025-04-23
chore: add dtls (prerelease) config

bennypowers:
## What I did

1. add config for dtls (prerelease)


markcaron (approved):
Language-server Gets Tokens Manual

PR #2345, opened 2025-04-28
fix(footer): overflow scrollbar hidden unless overflowed

zeroedin:
## What I did

1. Corrected visibility of overflow scrollbar

## Testing Instructions

1. View DP and turn on always visible scrollbars; DP preview shouldn't have a scrollbar in the universal footer unless overflowed. Previously looks like: 

image


.changeset/two-pillows-prove.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-footer>`: corrected overflow scroll bar visibility
```
adamjohnson, line 5:
```suggestion
`<rh-footer>`: only show vertical scrollbars if content overflows
```

adamjohnson (approved):
Layout gains trimmed margins

bennypowers (approved):

PR #2351, opened 2025-04-30
docs(pattern-card): adding definition list pattern

markcaron:
## What I did

1. Added a definition list card pattern

## Testing Instructions

1. Check out the new pattern
2. Designers: check styling as a new definition list standard
3. Devs: check HTML and CSS for the pattern


bennypowers:
cc @lkrzyzanek

marionnegp (changes requested):
I think there could be less space between a `<dt>` and `<dd>` to make them look more associated with each other. I think an 8px space between the two works well. 

Example:
Screenshot 2025-04-30 at 12 44 54 PM

It also looks like there isn't enough padding around all the content, compared to this [promo demo](https://deploy-preview-2351--red-hat-design-system.netlify.app/elements/card/demos/#demo-promo).

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+  <rh-cta slot="footer" href="https://pokemondb.net/pokedex/charmander">Charmander Pokédex</rh-cta>
+</rh-card>
+
+<style>
+  .example-dl {
+    /* Definition list itself */
+    & dl {
+      display: flex;
+      flex-flow: column;
+      gap: var(--rh-space-lg);
```
adamjohnson, line 23:
```suggestion
      gap: var(--rh-space-lg, 16px);
```

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+      margin: 0;
+      padding: 0;
+      & dt {
+        font-weight: var(--rh-font-weight-heading-medium);
+      }
```
adamjohnson, line 28:
```suggestion

      & dt {
        font-weight: var(--rh-font-weight-heading-medium, 500);
      }
```

Add space between declarations + fallback for font weight.

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+      & dd {
+        margin: 0;
+        padding-block-end: var(--rh-space-lg);
+        border-block-end: var(--rh-border-width-sm) solid var(--rh-color-border-subtle);
+      }
```
adamjohnson, line 33:
```suggestion

      & dd {
        margin: 0;
        padding-block-end: var(--rh-space-lg, 16px);
        border-block-end: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle);
      }
```

Add space between declarations + fallback for border width.

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+    }
+    /* For this Pokémon example */
```
adamjohnson, line 35: Add one extra line here between the `}` and the comment (appeases linter).

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+    }
+    /* For this Pokémon example */   
+    &::part(container) {
+      display: grid;
+      grid-template-areas: "image" "header" "body" "footer";
+      grid-template-columns: 1fr;
+      justify-items: stretch;
+      align-items: start;
+      gap: var(--rh-space-2xl);
+    }
```
adamjohnson, line 43:
```suggestion
    &::part(container) {
      display: grid;
      grid-template-areas: 'image' 'header' 'body' 'footer';
      grid-template-columns: 1fr;
      place-items: start stretch;
      gap: var(--rh-space-2xl, 32px);
    }
    
```

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+    @container (min-width: 576px) {
+      &::part(container) {
+        grid-template-areas: "image header" "image body" "image footer";
+        grid-template-columns: 1fr 2fr;
+      }
+    }
```
adamjohnson, line 49:
```suggestion
    @container (min-width: 576px) {
      &::part(container) {
        grid-template-areas: 'image header' 'image body' 'image footer';
        grid-template-columns: 1fr 2fr;
      }
    }

```

Single quotes appeases the linter.

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,56 @@
...
+    &::part(image) {
+      grid-area: image;
+      padding: var(--rh-space-2xl);
```
adamjohnson, line 52:
```suggestion
      padding: var(--rh-space-2xl, 32px);
```

adamjohnson (changes requested):
Changes look like more than they are. Just add some fallbacks. Perhaps consider alphabetizing CSS properties in each selector.

markcaron:
> It also looks like there isn't enough padding around all the content, compared to this [promo demo](https://deploy-preview-2351--red-hat-design-system.netlify.app/elements/card/demos/#demo-promo).

@marionnegp I think this may be a design bug / inconsistency between regular `<rh-card>` and `<rh-card variant="promo">`. I'm seeing some `@container` query adjustments to `#promo` that don't exist in Card:

image

— vs —

image

Anyway, I'm not adjusting these values in my pattern CSS. I am setting a `gap` between the left and right `grid` columns as `32px` though. Does it need to be `48px` like Promo Cards? Note: I didn't make this a Promo card.

docs/patterns/card/patterns/definition-list.html
markcaron: Done

adamjohnson (approved):
List Gains The Makeup

marionnegp (changes requested):
Makes sense that you wouldn't switch this to the promo variant. I think I got confused with some of the discussion from chat.

For spacing, I was thinking something like this:
Screenshot 2025-05-01 at 3 34 48 PM.

docs/patterns/card/patterns/definition-list.html
```diff
@@ -0,0 +1,60 @@
+<rh-card class="example-dl">
+  <img src="../charmander.png" alt="Charmander Pokémon" slot="image">
```
lkrzyzanek, line 2: Could you please add one more example where image is not present?
So the card contains just facts (DL) and no (IMAGE) tag?

lkrzyzanek (changes requested):

docs/patterns/card/patterns/definition-list.html
lkrzyzanek: The reason is that we would need to cover both scenarios where the same card may include an image and may not. So having 1 implementation covering both scenarios would be ideal.
Thanks!

marionnegp (changes requested):
Definition list spacing looks good to me, but now I'm seeing extra 16px of space above list items in the `List with dividers` example. Can that be fixed and changed to equal 16px, like the space below the list item text?
Screenshot 2025-05-27 at 12 26 01 PM

PR #2352, opened 2025-05-01
POC: open demos in codepen

zeroedin:
## What I did

1. Explored a proof of concept for creating Codepens from demo code.

## Testing Instructions

1.  Visit an element demo page, click open in codepen button on a demo, you should create a new Codepen with a working version of that demo.

## Notes to Reviewers

There is some hacky stuff here, namely dealing with lightdom CSS, where the method needs to be abstracted more to deal with any `<link rel>` stylesheets that could be present.   Also, we are reliant on a 3rd party CDN, meaning on release day likely, none of the demos would work. That's not great. I'd love to see our CDN be whitelisted for Codepen so we can rely on our infrastructure for these demos; that way, the code base will always be the same.

That said, I'm not convinced yet that the approach in this PR is the implementation we want. I want to look deeper into Codepen's APIs to see if there are other ways of doing this. 

This POC will explore the concept and maybe consider ways to improve it.  If you have an idea, jump in :)


markcaron:
@zeroedin, Looking good so far!

This might be an opportunity to integrate `typography.css` (and maybe any other ones we might need) into the CodePen.

PR #2354, opened 2025-05-03
docs: added url target to open in new tab

rohit-bharmal:
## What I did

1. Added target of url opening to new tab

## Testing Instructions

1. Test the "View in Own Tab" link on demo docs. It should open the link in new tab.


bennypowers (approved):

uxdot/uxdot-demo.ts
```diff
@@ -48,7 +48,7 @@ export class UxdotDemo extends LitElement {
                      icon-set="ui"
                      @click="${this.#reloadIframe}">Reload</rh-button>
           <rh-cta slot="footer" href="${this.demoSourceUrl}">View source on GitHub</rh-cta>
-          <rh-cta slot="footer" href="${this.demoUrl}">View In Own Tab</rh-cta>
+          <rh-cta slot="footer" href="${this.demoUrl}" target="_blank" rel="noreferrer noopener">View In Own Tab</rh-cta>
```
markcaron, line 51: Thanks for tackling this @rohit-bharmal! It was one of the things on my to-do list.

I could be wrong, but I don't believe we need the `rel="noreferrer noopener"` on these demo links since they're our own, internal links.

markcaron (changes requested):

uxdot/uxdot-demo.ts
bennypowers: Makes sense to me

markcaron (approved):
Link Gets Target Mechanism

PR #2356, opened 2025-05-06
fix(navigation primary): compact state await updateComplete

zeroedin:
## What I did

1. Added a check for `updateComplete` if the hamburger menu was unavailable.

## Testing Instructions

1. This is hard to test on the DP as it gets masked by DSD.    The best way I've found to confirm the change is by overriding the JS file on the [Code Pen ](https://codepen.io/zeroedin/pen/raavPGV)where the issue does occur with the change in this branch to confirm the issue is no longer present.

### DIFF
--- a/.changeset/forty-colts-read.md
+++ b/.changeset/forty-colts-read.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-navigation-primary>`: corrected state race condition on menu load'
+  
--- a/elements/rh-navigation-primary/rh-navigation-primary.ts
+++ b/elements/rh-navigation-primary/rh-navigation-primary.ts
@@ -410,11 +410,17 @@ export class RhNavigationPrimary extends LitElement {
   }
 
   async #openHamburger() {
+    if (!this._hamburger) {
+      await this.updateComplete;
+    }
     this._hamburger.open = true;
     await this.updateComplete;
   }
 
   async #closeHamburger() {
+    if (!this._hamburger) {
+      await this.updateComplete;
+    }
     this._hamburger.open = false;
     await this.updateComplete;
   }


.changeset/forty-colts-read.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-navigation-primary>`: corrected state race condition on menu load'
+
```
bennypowers, line 6:
```suggestion
`<rh-navigation-primary>`: prevent an error from being thrown in certain circumstances when the element connects
```

bennypowers (changes requested):
Can you add regression tests? use the `@web/test-runner-commands` package to set viewport size in beforeEach

elements/rh-navigation-primary/test/rh-navigation-primary.spec.ts
```diff
@@ -69,6 +69,10 @@ describe('<rh-navigation-primary>', function() {
         expect(element.shadowRoot?.querySelector('#container')?.classList.contains('compact')).to.be.true;
       });
 
+      it('should have a closed hamburger', function() {
+        expect(element.shadowRoot?.querySelector<HTMLDetailsElement>('#hamburger')?.open).to.be.false;
```
bennypowers, line 73: we should use ax tree assertions here instead of asserting against the dom, same below

bennypowers (changes requested):

elements/rh-navigation-primary/test/rh-navigation-primary.spec.ts
zeroedin: I don't believe the axe tree will represent what I'm testing for here. If you have an idea of what you are looking for, please do share. 

My current understanding by poking around in chrome/firefox accessibility tree view: 
In desktop viewport, the hamburger doesn't really exist in the axe tree. I'm guessing but its likely due to `display: none` on the summary?  So, the state of `expanded: true` on that node in the axe tree also doesn't exist.  I believe that is what I believe I would want to test for. If this is off base, let me know.

The only other opportunity that I'm not sure will give me the same coverage I'm looking for is if the ` { role: 'DisclosureTriangle', name: 'Menu' },` exists or not.

Also, we are doing similar tests in `rh-disclosure` 

```
 const details = element.shadowRoot!.querySelector<HTMLDetailsElement>('details');
 expect(details?.open).to.be.false;
```

bennypowers: see latest. I gave a talk on this last year. here are the slides: https://bennypowers.dev/decks/semantic-component-testing/
here's the video https://youtu.be/iNQNmiJbKEU?si=pz9EO0xXb-DDa2MS

elements/rh-navigation-primary/test/rh-navigation-primary.spec.ts
```diff
@@ -69,32 +142,83 @@ describe('<rh-navigation-primary>', function() {
...
       describe('interactions', function() {
         describe('toggling hamburger menu', function() {
-          let details: HTMLDetailsElement | null | undefined;
           let summary: HTMLElement | null | undefined;
 
           beforeEach(async function() {
-            details = element.shadowRoot?.querySelector('#hamburger');
+            const details = element.shadowRoot?.querySelector('#hamburger');
             summary = details?.querySelector('summary');
             summary?.click();
```
bennypowers, line 176: replace these with keyboard presses i.e tab sequences

- avoids brittle shadow queries
- as a bonus, we gain keyboard a11y testing

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -410,11 +410,17 @@ export class RhNavigationPrimary extends LitElement {
   }
 
   async #openHamburger() {
+    if (!this._hamburger) {
+      await this.updateComplete;
+    }
     this._hamburger.open = true;
```
bennypowers, line 416: perhaps unrelated to this PR, if so spin out to a separate issue

this (and below) should not imperatively set open, instead, use the lit template. set private state on the class, and have that toggle the open attr or prop via the template. ensure state is synchronized in the toggle listener.

we should not use the dom api to modify elements that lit manages, as lit can get temperamental about that.

bennypowers (changes requested):

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: https://github.com/RedHat-UX/red-hat-design-system/issues/2408

elements/rh-navigation-primary/test/rh-navigation-primary.spec.ts
zeroedin: https://github.com/RedHat-UX/red-hat-design-system/issues/2408

PR #2357, opened 2025-05-06
feat(breadcrumb): add truncation

### LINKED ISSUE #2262: [feat] Improved responsiveness for `<rh-breadcrumb>`
### Description

In the initial issue to add breadcrumbs, there was a lot of discussion about having a truncated version of `<rh-breadcrumb>` for small screens. The Docs team (and possibly others) would like this feature for mobile implementation.

### Suggested solution

**Design:** `Responsive approach 2`in [this Figma mockup](https://github.com/RedHat-UX/red-hat-design-system/issues/1458)

### Screenshots

_No response_

### Example API

```HTML

```

### Additional context

_No response_

adamjohnson:
## What I did

1. Added a prop, `truncate`, to allow middle breadcrumb items to be temporarily replaced by a button.
2. Users can click the button to 1) show the middle breadcrumb buttons and 2) remove the button from the DOM.

### How truncation works

1. Requires the `truncate` boolean property
2. Only truncates when there are more than four breadcrumb items
3. Starts truncation at the second list item
4. Ends truncation at the third to last list item

## Testing Instructions

1. View the [DP demo](https://deploy-preview-2357--red-hat-design-system.netlify.app/elements/breadcrumb/demo/truncate/).
1. Ensure I am taking the right approach in the code.

## Notes to Reviewers

Based off of [this Figma](https://www.figma.com/proto/FSkPCjWTWK63QfSnX8rChZ/rh-breadcrumb?page-id=1%3A3&type=design&node-id=1-218&viewport=122%2C139%2C1&t=AmmdkOlWBdRh2fWu-1&scaling=min-zoom).

We can optionally add other properties to tell where to start and end truncation. The figma hints at this.

The Figma shows no truncation at LG and XL viewports. Ideally this could be handled by a Container Size Controller; although, last I looked, there was no JS API for container queries. Perhaps this could be handled by our ScreenSizeController. 

## To do's

- [x] Remove errant commits from this PR
- [x] Fix tests
- [x] Add new tests for `truncate` functionality

Closes #2262.

### DIFF
--- a/elements/rh-breadcrumb/rh-breadcrumb-lightdom.css
+++ b/elements/rh-breadcrumb/rh-breadcrumb-lightdom.css
@@ -17,123 +17,125 @@ rh-breadcrumb {
   & [hidden] {
     display: none !important;
   }
+
+  & > ol {
+    container-type: inline-size;
+    container-name: breadcrumbs-list;
+    font-size: var(--rh-font-size-body-text-sm, 0.875rem);
+    line-height: var(--rh-line-height-body-text, 1.5);
+    margin-block: 0;
+    padding-inline-start: 0;
+    & > li {
+      display: inline-flex;
+      margin-block-end: var(--rh-space-md, 8px);
+      padding-inline-end: var(--rh-breadcrumb-li-padding-inline-end, var(--rh-space-lg, 16px));
+      vertical-align: bottom;
+      align-items: center;
+      gap: var(--rh-space-lg, 16px);
+
+      &:not(:first-of-type):before {
+        display: inline-block;
+        background-color: var(--rh-color-icon-secondary);
+        mask-image: var(--_breadcrumb-caret-image-default);
+        mask-position: center center;
+        mask-repeat: no-repeat;
+        height: var(--rh-font-size-body-text-sm, 0.875rem);
+        width: var(--rh-font-size-body-text-sm, 0.875rem);
+        content: '';
+      }
+
+      &:first-child {
+        padding-inline-start: 0;
+        & > a {
+          margin-inline-start: 0;
+        }
+      }
+
+      &:last-child {
+        background-image: none;
+      }
+
+      & > a {
+        text-decoration: underline;
+        &:is(:hover, :focus, :active) {
+          color: var(--_breadcrumb-link-color-hover);
+        }
+
+        &:is(:focus, :focus-visible) {
+          border-radius: var(--rh-border-radius-default, 3px);
+          outline: 2px solid var(--_breadcrumb-link-focus-outline-color);
+        }
+
+        &:visited {
+          color: var(--_breadcrumb-link-color-visited);
+        }
+
+        &:visited:hover {
+          color: var(--_breadcrumb-link-color-visited-hover);
+        }
+
+        &[aria-current='page'] {
+          cursor: default;
+          pointer-events: none;
+        }
+      }
+
+      &:last-child,
+      & > a[aria-current='page'] {
+        color: var(--_breadcrumb-link-color-current-page);
+        text-decoration: none;
+      }
+
+    }
+
+    & > li,
+    & > li > a {
+      color: var(--_breadcrumb-link-color);
+    }
+  }
+
+  &[variant='subtle'] > ol > li {
+    &:not(:first-of-type):before {
+      background-color: var(--rh-color-icon-subtle, #707070);
+    }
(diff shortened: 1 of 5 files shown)


bennypowers:
Introducing a new attribute which turns container/responsive feature on and off seems excessive to me. What does `<rh-breadcrumb truncate>` mean when the viewport is 1200px wide? What does `$('rh-breadcrumb').removeAttribute('truncate')` mean when the container is 400px wide?

What if instead we used a ResizeObserver whose callback checks the elements' computed style for a style applied via a container query?

adamjohnson:
Thanks for the thoughtful feedback!

> Introducing a new attribute which turns container/responsive feature on and off seems excessive to me.

I've always seen this as an opt-in feature. Some users may be perfectly fine with breadcrumbs wrapping onto the next line at smaller viewports, and changing that behavior may lead to unwanted surprises.

> What does `<rh-breadcrumb truncate>` mean when the viewport is 1200px wide? 

That the breadcrumbs are truncated, awaiting your steady hand to click the button to show them all!

> What does `$('rh-breadcrumb').removeAttribute('truncate')` mean when the container is 400px wide?

Without the `truncate` attribute, the breadcrumbs will simply wrap, preserving the current default behavior regardless of viewport size.

> What if instead we used a `ResizeObserver` whose callback checks the elements' computed style for a style applied via a container query?

If this team decides to conditionally truncate breadcrumbs based off of screen size, I can look into using `ResizeObserver` for this functionality.

@coreyvickery and @marionnegp would love to get your feedback on this topic too.

bennypowers:
OK thanks for clarifying. If that's the intent, then it makes sense to add a flag

elements/rh-breadcrumb/rh-breadcrumb.ts
```diff
@@ -49,16 +60,80 @@ export class RhBreadcrumb extends LitElement {
...
+    if (this.#list.children.length < 5) {
+      return;
+    }
+
+    const middleItems = this.#list.querySelectorAll('li:nth-child(n+2):nth-last-child(n+3)');
+    for (const item of middleItems) {
+      item.setAttribute('hidden', 'true');
+    }
+
+    const [truncationBtnString] = truncationBtn.strings;
```
bennypowers, line 108: Accessing the strings property here doesn't seem proper.

It might be better here to render() to a throwaway document fragment, or use a template element.

Even better would be to render this button in the shadow root

bennypowers (changes requested):

coreyvickery:
> @coreyvickery and @marionnegp would love to get your feedback on this topic too.

@adamjohnson What you outlined sounds good to me.

elements/rh-breadcrumb/rh-breadcrumb-lightdom.css
```diff
@@ -101,3 +101,39 @@ rh-breadcrumb[variant='subtle'] > ol > li:last-child,
 rh-breadcrumb[variant='subtle'] > ol > li > a[aria-current='page'] {
   color: var(--_breadcrumb-link-color-current-page-subtle);
 }
+
+[hidden] {
```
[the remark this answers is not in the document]
adamjohnson: Great suggestion. Done in [f4d60e1](https://github.com/RedHat-UX/red-hat-design-system/pull/2357/commits/f4d60e14b3faa5d0f2b0775743ae49e7d5df0427).

coreyvickery (approved):
@adamjohnson If everyone else is good, I'm good.

IEEIGIG. Not as flashy as LGTM, but whatever.

bennypowers (changes requested):
Needs tests

elements/rh-breadcrumb/rh-breadcrumb.ts
```diff
@@ -49,15 +70,63 @@ export class RhBreadcrumb extends LitElement {
...
+    render(truncationBtn, container);
+    middleItems[0].before(container);
+  }
+
+  #onTruncationClick(event: Event): void {
+    let listItems: NodeListOf<Element>;
+    if (!isServer
+      && this.truncate
+      && isTruncateButtonDescendant(event.target)
+      && (listItems = this.querySelectorAll(':scope > ol > li'))) {
```
adamjohnson, line 124: What's the browser support for `:scope`?

bennypowers (approved):
Little grouped tags, masked

PR #2360, opened 2025-05-07
fix(navigation-primary): item menu padding

zeroedin:
## What I did

1. To support the [Account dropdown from design](https://www.figma.com/design/Ob5kpQip4qdSFJ7gTXSP2r/Unified-Navigation?node-id=4153-58012&m=dev), we need the ability to remove the padding from the `rh-navigation-primary-item-menu` container.
2. Added `--rh-navigation-primary-item-menu-padding-block-end`

## Testing Instructions

1. View deploy preview demo
2. Add a new CSS rule  `rh-navigation-primary-item { --rh-navigation-primary-item-menu-padding-block-end: 0; }` 
3. The item dropdown menu should have zero bottom padding.

## Notes to Reviewers

I think this is likely a `minor` change due to the addition of the new functionality, but because we are releasing nav in a controlled way right now and know our consumers, maybe we can squeeze this as a patch-level concern. Thoughts?  Will add changeset when we resolve this question.

### DIFF
--- a/elements/rh-navigation-primary/rh-navigation-primary-item-menu.css
+++ b/elements/rh-navigation-primary/rh-navigation-primary-item-menu.css
@@ -5,18 +5,16 @@
 #container {
   color: var(--rh-color-text-primary);
   background-color: var(--rh-color-surface);
-  padding:
+  padding-inline: var(--rh-space-lg, 16px);
+  padding-block:
     var(--rh-space-lg, 16px)
-    var(--rh-space-lg, 16px)
-    var(--rh-navigation-primary-item-menu-padding-block-end, var(--rh-space-2xl, 32px))
-    var(--rh-space-lg, 16px);
+    var(--rh-navigation-primary-item-menu-padding-block-end, var(--rh-space-2xl, 32px));
 
   @container (min-width: 768px) {
-    padding:
-      var(--rh-space-lg, 16px)
-      var(--rh-space-lg, 16px)
-      var(--rh-navigation-primary-item-menu-padding-block-end, var(--rh-space-2xl, 32px))
-      var(--rh-space-lg, 16px);
+    padding-inline: var(--rh-space-2xl, 32px);
+    padding-block:
+      var(--rh-space-2xl, 32px)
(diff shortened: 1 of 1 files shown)


zeroedin:
> Some thoughts on how we might patch this without adding any new CSS:
> 
> * How tight are the constraints on the content which needs this feature? In other words, is this a variation of the "disclosure host styles when there's a jump links inside" problem?

In the future, yes/maybe, given @MagRat04 work on `rhx-account-dropdown` (working title), that would be the content region of the account dropdown specifically. Now, I can't speak to his work and alignment with what we see design-wise in the rh-navigation-primary figma. But my general understand is this is or could be a case of this content is inserted into this parent component, now we need the parent to do something different then normal, maybe preferably without any additional configuration.


> * What if we invert things and remove the padding on the container for everyone, but then add margins to the content? How annoying would that be?

I'm hesitant on this.   That would mean for all dropdowns, whereas this change is really for just 1.   I feel like flipping it is a bad idea.  The default is you have padding.  The account dropdown right now is the exception.

zeroedin:
The padding is also ONLY removed in the "unauthed" state.

Authed:
image

Unauthed:
image

The original slot's intent was to maintain the padding across all the dropdowns, which was also because those dropdowns hadn't been designed when the component was created.

I'm wondering, given @MagRat04's work, if we can use a check to see if the `rhx-account-dropdown` is slotted as we suggest in `rh-disclosure` for `rh-jump-links` and then remove the bottom padding.

For RHDC purposes, because they likely won't use `rhx-account-dropdown,` I wonder if they would find it acceptable to use `rhx-account-dropdown` as a custom-element wrapper like div to trigger the same outcomes... 🤔   That said, I also wonder the barriers to adoption of `rhx-account-dropdown` across the board.

bennypowers:
I'd prefer if we get some clarity on the allowed content from our stakeholders before proceeding with this

markcaron:
@zeroedin @bennypowers where are we at on this fix? Will it make it into Diglett?

bennypowers:
@markcaron no we can't move on this until PMs come in with a clearly delineated business case e.g. "when there's a login avatar in the menu, remove the padding" or something like that

PR #2365, opened 2025-05-08
fix(dialog): scrollbar state after closing

brianferry:
Adding an event listener to the native dialog element to handle if the dialog is cancelled outside of the context of the  element

## What I did

1.  I added an event listener to the dialog in the `firstUpdated()` function that calls a function called `#onNativeDialogCancel` when the dialog element sends a cancel event.
2. The native dialog cancel prevents the default behavior and instead calls out components `cancel` method directly.

## Testing Instructions

1.  Open the demo site
2. Go to the `Lots of Content` page
3. Open the dialog
4. Tab out of the dialog
5. Click the address bar
6. Tap the `ESC` key twice
7. The scrollbar should be enabled.

### DIFF
--- a/.changeset/seven-dragons-jam.md
+++ b/.changeset/seven-dragons-jam.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-dialog>`: add event listener to native dialog close event to enable scrollbar on cancelling the dialog.
+  
--- a/.changeset/seven-dragons-jam.md
+++ b/.changeset/seven-dragons-jam.md
@@ -2,5 +2,5 @@
 "@rhds/elements": patch
 ---
 
-`<rh-dialog>`: add event listener to native dialog close event to enable scrollbar on cancelling the dialog.
+`<rh-dialog>`: The dialog element will now reset scrollbar when closed, even if it was not in focus at the time of closure.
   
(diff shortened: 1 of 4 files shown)


.changeset/seven-dragons-jam.md
bennypowers, line 1: Changesets should not describe internal apis (event listener) but rather describe user-facing changes

elements/rh-dialog/rh-dialog.ts
```diff
@@ -124,6 +124,13 @@ export class RhDialog extends LitElement {
   disconnectedCallback() {
     super.disconnectedCallback();
     this.#triggerElement?.removeEventListener('click', this.onTriggerClick);
+    this.dialog?.removeEventListener('cancel', this.#onNativeDialogCancel);
```
bennypowers, line 127: Use lit template syntax instead of imperatively adding listener

elements/rh-dialog/rh-dialog.ts
```diff
@@ -253,6 +260,11 @@ export class RhDialog extends LitElement {
     }
   }
 
+  #onNativeDialogCancel(event: Event) {
```
bennypowers, line 263: Canyou add tests for this, ensuring g that our behaviour matches the native dialog apis? 

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog#handling_the_return_value_from_the_dialog

bennypowers (changes requested):

bennypowers (approved):
Nice work. Clean solution, nice tests

PR #2379, opened 2025-05-18
Docs/knobs/initial

bennypowers:
## What I did

1. knobs (more info tbd)

## Testing Instructions

1. /elements/accordion/


uxdot/uxdot-demo.css
adamjohnson, line 1: Screenshot showing inadequate color contrast on code elements inside tooltips in the knobs container

`<code>` elements inside the `#knobs` container have insuficient color contrast. It looks like [this rule](https://github.com/RedHat-UX/red-hat-design-system/blob/main/docs/styles/styles.css#L248) is breaking the color contrast here. Will probably have to add a few lines to the `uxdot-demo.css` to fix it.

uxdot/uxdot-knob-attribute.ts
```diff
@@ -0,0 +1,198 @@
...
+            .replace(/^\s*['"]([^'"]+)['"].*$/m, '$1'));
+
+    const isIconSet = this.name === 'icon-set' || (this.tag === 'rh-icon' && this.name === 'set');
+
+    return html`
+      <li data-name="${this.name}" @change="${this.#onChange}">
+        <label for="knob">
+          <code>${this.name}</code>
+          <rh-tooltip>
+            <rh-icon icon="information" set="ui" tabindex=0></rh-icon>
```
adamjohnson, line 61: VoiceOver is announcing the tooltip content strangely here ("Group"). It might be better to remove the `tabindex` and nest the `<rh-icon>` inside an interactive element (`<button>`, `<rh-button>`, `<a>`, etc).

adamjohnson (changes requested):
Some initial a11y feedback.

coreyvickery:
> what minimal changes need to be made here to the usability and design in order for you to be able to live with this?

@bennypowers I'd like to take at least one quick pass at design before the end of the week (which this week is Thursday).

uxdot/uxdot-knob-attribute.ts
bennypowers: How's [7218a28](https://github.com/RedHat-UX/red-hat-design-system/pull/2379/commits/7218a287f7202688ae189d4b6c8ffb8db0799db9) for that?

bennypowers:
this remains blocked by https://github.com/RedHat-UX/red-hat-icons/pull/22

cc @markcaron @adamjohnson

bennypowers:
https://github.com/RedHat-UX/red-hat-icons/releases/tag/v1.3.0

bennypowers:
maybe it's not in the files list for the icons repo package.json? in that case we need to add it to that list, then release a patch

uxdot/uxdot-knob-attribute.ts
```diff
@@ -0,0 +1,249 @@
...
+            ${option}
+            <rh-icon slot="icon"
+                     icon="${option}"
+                     set="${this.#iconSet}"></rh-icon>
+          </pf-option>`)}
+        </pf-select>` : isUnionType ? html`
+        <pf-select id="knob"
+                   data-kind="enum"
+                   value="${ifDefined(this.#values.get(this.name))}">
+          <pf-option .value="${null}">Choose a Value</pf-option>${options.map(option => html`
```
markcaron: For whatever reason it seems to be inserting in `variant="Choose a Value"` rather than `""` (null).

image

The actual `option` value is `"Choose a Value"`

image

markcaron: When toggling between its default state to `promo` and then selecting `Choose a Value`, it does some weird styling (see the image above, where there's margin before the heading).

markcaron (dismissed):

uxdot/uxdot-knob-attribute.ts
bennypowers: ok PTAL @markcaron when the DP finishes deploying. There's an unrelated bug with card where toggling the promo attribute on and off has strange effects, but that doesn't have to do with this pr

bennypowers:
@coreyvickery it's not particularly easy to change the height of iframes

coreyvickery:
> @coreyvickery it's not particularly easy to change the height of iframes

@bennypowers Would JavaScript have to be added in order to do that?

bennypowers:
> > @coreyvickery it's not particularly easy to change the height of iframes
> 
> @bennypowers Would JavaScript have to be added in order to do that?

To the demo partials. Not something I want to tackle in this PR, which has been open for a month and a half

If there aren't any **can't live without it** show-stoppers here, let's merge *now* and *iterate*.

coreyvickery:
@bennypowers Got it.

If it's easy to tackle the four bullets, I'm good with that and then we can iterate in Duduo.

adamjohnson:
@coreyvickery 

> If it's easy to tackle the four bullets, I'm good with that and then we can iterate in Duduo.

Four bullets tackled. PTAL.

coreyvickery (approved):
Good to go.

PR #2386, opened 2025-05-21
fix(tooltip): contrasting color-scheme

bennypowers:
## What I did

1. compute the contrasting "color-scheme" value for tooltips when they open

## Testing Instructions

1. pick a page with tooltips, maybe in a demo or a code block, preferably with themable content like `<code>`
2. toggle color schemes back and forth
3. open the tooltip and observe that the themable content is legible

## Notes to Reviewers
This came up when working on knobs #2379.

This all happens in the `open()` method, so it should be client-side only and not cause ssr mismatches.

also NB that we neglected to delete the deprecated properties when releasing cubone, so we should prioritize that for the next breaking change, whenever that might be

### DIFF
--- a/elements/rh-tooltip/rh-tooltip.ts
+++ b/elements/rh-tooltip/rh-tooltip.ts
@@ -91,6 +91,7 @@ export class RhTooltip extends LitElement {
   });
 
   #initialized = false;
+  #style?: CSSStyleDeclaration;
 
   get #content() {
     if (!this.#float.open || isServer) {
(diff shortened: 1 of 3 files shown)


elements/rh-tooltip/rh-tooltip.ts
```diff
@@ -113,6 +114,10 @@ export class RhTooltip extends LitElement {
   override render() {
     const { alignment, anchor, open, styles } = this.#float;
 
+    const scheme = this.#style?.colorScheme ?? 'light';
```
zeroedin, line 117: Oh ok, well then. If this is this easy, then I need to revisit primary-nav border-width issue. I think adding this simple js is an easy extra lines of code trade off.  Think we should we abstract this and implement as part of the @themeable decorator?

adamjohnson (approved):
Low-contrast Gone, Tooltip Magnified

If you want to go the extra mile and have `<code>` elements have a distinct background color on dark schemes (as they do on light schemes), replace line 250 in `docs/styles/styles.css` with the following:

```css
background:
      light-dark(oklch(from var(--rh-color-surface) calc(l - 0.05) c h),
        oklch(from var(--rh-color-surface) calc(l + 0.1) c h));
```

PR #2389, opened 2025-05-21
fix(code-block): default content for fabs

### LINKED ISSUE #2388: [bug] `<rh-code-block>`  action fabs should have default content
### Describe the bug

Users who set the `actions` attr on codeblock might expect things to "just work". at the moment, if they don't also slot in content to label the actions, the end user will have a confusing experience with empty tooltips.

### Screenshots

Image

### Steps to reproduce

see html below

### Expected behaviour

codeblock should instead provide default content for those slots, but still allow customization for i18n purposes

### Element HTML

```HTML
<rh-code-block actions="copy wrap">
  <script type="text/plain">Hello</script>
</rh-code-block>
```

### Operating System (OS)

Windows, Linux, macOS, iOS

### Browser

_No response_

bennypowers:
Closes #2388 

## What I did

1. add default content for code-block fabs
2. document the actions attribute

## Testing Instructions

1. see issue, use example html
2. fabs should have reasonable english content

### DIFF
--- a/elements/rh-code-block/rh-code-block.ts
+++ b/elements/rh-code-block/rh-code-block.ts
@@ -84,11 +84,30 @@ export class RhCodeBlock extends LitElement {
 
   static styles = [style];
 
+  /**
+   * Space- or comma-separated list of code block action buttons to display, containing either 'copy', 'wrap', or both.
+   * 'copy' adds a button that copies the text content to the clipboard. 'wrap' adds a button that toggles line wrap.
+   *
+   * To override the default labels, e.g. for purposes of internationalization, use the
+   * `action-label-copy` and `action-label-wrap` slots. Each slot may receive two elements,
+   * one for the action's default state (e.g. "Copy to clipboard"),
+   * and one for the actions alternative state, e.g. "Copied!".
+   * The active-state element must have the attributes `hidden data-code-block-state="active"`
+   *
+   * @example html```
+   *          <rh-code-block actions="copy wrap">
+   *            <span slot="action-label-copy">Copy to Clipboard</span>
+   *            <span slot="action-label-copy" hidden data-code-block-state="active">Copied!</span>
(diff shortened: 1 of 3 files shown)


zeroedin (approved):
🥬 🧄 🍅 🍄 

Code looks good, Demo works well!

elements/rh-code-block/rh-code-block.ts
```diff
@@ -182,16 +195,20 @@ export class RhCodeBlock extends LitElement {
         <div id="actions"
              @click="${this.#onActionsClick}"
              @keyup="${this.#onActionsKeyup}">
-        <!-- <slot name="actions"> -->${this.actions.map(x => html`
+        ${this.actions.map(x => html`
           <rh-tooltip>
-            <slot slot="content" name="action-label-${x}"></slot>
+            <slot slot="content" name="action-label-${x}">${x === 'copy' ? html`
+              <span>Copy to Clipboard</span>
+              <span hidden data-code-block-state="active">Copied!</span>` : html`
```
adamjohnson, line 202: When listening to these via VoceOver in Safari, both of these text nodes seem to be copied over to the `<div role="status">` via the tooltip. Safari reads, "Copy to clipboard, Copied!" (even if the user did not activate the button). 

Same thing for the Toggle Word Wrap / Overflow tooltip/button.

The behavior in this PR is better than what's currently on uxdot (a completely unlabeled tooltip & button), so, if this needs merged ASAP I'm not going to block it. 

But, is there a better way to communicate state on button activation?

Also, it's probably a good idea to explore adding an aria-labelledby to the `<button>` if it does not already have an accessible label. The changes mentioned above _might_ make this point moot.

marionnegp (approved):

elements/rh-code-block/rh-code-block.ts
bennypowers: #2400

PR #2390, opened 2025-05-21
fix(code-block): autoload tooltip for fabs

### LINKED ISSUE #2121: [bug] `rh-code-block` does not load `rh-tooltip` when actions are used
Documentation does not specify when using `rh-code block` that `rh-tooltip` will need to be loaded as well when using the actions.  If `rh-tooltip.js` isn't loaded, then the actions of `copy` and `wrap` do not function as intended.

See #2120 as an example of this fixed manually for uxdot documentation pages.

The question is should `rh-code-block` dynamically load `rh-tooltip` if actions are specified?

bennypowers:
Closes #2121

## What I did

1. lazy load tooltip client side when actions are used

## Testing Instructions

1. load the actions demo, tooltips should work

### DIFF
--- a/.changeset/thin-baboons-do.md
+++ b/.changeset/thin-baboons-do.md
@@ -1,0 +1,4 @@
+---
+"@rhds/elements": patch
+---
+`<rh-code-block>`: automatically load tooltip when action buttons are added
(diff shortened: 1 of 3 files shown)


zeroedin (approved):
Loading Given Tootltips Masterfully

PR #2395, opened 2025-05-21
docs(navigation-primary): use rh-icon profile instead of rh-avatar

zeroedin:
## What I did

1. In the lastest designs the `<rh-icon icon="profile" set="ui">` is used instead of `<rh-avatar>`

## Testing Instructions

1. View deploy preview, all demos should reflect this change

## Notes to Reviewers

This is more a question of when we use `<rh-avatar>` moving forward.

It looks like the authenticated state of `<rh-avatar>` may be restricted to internal use in the `<rhx-account-dropdown>` instead of a first-class citizen on the nav bar as it was prior primary nav designs. Instead, we are changing the text: Login to Account, and the icon stays the same. For compact viewport sizes, the profile icon remains with no other indication of the login status. @marionnegp @coreyvickery Is this correct?

Related to #2950


marionnegp:
The icon that's being used in the current nav was different from the default icon for rh-avatar, so I had flagged that as something to change during RHDC's QA. Is it possible to slot in a different default icon in rh-avatar when the user is signed out?

zeroedin:
> On the rh-avatar side, is it possible to slot in a different default icon when the user is signed out? That might allow `rh-avatar` to be used as a first-class citizen in both states.

Could we update which icon represents `plain` in `rh-avatar`? Would that work for you?

https://ux.redhat.com/elements/avatar/demos/#demo-plain

bennypowers:
> > On the rh-avatar side, is it possible to slot in a different default icon when the user is signed out? That might allow `rh-avatar` to be used as a first-class citizen in both states.
> 
> Could we update which icon represents `plain` in `rh-avatar`? Would that work for you?
> 
> https://ux.redhat.com/elements/avatar/demos/#demo-plain

this is preferable to having two different instances of what "avatar" means, one for nav-1, and a different one for everything else. There could even perhaps be a slot for the default icon in avatar

bennypowers:
this should maybe target diglett branch?

zeroedin:
> this should maybe target diglett branch?

This might actually be closed if we plan to make changes to `rh-avatar` instead.  This PR was in response to how nav-1 is currently implemented on RHDC, if that is also subject to change this PR isn't valid at this time.  Going to move it to draft for now.

zeroedin:
@marionnegp @bennypowers are we suggesting the path forward with this is a slot addition to `rh-avatar`

https://github.com/RedHat-UX/red-hat-design-system/blob/adabbe6653c65a5c1e76fa5c44f1a85abbf4e53a/elements/rh-avatar/rh-avatar.ts#L84-L92

Whereas if `src` property isn't set, a `slot[name="icon"]` with the current `<svg>` as the default output?

```html
        <canvas part="canvas"></canvas>` : this.src ? html`
        <img src="${this.src}" role="presentation" part="img">` : html`
        <slot name="icon">
          <svg xmlns="http://www.w3.org/2000/svg" style="enable-background:new 0 0 36 36" viewBox="0 0 36 36" role="presentation" part="img" id="default">
            <path d="M0 0h36v36H0z" class="st1"/><path d="M17.7 20.1c-3.5 0-6.4-2.9-6.4-6.4s2.9-6.4 6.4-6.4 6.4 2.9 6.4 6.4-2.8 6.4-6.4 6.4z" class="st3"/>
            <path d="M13.3 36v-6.7c-2 .4-2.9 1.4-3.1 3.5l-.1 3.2h3.2z" class="st2"/>
            <path d="m10.1 36 .1-3.2c.2-2.1 1.1-3.1 3.1-3.5V36h9.4v-6.7c2 .4 2.9 1.4 3.1 3.5l.1 3.2h4.7c-.4-3.9-1.3-9-2.9-11-1.1-1.4-2.3-2.2-3.5-2.6s-1.8-.6-6.3-.6-6.1.7-6.1.7c-1.2.4-2.4 1.2-3.4 2.6-1.7 1.9-2.6 7.1-3 10.9h4.7z" class="st4"/>
            <path d="m25.9 36-.1-3.2c-.2-2.1-1.1-3.1-3.1-3.5V36h3.2z" class="st2"/>
          </svg>
        </slot>
        `}
```

bennypowers:
Best: just use Avatar, without changing the design, OR decide that avatar legitimately has this new variant (where else would it be used?) 
Second best: use private CSS vars etc to modify avatar without adding any new public apis
Distant third: new slots, parts, etc
Avoid strenuously: make a new element

PR #2400, opened 2025-05-22
fix(tooltip): better sr support

bennypowers:
## What I did

1. do a better job of calculating text content when announcing with the tooltip.

## Testing Instructions

1. see code block and tooltip demos. codeblock in particular see in #2389


adamjohnson (approved):
Live Guidance Told Meaningfully

PR #2405, opened 2025-05-27
fix(navigation primary): secondary items fit content at

zeroedin:
## What I did

1. First pass at secondary item dropdowns `menu-fit-content-at` attributes. Allows secondary items to change when they move from mobile full-screen to `fit-content` width

### TODO:

- [ ]  Figure out the separation of concern between `compact` (padding changes) and these new breakpoints
- [ ] Are there other issues with this approach not thoughtfully processed yet?
- [ ] Likely rebase against `staging/diglett` due to this change being a `minor`
- [ ] Bring up change to breakpoints names/values that were missed in review for `hide-at.`  Breaking? If so, we may want to revert that commit and open an issue for the change on the next available major.

## Testing Instructions

1.

## Notes to Reviewers

ts.


bennypowers:
Can we do this with CSS only? Adding SSR complexity to this element specifically is likely to have downstream implications

Issue #2148, opened 2025-01-28
[feat] interactive states for `<rh-tag>`


marionnegp:
### Description

Docs for the linked `<rh-tag>` show [states for linked tags](https://ux.redhat.com/elements/tag/style/#interaction-states), but they have to be implemented.

We also need to consider instances where `<rh-tag>` acts as a button, and we may need to show a different affordance for interactivity. (For example, a tag in My Red Hat can trigger a popover.) [PatternFly's tag demos](https://ux.redhat.com/elements/tag/style/#interaction-states) currently show clickable tags and linked tags that use the same states.

Additional updates:
- When tag is used as a button, a space bar should be able to trigger it. Currently it's trigger by the enter key only.

### Suggested solution

- [ ] Implement states for linked tags
- [ ] Design clickable tag states
- [ ] Fix keyboard interactivity if it's used as a button

### Example API

```HTML

```

markcaron:
Since we currently have linkable `<rh-tag>` elements, I'm thinking it wouldn't be a stretch to provide the option for clickable through a similar approach.

### Linkable:
`<rh-tag href="#">`

### Clickable:
`<rh-tag action="{namedAction}">`

Where `{namedAction}` could be a string like `popovertarget="popoverID"`

The attributes would be an either-or situation, as you couldn't have both at the same time. One would take precedence.

coreyvickery:
@markcaron @marionnegp What do you think of these state explorations for link vs. button?

<img width="882" alt="Image" src="https://github.com/user-attachments/assets/2a705714-92d3-4de2-b6b5-cf2b446078f5" />

marionnegp:
@coreyvickery, tag as a button looks good to me!

Does tag as a link need a hover state? The typical hover state for underlined inline links would be for the underline to move down a couple pixels, but I don't think that works as well here.

adamjohnson:
`<button>`'s can have a [`disabled` state](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled). Would be nice to include design specs for `disabled` as well. 

Keep in mind WCAG states that `disabled` elements do not need to meet color contrast guidelines (see [incidental](https://www.w3.org/TR/WCAG21/#contrast-minimum)).

coreyvickery:
@marionnegp @adamjohnson For the `link` variant, I got rid of the underline on `default`, but it could appear on `hover` and `active` states. Do you think that would be enough to distinguish?

[...]

PR #2407, opened 2025-05-28
feat(tag): add interactive states

adamjohnson:
## What I did

1. Add interactive states to `<rh-tag>` (aka `<a>` ~~and `<button>`~~).
2. See #2148 

## Testing Instructions

1. View the Tag [links demo](https://deploy-preview-2407--red-hat-design-system.netlify.app/elements/tag/demo/links/) in the DP.
2. ~~View the Tag [Buttons demo](https://deploy-preview-2407--red-hat-design-system.netlify.app/elements/tag/demo/buttons/) in the DP.~~
3. ~~View the [updated docs](https://deploy-preview-2407--red-hat-design-system.netlify.app/elements/tag/) and compare them to the [Figma](https://www.figma.com/proto/SjE4niXxxm9WJzsVThRXFw/Element---Tag-updates?page-id=1%3A3&node-id=3-2834&p=f&viewport=-8588%2C94%2C1&t=CoGxWuvaQYPmKBgP-1&scaling=scale-down&content-scaling=fixed).~~
4. Ensure I implemented linked tags correctly. 

## Notes to Reviewers

~~Do I need `formDisabledCallback`/`formStateRestoreCallback` here?~~

~~There is little visual affordance between a default tag and a `button` tag. Is this acceptable?~~

## To Do's

- [x] Add updated tag docs
- [x] Add changeset
- [x] Update docs in Figma
- [x] Update docs on PR to reflect no `<rh-tag>` button variant


elements/rh-tag/rh-tag.ts
```diff
@@ -38,26 +39,48 @@ import styles from './rh-tag.css';
...
+  /** The size of the tag. */
   @property() size?: 'compact';
 
   /** optional href for linked tag. */
   @property() href?: string;
 
-  /** The color of the label. */
+  /** Make the tag an HTML button element. */
+  @property({ type: Boolean, reflect: true }) button = false;
+
```
bennypowers, line 65:
```suggestion
```

Just use `type="button"`

elements/rh-tag/rh-tag.ts
```diff
@@ -38,26 +39,48 @@ import styles from './rh-tag.css';
...
+  /**
+   * Optional type for button tag.
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type
+   */
+  @property() type?: 'button' | 'submit' | 'reset' = 'button';
```
bennypowers, line 70:
```suggestion
  @property() type?: 'link' | 'button' | 'submit' | 'reset' = 'link';
```

elements/rh-tag/rh-tag.ts
```diff
@@ -38,26 +39,48 @@ import styles from './rh-tag.css';
 export class RhTag extends LitElement {
   static readonly styles = [styles];
 
+  static readonly formAssociated = true;
```
bennypowers, line 42: this is tricky...

if it's a link, it shouldn't be a FACE. i.e. it should only be a FACE when it's a button

but `formAssociated` is a static property, it applies to all instances of the element, regardless of their particular attributes. This is a platform limitation that we can't work around, and I'm unaware of any proposals to change that.

If you're not comfortable with links, or even UI-only tags (i.e. pills) potentially being submitted with a form, then we need to send this feature back to the drawing board

```html
<form id="tagForm">
  <rh-accordion>
    <rh-accordion-header>Details<rh-tag slot="accents">New</rh-tag></rh-accordion-header>
    <rh-accordion-panel><label>Name <input name="name"></label></rh-accordion-panel>
    <rh-accordion-header>Payment<rh-tag slot="accents">Special</rh-tag></rh-accordion-header>
    <rh-accordion-panel><label>Payment info <input name="payment"></label></rh-accordion-panel>
  </rh-accordion>
  <rh-button type="submit">Submit</rh-button>
</form>

<script type="module">
  import '@rhds/elements/rh-accordion/rh-accordion.js';
  import '@rhds/elements/rh-button/rh-button.js';
  import '@rhds/elements/rh-tag/rh-tag.js';
  const tagForm = document.getElementById('tagForm');
  tagForm.addEventListener('submit', function(e) {
    e.preventDefault();
    console.log(
      e.target.elements,
      new FormData(e.target),
    );
  })
</script>
```

image

As you can see, the elements are enumerated in the form, which could be confusing and lead to errors when iterating over the elements

However, if the author omits the `name`, it appears they won't be submitted. But that could easily happen if an author tries to use tag for an `<a name>`-style anchor

bennypowers (changes requested):

elements/rh-tag/rh-tag.ts
adamjohnson: > As you can see, the elements are enumerated in the form, which could be confusing and lead to errors when iterating over the elements

> If you're not comfortable with links, or even UI-only tags (i.e. pills) potentially being submitted with a form, then we need to send this feature back to the drawing board

Indeed, this is concerning. I can see developers using `<rh-tag>` inside forms for labeling or other design purposes. Having a regular `<rh-tag>` included in form submissions feels like a bug, and we'll likely hear about it from our users.

What's also potentially confusing about the button variant is that both the button variant and a normal (non-interactive) `<rh-tag>` look identical. The only visual differences for the button variant are a thicker border on hover and a focus ring on focus/active states. See this image:

A side by side comparison of regular tags and button tags, they look identical

So, should our next step be to remove the `<button>` functionality from `<rh-tag>`? This would require updates to the Figma documentation (and subsequently in this PR), in addition to component code changes.

In the meantime, users could implement interactive tags like this:

```html
<button>
  <rh-tag color="green">Click me</rh-tag>
</button>
```

We'd be relying on them to make good design and development choices.

Alternatively, they could use the `href` / link variant to handle their interactive needs (even if, technically, a `<button>` might be a more semantically appropriate element choice).

I'd like to hear from @coreyvickery and @marionnegp about these changes and where we might go from here. I know our design team has invested significant time and energy in these decisions, and I want to ensure their concerns are heard and considered.

marionnegp: If we remove the button functionality, would the engineer have to add the interaction styles manually to use `rh-tag` in `<button>`?

adamjohnson: Yes.

adamjohnson: After chatting with Marionne in Office Hours, I'm going to go ahead and remove the `<button>` functionality from `<rh-tag>`. Corey and Marionne are going to work on updated docs. When docs are ready, I can code them up in this PR.

In the interim, we could consider making rh-tags-as-buttons a pattern.

marionnegp:
@adamjohnson, I removed mentions of button tags from the docs, but I did add a link to a tag pattern page. Do you want me to add that page in this PR or spin up a new one?

.changeset/dry-moles-sell.md
bennypowers, line 1: update changeset

elements/rh-tag/rh-tag.ts
```diff
@@ -74,26 +79,49 @@ export class RhTag extends LitElement {
...
+  }
+
+  #onKeyDown(event: KeyboardEvent): void {
+    if (this.disabled && event.key === 'Enter') {
+      event.preventDefault();
+    }
+    if (this.disabled && event.key === ' ') {
+      event.preventDefault();
+    }
+  }
```
bennypowers:
```suggestion
```
presumably these can be removed if we're not implementing FACE

bennypowers (changes requested):

adamjohnson:
> Do you want me to add that page in this PR or spin up a new one?

@marionnegp Add it to this PR. Thanks!

bennypowers (approved):

Issue #847, opened 2023-03-24
[bug] <rh-cta> arrow/icon trailing CTA message should never appear on a line by itself


nikkimk:
### Describe the bug

This is to align the CTA with the design spec whereas the arrow  `should always be connected to at least one word on the same line`

[Long text labels](https://ux.redhat.com/components/call-to-action/#long-text-labels)

![Screenshot 2023-03-24 at 9 01 14 AM](https://user-images.githubusercontent.com/13666/227528303-9f27a0b7-c963-47ff-b441-3d61cb543764.png)

This came up as a possible bug when I worked on #835, then we discussed it more during the catch up office hours see the demo `Force wrap to test arrow line-wrap` in the [demo](https://ux.redhat.com/components/call-to-action/demo/)

![image](https://user-images.githubusercontent.com/13666/227528742-a44b2229-ac02-47e7-a546-b5df2c448968.png)

The 2 bugs here are as we understand:

1. the arrow scales down when the flex box runs out of space
2. the arrow does not stay connected with the text when wrapped, rather centered and offset.  Potential issue with a fix could still be that the `arrow trailing the CTA message should never appear on a line by itself` and maybe why the decision to offset and center the arrow is currently against spec?

### Which Component?

rh-cta

### Steps to reproduce

1. Go to the [CTA demo](https://ux.redhat.com/components/call-to-action/demo/)
2. Notice that the arrow in the examples under "Force wrap to test arrow line-wrap" to not fall inline with the text.

### Expected behaviour

The arrow should appear inline with the text and never on a line with itself

### Operating System (OS)

macOS

zeroedin:
Closed in https://github.com/RedHat-UX/red-hat-design-system/pull/840

PR #2416, opened 2025-06-02
fix(cta): let long CTA's wrap to a second line

### LINKED ISSUE #2412: [bug] `<rh-announcement>` does not wrap long ctas on small screens
### Describe the bug

On mobile size screens, long rh-cta text does not wrap and causes rh-announcement to extend longer than the viewport, causing text to appear out of frame and a horizontal scrollbar.

Demo here: https://codepen.io/alypilkons/pen/WbvrvVv?editors=1100

### Screenshots

Image

### Steps to reproduce

1. Go to: https://codepen.io/alypilkons/pen/WbvrvVv?editors=1100
2. Resize to a mobile view and look at rh-announcement element

### Expected behaviour

On small screens, rh-announcement should wrap text enough so that the banner does not extend longer than the viewport or cause any horizontal scrolling.

### Element HTML

```HTML

```

### Operating System (OS)

iOS, macOS

### Browser

Chrome

_No response_

adamjohnson:
## What I did

1. Fixes #2412.

## Testing Instructions

1. Open the [Image Position](https://deploy-preview-2416--red-hat-design-system.netlify.app/elements/announcement/demo/image-position/) demo for Announcement in the DP.
2. In DevTools, change the CTA text to be something longer, eg `Wild cats roam through a forest of uncertainty`
3. Resize your screen to be ~320px.
4. Note how the CTA will wrap to a second line and not cause horizontal scrolling at any viewport below 767px.

## Notes to Reviewers

Changed this PR to keep functionality implemented in #847 while allowing an escape hatch to fix horizontal scrollbars if needed (as seen in Annoucnement).

Reviewers should ensure the changeset is appropriate for this change.


adamjohnson:
@bennypowers See "Notes to reviewers". tl;dr, with this change, an arrow could appear on its own line. IMO this is the lesser of two evils. 

It doesn't look like there's any reliable way in CSS to keep word(s) + icon on the same line without having horizontal scrollbars.

adamjohnson:
> Is the test case supplied by op within the guidelines?

Yes. Default CTA character count guideline is [55 characters](https://ux.redhat.com/elements/call-to-action/guidelines/#character-count). Aly's example is 50 characters: `Learn More about adipisicing elit adipisicing elit`

> Can we add an exception for ctas inside announcements...?

We could. But this issue really exists inside `<rh-cta>`, not `<rh-announcement>`.

IMO it would be better to fix this issue at the root component—in `<rh-cta>`.

Here's what I would propose:

`rh-cta.css`:

```diff
#container {
  position: relative;
-  white-space: nowrap;
+  white-space: var(--rh-cta-white-space, normal); 👈 Add a public var, set the fallback to the defaul (let text wrap)
  color: var(--_color);
  ...
}
```

Then, if people say, "the icon is on it's own line!" we say, "Do this:"

```css
rh-cta#special-snowflake {
  --rh-cta-white-space: nowrap;
}
```

PR #2428, opened 2025-06-06
refactor(navigation-primary): simplify gradient implementation

zeroedin:
## What I did

1. Replaced `mask` rule usage and the weird padding adjustments with a more straightforward implementation of linear-gradient on padding-box and conic-gradient on border-box on the psuedo :before element background.
2. Broke gradient into individual parts to reuse to construct the conical gradient.
3. Border radius now is easily inherited from the summary

## Testing Instructions

1. View Deploy Preview

## Notes to Reviewers

 - Styles shouldn't visually change; this should only affect internal styles.


elements/rh-navigation-primary/rh-navigation-primary-item.css
```diff
@@ -229,29 +223,36 @@
...
+          &:is(:active, :active:hover) {
+            &:before {
+              border: var(--rh-border-width-md, 2px) solid transparent;
+            }
+          }
 
+          &:is(:active, :active:hover, :hover, :focus-visible) {
             &:before {
-              --_mask: conic-gradient(#000000 0 0);
+              background:
```
bennypowers, line 252: might be worth naming this as a --_var, just for consistency with --_linear-gradient

elements/rh-navigation-primary/rh-navigation-primary-item.css
```diff
@@ -297,14 +297,14 @@
         }
 
         &[open] {
-          /* stylelint-disable-next-line no-descending-specificity */
-          summary {
-            background-color: var(--_background-color);
-
+          & summary {
+            /* stylelint-disable-next-line no-descending-specificity */
```
bennypowers, line 301: a why comment would be appreciated

bennypowers (approved):
LGTM. left some optional comments.

elements/rh-navigation-primary/rh-navigation-primary-item.css
zeroedin: Instead of a why, I refactored to address the underlying issue.

zeroedin: Could you give this a good once-over again? Was a significant change.

bennypowers (approved):

PR #2430, opened 2025-06-06
feat(skeleton): add `<rh-skeleton>`

### LINKED ISSUE #2108: [feat] Create `<rh-skeleton>` component/pattern
### Description

Does RHDS have any plans on implementing a skeleton loader, similar to https://www.patternfly.org/components/skeleton

### Acceptance Criteria

- [ ] Design done
- [ ] Development done

### Images

_No response_

### Design file

_No response_

### Other resources

_No response_

adamjohnson:
## What I did

1. Added `<rh-skeleton>`.

## Testing Instructions

1. Load up the following demos:
    * [Default](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/demo/)
    * [Text](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/demo/text/)
    * [Rectangle](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/demo/rectangle/)
    * [Circle](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/demo/circle/)
    * [Square](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/demo/square/)
    * [Color Context](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/demo/color-context/)
1. Ensure sizes match what's in the [Figma](https://www.figma.com/design/xtlEPL22Eaig7iZRk5cLBQ/rh-skeleton?node-id=2061-768&m=dev).

## Todo

- [x] Docs

## Notes to Reviewers

Docs should reflect that this component is primarily lightdom based. They should also note that the width and height of this element can be customized to fit the user's needs, eg `rh-skeleton { inline-size: 55px; block-size: 44px; }`.

Closes #2108.


coreyvickery (approved):
@adamjohnson Looks good. I think Marionne will tackle the docs when she returns.

bennypowers (changes requested):
I'm going to want to review this more in depth when i get some down time in .cz or after my trip, so this comment is just a placeholder for later review

adamjohnson:
@bennypowers would love your feedback on this when you have a couple free moments. TY!

marionnegp:
@adamjohnson, here are the [skeleton docs](https://www.figma.com/design/ygvU8UVhRiabE6EZlMIH84/Element---Skeleton?node-id=1-3&m=dev&t=obhrjr9xYWE8Ox82-1). Feel free to leave any feedback or questions if you spot anything!

elements/rh-skeleton/rh-skeleton-lightdom.css
```diff
@@ -0,0 +1,122 @@
...
+}
+
+@keyframes shimmer {
+  100% {
+    transform: translateX(100%);
+  }
+}
+
+rh-skeleton[size='xs'] {
+  block-size: calc(var(--rh-length-sm, 6px) * 3);
```
bennypowers: xs => sm?
sm => xl - 3?
md => xl?
lg => xl + 3?


this is confusing, no? Do we even need a size attribute? can't users use css for presentation details like this?

elements/rh-skeleton/rh-skeleton.ts
```diff
@@ -0,0 +1,40 @@
...
+  @property({ reflect: true })
+  size: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2xl' | '3xl' | '4xl' | '5xl' | '6xl' | null = 'md';
+
+  render() {
+    return html`
+      <p class="visually-hidden">
+        <slot>
+            Loading...
+        </slot>
+      </p>
```
bennypowers:
```suggestion
      <slot class="visually-hidden">Loading...</slot>
```

bennypowers (changes requested):

marionnegp (changes requested):
Sorry for the late review! 

- [x] Can the gradient opacities be: 0, 0.05, 0.1, 0.05, 0? Not sure if the 0.05 is possible here. Colors and color stops are right though!
- [x] Square skeleton should use `--rh-border-radius-default`
- [ ] Got some changes to the text skeleton heights, if we keep the `size` attribute:
xl (is 30, should be 26)
2xl (is 36, should be 31)
3xl (is 48, should be 37)
4xl (is 64, should be 47)
5xl (is 80, should be 52)
6xl (is 96, should be 62)

elements/rh-skeleton/rh-skeleton.ts
adamjohnson: Looks like, in order for visually hidden to work per your suggestion, you can't target the slot directly since the slotted content is just text content and not an HTML element. 

So, I made it a `<span>` tag instead of a `<p>` tag and tightened up the whitespace. [1ac6160](https://github.com/RedHat-UX/red-hat-design-system/pull/2430/commits/1ac6160ef1c77a36129f80048651fe2f58ea753b)

elements/rh-skeleton/rh-skeleton-lightdom.css
```diff
@@ -0,0 +1,122 @@
...
+  block-size: var(--rh-length-xl, 24px);
+}
+
+rh-skeleton[size='lg'] {
+  block-size: 27px;
+}
+
+rh-skeleton[size='xl'] {
+  block-size: 26px;
+}
```
adamjohnson: @marionnegp Can you verify that this sizing / naming convention is correct? In the [FIgma](https://www.figma.com/design/xtlEPL22Eaig7iZRk5cLBQ/rh-skeleton?node-id=2061-780&m=dev) under "Using text placeholders" there's no `XL` table row. Should there be?

Arrows pointing to the using text placeholders 2xl and 3xl table rows

elements/rh-skeleton/rh-skeleton-lightdom.css
adamjohnson: Indeed, these values are non-standard because they include both the height of the actual characters + line height. See [the Figma](https://www.figma.com/design/xtlEPL22Eaig7iZRk5cLBQ/rh-skeleton?node-id=2061-780&m=dev).

The designers have specifically included sizing for this component in the design specs. Relying on third parties to get these specs correct on their own seems like wishful thinking. My vote would be to keep the `size` attribute. But, I'm not diametrically opposed to removing it either.

Indeed, users can override properties in `rh-skeleton`'s lightdom if they should choose to do so.

elements/rh-skeleton/rh-skeleton-lightdom.css
```diff
@@ -35,43 +35,43 @@ rh-skeleton {
...
 }
 
 rh-skeleton[size='5xl'] {
-  block-size: var(--rh-length-5xl, 80px);
+  block-size: 52px;
 }
 
 rh-skeleton[size='6xl'] {
-  block-size: var(--rh-length-6xl, 96px);
+  block-size: 62px;
```
bennypowers: This is still confusing

We have length tokens for a reason, why are we intentionally ignoring them? This breaks user (ie downstream developer) expectations. XL should be the same thing in all cases

marionnegp: I based the text token heights off of the amount of space a text block would take up, which I've found was bigger than the font size. I believe PF did the same thing. This resulted in some dimensions that aren't tied to our tokens.

marionnegp: PF calculates their text skeleton heights like this: 
`<div class="pf-v6-c-skeleton pf-m-text-4xl"><span class="pf-v6-screen-reader">Loading text content</span></div>`

@bennypowers, do you think that makes more sense? This is how I got to the text skeleton heights. I did round to the closest whole number, but that can be adjusted if the above approach feels more scalable.

Would we still need a different naming convention? Other than using arbitrary numbers in place of t-shirt sizing, which we don't really do, there wasn't a good example of other naming conventions for size variants like this.

bennypowers: it would make sense to me if the "text-like" skeletons used tshirt sized based on font tokens, with some behind-the scenes adjustments to "look like" font sizes, and if the "graphics-like" skeletons used tshirt sizes based directly on length tokens. 

this might introduce some internal inconsistencies, like having a size 2xl circle composed with a size md rectangle, or whatever it ends up being

but i think would be the easiest to explain and understand

and those tradeoffs *must* be clearly and explicitly documented in the source code

marionnegp: I'm good with updating the shape skeleton naming to reflect our length tokens, especially since their sizes won't match up with the heights of a text skeleton.

_Some_ of the text skeleton tshirt sizes do match with their related body text token, but this does break down as we associate heading text sizes with a skeleton. What do you mean by "with some behind-the scenes adjustments to 'look like' font sizes"?

bennypowers: Let's say we decided that the tshirt size's corresponding length token covers the vertical distance from the bottom of the descender to the top of the ascender in the digraph "bp". Now let's say that from design considerations, we want the actual height of the skeleton for that tshirt size to end from the top to the bottom of the letter "o" in "bop". As long as the calculations which do that are private, and are explained with comments in the CSS, that's fine by me. The reason I'm okay with that is because we would still be preserving the semantic meaning of the T-shirt sizes, so even though the rectangle isn't exactly the same as 2XL, let's say, but in the sense that it is typographically equivalent to 2XL, it's fine.

Note that the examples I gave above are just arbitrary. I don't specifically mean those letters in those sizes, so those would have to be taken into consideration by design.

marionnegp: If I'm reading your comment correctly, maybe we need to make two sets of text skeletons, one for body tokens and one for heading tokens. Otherwise, there isn't a good way to have the skeleton t-shirt size correspond with the t-shirt sizes in our type tokens. The downside is that this feels kinda duplicative size-wise. A benefit would be that this should make it easier for designers and devs to find the right text skeleton without having to check our docs for sizing.

For the other shapes, it does feel weird to name the smallest built-in size `2xl` to correspond with the length token. Would it be preferable (or possible) to omit the `size` attribute for the shape skeletons only?

adamjohnson: > ..maybe we need to make two sets of text skeletons

API example:

```html
<rh-skeleton type="heading" size="2xl"></rh-skeleton>
<rh-skeleton type="body" size="md"></rh-skeleton>
```

> Would it be preferable (or possible) to omit the size attribute for the shape skeletons only?

Yes, this is possible (via CSS [`:not()`](https://developer.mozilla.org/en-US/docs/Web/CSS/:not)). We'd then just have to pick one default size for `type="circle|square|rectangle"`. 

WDYT @bennypowers?

bennypowers: It make sense to me. Will need good docs, to explain among other things that if the built in sizes aren't enough, there's other problems on the page

Not to beat a dead horse, but if this was keyed off the font size tokens, instead of attributes, it would *just work*, so long as the page was built using the tokens

marionnegp:
Here are [updated docs](https://www.figma.com/design/ygvU8UVhRiabE6EZlMIH84/Element---Skeleton?node-id=1-3&m=dev&t=69gYfXwDnOyksf7s-1), which show separate body text skeletons and heading skeletons that align better with our tokens. The shape skeleton uses a default size of `md`, which corresponds to `--rh-length-md`. I can work in this same branch to code these docs updates tomorrow.

marionnegp:
I've pushed docs updates, but I'm not sure how to get the knobs/playground on the "Overview" subpage, @adamjohnson.

elements/rh-skeleton/rh-skeleton-lightdom.css
adamjohnson: Handled in [c865b1a](https://github.com/RedHat-UX/red-hat-design-system/pull/2430/commits/c865b1afa7a86435648608b1bd41a182a51165f1).

adamjohnson:
I've updated the component to include `type="body-copy"` and `type="heading"` + associated styles and token values. PTAL for design and dev review.

@bennypowers is there anything we need to do to get the knobs/playground on the [Overview page](https://deploy-preview-2430--red-hat-design-system.netlify.app/elements/skeleton/#overview) in docs?

marionnegp (approved):

Issue #1329, opened 2023-11-09
[feat] `<rh-sidenav>` element


markcaron:
We need a new Side Navigation component for use on documentation (like our [ux.redhat.com](https://ux.redhat.com)) or web applications.

Based on prior art, we have 2 quite similar designs:

- [PatternFly's Navigation component](https://www.patternfly.org/components/navigation)
- [Our UX dot documentation's side nav](https://ux.redhat.com)

The priority is set to `high` because our current side nav on UX dot poses an inaccessible anti-pattern with it's checkbox-based hamburger menu.

markcaron:
Worth considering whether this follows the other Navigation naming as `<rh-navigation-side>`

wesleymiles:
Chiming in to say that the docs.redhat.com team would like to use this component if possible for the new docs site which will be launched by summit.

coreyvickery:
@marionnegp Miro link for exploring designs.

https://miro.com/app/board/uXjVN1WFZHg=/

zeroedin:
`<uxdot-sidenav>` has been created in #1465 to serve ux.redhat.com until more discovery on `<rh-sidenav>` is done

coreyvickery:
@markcaron @hellogreg I would also like to get this done in `Charmander`. The design is  still in progress, but a lot of work has been done so far. It would be nice to get it in front of design leads very soon, but I don't know what that new process is now.

coreyvickery:
Update 5/27: Team and stakeholders reviewing specs.

https://www.figma.com/proto/aOOto0i3Ph9AfBy42VVej8/rh-side-nav?page-id=1%3A16&node-id=1-2028&viewport=119%2C146%2C1&t=ow7NYJOKiqR9QBTA-1&scaling=min-zoom

coreyvickery:
@markcaron Do we have a developer assigned to work on this yet?

bennypowers:
We spoke about this at office hours today:

- "side nav" and "table of contents" should be considered patterns, not elements
- the table of contents pattern is composed of a tree view with an optional filter input
- the side nav pattern is composed of at least two new elements
    1. working title: drawer
    2. working title: tree view (aka drill down?) (nested list of page links)
    3. working title: learning path (linear list of nav i.e. extra-page links) with descriptions
- the side nav pattern may also have things like header widgets (filter search, logo, etc)
- the side nav pattern can compose in a learning-path element instead of the tree view

markcaron:
[...]

PR #2443, opened 2025-06-13
feat(navigation-drawer): add `<rh-navigation-drawer>`

### LINKED ISSUE #2465: [feat] `<rh-drawer>`
### Description

TBD

### Suggested solution

TBD

### Screenshots

_No response_

### Example API

```HTML

```

### Additional context

_No response_

zeroedin:
## What I did

1. Added `<rh-navigation-drawer>`

Closes #2465
Related issue: https://github.com/RedHat-UX/red-hat-design-system/issues/1329

## Testing Instructions

1. View [Deploy Preview](https://deploy-preview-2443--red-hat-design-system.netlify.app/elements/navigation-drawer/demos/)

### DIFF
--- a/.changeset/mighty-banks-move.md
+++ b/.changeset/mighty-banks-move.md
@@ -2,7 +2,7 @@
 "@rhds/elements": minor
 ---
 
-- ✨ Added `<rh-navigation-drawer>`
+✨ Added `<rh-navigation-drawer>`
 
 Navigation drawer is a vertical panel that provides a slottable area for navigation content. It is one part of the composable vertical navigation pattern, allowing users to access navigation items in a space-efficient way, collapsing to a disclosure pattern on narrow screens.
 


bennypowers:
maybe this should be called `<rh-drawer>`?

zeroedin:
> maybe this should be called `<rh-drawer>`?

I referenced this question in the issue before I saw your comment here.  My thought is that `<rh-drawer>` is different. This is explicitly designed as a navigation drawer; a standard drawer pattern has a toggle/trigger external to itself. The nav drawer collapses  at mobile viewport to a disclosure whereas the standard toggled drawer doesn't have a separate mobile experience.  This combination of things sets it apart from patterns found in other design system such as Patternfly.

.changeset/mighty-banks-move.md
```diff
@@ -0,0 +1,22 @@
+---
+"@rhds/elements": minor
+---
+
+- ✨ Added `<rh-navigation-drawer>`
```
adamjohnson:
```suggestion
✨ Added `<rh-navigation-drawer>`
```

elements/rh-navigation-drawer/rh-navigation-drawer.css
adamjohnson, line 1: Gonna need a few styles in this file for the blue focus ring. Right now it is just the browser default focus ring.

elements/rh-navigation-drawer/rh-navigation-drawer.ts
```diff
@@ -0,0 +1,195 @@
...
+
+import styles from './rh-navigation-drawer.css';
+
+
+/**
+ * Navigation Drawer is a component that displays a navigation menu in a vertical drawer in wide viewports and a disclosure in narrow viewports.
+ * @summary Navigation drawer
+ * @slot - Drawer content such as vertical navigation elements
+ * @slot header - (optional) header content such as titles and filters
+ * @slot footer - (optional) Footer content such as links and copyright info
```
adamjohnson, line 31: Document `--rh-z-index-navigation-drawer` here as well.

elements/rh-navigation-drawer/demo/expand-at-attribute.html
```diff
@@ -0,0 +1,32 @@
+<div class="container">
+  <div class="flex-wrap">
+    <rh-navigation-drawer expand-at="sm" collapsible>
+      Drawer content
+    </rh-navigation-drawer>
+    <h2>Document content</h2>
```
adamjohnson, line 6: Is it the expected behavior to have this "Document content" _not_ transition left when collapsing the drawer? 

AKA the "Document content" stays in the exact same location no matter if the drawer is expanded or collapsed?

adamjohnson (commented):
Few nitpicks. Overall, great work as always!

elements/rh-navigation-drawer/demo/expand-at-attribute.html
zeroedin: No it should all transition, might have missed that, will take a closer look.

zeroedin: That is up to the document and how the layout is setup. Maybe we need a demo showing a static grid usage as well.

zeroedin: On second thought, I had already done that. The container attribute allows you to do this; otherwise, the component will grab its next closest parent for container query inline-size info.

https://deploy-preview-2443--red-hat-design-system.netlify.app/elements/navigation-drawer/demos/#demo-container-attribute

zeroedin: And on third thought, what you are saying is actually the opposite of what I was thinking was happening.  The issue here is reflow of the flexbox. In order to trigger a reflow of the parent, a width calculation would need to change.  If we change a width we end up squishing the content of the drawer on the animation.  I'll have to think deeper on this and see what can be done.

zeroedin: Pushed some code that might resolve this issue.

adamjohnson (commented):
Still needs docs. Otherwise, LGTM!

coreyvickery:
@adamjohnson @zeroedin Can this be closed without docs? They'll be done next week.

bennypowers (changes requested):
We can't have multiple different drawer elements. see my comments on the issue.

bennypowers:
Notes from chat:
eventually drawer will have to support many different use cases, e.g

- has to be able to pull out of any screen edge (think AI chat)
- has to be able to slot in toggle button if desired (like ux-dot's close button)
- support push vs overlay
- etc

Open questions and review note
- the "it turns into a disclosure" part of this element, can we make that a pattern thing, or maybe move that into the content elements
- the container attribute is a bit of a code smell, points to larger design issues
- this should probably be part of a larger layout system

Counterpoint: material 3 has a navigation drawer

all in all, I think it's better that we put this feature back in the oven for more design, and shoot for a duoduo release

zeroedin:
Expanding on what Benny already mentioned.

The container attribute was intended to target an ancestor not explicitly owned by the component. This is necessary for the resize observer to set the "compact" state switching to a disclosure for when the targeted container is not the direct parent of the component, but a higher ancestor in a complex layout. 

It is a code smell and not ideal. My thought at the time of writing was trying to navigate the possibility of a nested container-type causing problems in the ancestor tree.  

Furthermore, the demo isn't complete enough to fully demonstrate its real need. For the attribute to be truly effective, the `.left-column` would also have to be of the container-type: inline-size. However, if that were the case, I believe the internal CSS would break anyway due to the `expand-at` attribute requiring CSS that would then have to explicitly call a container query name, which we can not set dynamically.

We may have to take a more layout-based approach as @bennypowers  mentioned vs the self-contained approach taken so far:

```
<div class="container">
  <rh-navigation-drawer>...</rh-navigation-drawer>
  Your  page content
</div>
```

vs layout based where the drawer has a page content slot itself and then becomes the container:

```
<rh-drawer>
  <div slot="panel">...</div>
  Your page content
</rh-drawer>
```

Issue #1756, opened 2024-08-13
[feat] Create `<rh-progress-stepper>` component


marionnegp:
### Description

We have an [existing design](https://ux.redhat.com/elements/progress-steps/) for progress steps in RHDS, but we may want to update it before it's built in the repo. 

Things to consider:
- What needs to be updated in this design? (Example: adding icons into the little circles)
- How does this work against backgrounds that aren't white or gray/black? Do we need another variant? (The [Summit 2024 interactive quiz](https://www.figma.com/proto/ud1sYpEP3rRriKit7o0DOv/2024-Custom-Summit-Web-Work?node-id=165-854&t=79Zj4CT9UMJzSH3q-1) is an example of a design that needed progress steps but ran into issues with the colors.)
- Should the name be changed from progress steps to progress stepper?

Examples:
- PatternFly Elements: [`<pf-progress-stepper>`](https://patternflyelements.org/components/progress-stepper/)
- PatternFly v5: [progress stepper](https://www.patternfly.org/components/progress-stepper)
- PatternFly v6: [progress stepper](https://staging-v6.patternfly.org/components/progress-stepper)

### Acceptance Criteria

- [ ] Design done
- [ ] Development done

marionnegp:
@markcaron @hellogreg, should this be in the Cubone or Diglett milestone?

markcaron:
@marionnegp Cubone works. We should have enough lead time to make that deadline.

KamiQuasi:
Experimentation on accessible structure and setup is happening here: https://codepen.io/kamiquasi/pen/zxYgGqE?editors=1100

Noting that feature with the "popover" helper will likely happen after we implement popover.

PR #2449, opened 2025-06-16
feat(progress-stepper): add <rh-progress-stepper> element

KamiQuasi:
## What I did

1. Added: `<rh-progress-steps>` and `<rh-progress-step>`

Closes: #1756 

## TODO

- [x] Fix inactive line color
- [x] Fix vertical label issue
- [x] Dark mode demo
- [x] Tests
- [x] Docs / JSDoc
- [x] Changeset 

## Testing Instructions

1. Load the [Demo](https://deploy-preview-2449--red-hat-design-system.netlify.app/elements/progress-stepper/demos/)
2. Check Horizontal Steps
3. Check Vertical Steps

## Notes to Reviewers

[Figma Specs](https://www.figma.com/design/GrBO7J0H25I0rMknBSbRJC/rh-progress-stepper?node-id=7523-1567&p=f&m=dev)


elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,143 @@
...
+
+    /* Body text/md - Medium weight */
+    font-weight: var(--rh-font-weight-body-text-medium, 500);
+    text-decoration:
+      underline
+      dotted
+      var(--rh-color-text-secondary, light-dark(var(--rh-color-text-secondary-on-light, #4d4d4d), var(--rh-color-text-secondary-on-dark, #c7c7c7)))
+      1px;
+    text-decoration-skip-ink: none; /* 9.375% */
+    text-underline-offset: 30%; /* 4.8px */
```
markcaron: I know the Figma values may have `30%`, but I believe that's to deal with Figma's own discrepancies with CSS. Visually, `30%` here looks off.

image

So we should go with `text-underline-offset: 5px;` instead. This is the exact same offset value as our [text underlines for links](https://ux.redhat.com/foundations/interactions/links/#default). And, we might need to remove `text-underline-position: from-font;`

image

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,143 @@
...
+}
+
+:host {
+  grid-row: var(--stepGridRow);
+  width: 100%;
+  height: 100%;
+  padding-bottom: var(--stepCompactVPadding, var(--rh-length-2xl, 32px));
+  padding-right: var(--stepCompactHPadding, var(--rh-length-5xl, 80px));
+  display: grid;
+  background:
```
markcaron: This is a neat idea for how to handle the lines connecting each step, but it looks a fuzzy/blurry since it's a gradient.

image

Narrowing the gradient values to `13px / 14px / 15px` is too thin, though. We might need a different approach, unless there's a way to make this gradient less blurry.

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,143 @@
...
+  grid-template-rows: var(--stepTemplateRows, 24px auto);
+  justify-self: var(--stepJustifySelf, start);
+
+  /* Body text/md */
+  font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif);
+  font-size: var(--rh-font-size-body-text-md, 1rem);
+  font-style: normal;
+  font-weight: var(--rh-font-weight-body-text-regular, 400);
+  line-height: 1.5; /* 150% */
+  gap: var(--stepRowGap, 16px) var(--stepColGap, 16px);
```
markcaron: The `gap` here is apply to the space between the `<strong>` and the `<p>` (description) as well, which is too much. Might need to group these 2 items so they don't get the same game and are "attached" to one another logically.

image

Mockup:
image

markcaron (dismissed):
I may have some more thoughts later, but wanted to get these in as soon as possible.

elements/rh-progress-stepper/rh-progress-step.css
KamiQuasi: Fixed the gap so it is the correct 4px variable. To compensate for the vertical distance between the graphical steps and text on the horizontal stepper there is an additional 4px of padding on top of the text.

elements/rh-progress-stepper/rh-progress-step.css
KamiQuasi: Fixed the gradient so it is hard line transitions between colors. Just set the end and starts to be the same value.

elements/rh-progress-stepper/rh-progress-step.css
KamiQuasi: Switched to the 5px, and had to take the underline position off as thought, because if the text scrolled it lost the underline on the second line.

markcaron (commented):
List Graduated; Track Marched.

Visuals look good, IMO. Will need at least Adam's sign off on accessibility.

elements/rh-progress-stepper/demo/compact-vertical.html
```diff
@@ -24,6 +17,13 @@
     description="This step has failed and needs to be retried">
   </rh-progress-step>
 
+  <rh-progress-step
+    state="active"
+    label="Active Step"
```
bennypowers: if this renders visible test, that's fine. if it's only for SR, we've used the conventional attribute `accessible-label`

elements/rh-progress-stepper/demo/vertical-stepper.html
```diff
@@ -0,0 +1,41 @@
+<rh-progress-stepper orientation="vertical">
+  <rh-progress-step
+    state="active"
+    label="Active Step"
+    description="Currently working on this step"
+    aria-current="step">
+  </rh-progress-step>
+  
+  <rh-progress-step state="complete">
+    <a href="https://www.redhat.com" slot="label">Complete Step</a>
```
bennypowers: how about `<rh-progress-step href="...">`?

elements/rh-progress-stepper/demo/rh-progress-stepper.html
```diff
@@ -0,0 +1,44 @@
...
+    state="fail"
+    label="Failed Step"
+    description="This step has failed and needs to be retried">
+  </rh-progress-step>
+
+  <rh-progress-step
+    state="active"
+    label="Active Step"
+    description="Currently working on this step"
+    aria-current="step">
```
bennypowers: can we handle aria internally?

package.json
```diff
@@ -33,6 +33,7 @@
     "Em Nouveau <enouveau@redhat.com>",
     "Ivana Rodriguez <ivana@redhat.com>",
     "Kelsey Swanson <kswanson@redhat.com>",
+    "Luke Dary <ldary@redhat.com>",
```
bennypowers, line 36: *nice*

netlify.toml
bennypowers, line 1: good catch!

elements/rh-progress-stepper/test/rh-progress-stepper.spec.ts
bennypowers, line 1: some more tests i think would be appreciated. particularly ax tree assertions

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @summary       Progress Stepper
+ * @slot          - Use this slot for `<rh-progress-step>` items
+ *
+ */
+@customElement('rh-progress-stepper')
+export class RhProgressStepper extends LitElement {
+  static readonly styles: CSSStyleSheet[] = [styles];
+
+  @property({ reflect: true }) orientation: ProgressStepsOrientation = 'horizontal';
```
bennypowers, line 71: props need jsdoc: think cem as llm fodder

elements/rh-progress-stepper/rh-progress-stepper.css
```diff
@@ -0,0 +1,144 @@
...
+        var(--rh-color-text-secondary-on-dark, #c7c7c7)))
+      1px;
+    text-decoration-skip-ink: none; /* 9.375% */
+    text-underline-offset: 5px; /* 4.8px */
+  }
+
+  /* & ::slotted(rh-progress-step) {
+    --stepLineColor: var(--rh-color-text-primary);
+  } */
+
```
bennypowers:
```suggestion
```

elements/rh-progress-stepper/rh-progress-stepper.css
```diff
@@ -0,0 +1,144 @@
...
+  & ::slotted(rh-progress-step:last-of-type) {
+    --stepLineStyle: none;
+  }
+}
+
+:host([orientation='vertical']) {
+  & #step-list {
+    grid-auto-flow: row;
+
+    --stepLineAngle: 90deg;
```
bennypowers:
```suggestion
    --_step-line-angle: 90deg;
```
etc

elements/rh-progress-stepper/rh-progress-stepper.css
```diff
@@ -0,0 +1,144 @@
...
+    padding: 0;
+    position: absolute;
+    white-space: nowrap;
+
+    /* End Visual Hiding */
+    font-weight: var(--rh-font-weight-body-text-medium, 500);
+    text-decoration:
+      underline
+      dotted
+      var(--rh-color-text-secondary,
```
bennypowers: remind me @zeroedin what was our conclusion, that these need fallbacks in element css? we can rely on base-theme being there, so I think we can remove these

elements/rh-progress-stepper/rh-progress-stepper.css
```diff
@@ -0,0 +1,144 @@
...
+        light-dark(var(--rh-color-text-secondary-on-light, #4d4d4d),
+        var(--rh-color-text-secondary-on-dark, #c7c7c7)))
+      1px;
+    text-decoration-skip-ink: none; /* 9.375% */
+    text-underline-offset: 5px; /* 4.8px */
+  }
+
+  /* & ::slotted(rh-progress-step) {
+    --stepLineColor: var(--rh-color-text-primary);
+  } */
```
bennypowers:
```suggestion
```

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -0,0 +1,88 @@
...
+    super.connectedCallback();
+    this.role = 'listitem';
+  }
+
+  render() {
+    return html`
+      <slot name="icon">
+      <rh-icon
+        icon="${this.#icon}"
+        set="${this.state === 'custom' ? this.customIconSet : 'ui'}"
```
bennypowers:
```suggestion
          icon="${this.#icon}"
          set="${this.state === 'custom' ? this.customIconSet : 'ui'}"
```
nit

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -0,0 +1,88 @@
...
+@customElement('rh-progress-step')
+@themable
+export class RhProgressStep extends LitElement {
+  static styles = styles;
+
+  @property({ reflect: true }) state: ProgressStepState = 'inactive';
+  @property({ type: String }) label = '';
+  @property({ type: String }) description = '';
+  @property({ type: String }) customIcon = '';
+  @property({ type: String }) customIconSet = 'ui';
```
bennypowers: props need well-written jsdoc. think cem as llm fodder

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -0,0 +1,88 @@
...
+ * 
+ * @slot - The content of the progress step.
+ * @slot icon - The icon to display for the progress step.
+ * @slot label - The label to display for the progress step.
+ * @slot description - The description to display for the progress step.
+ */
+@customElement('rh-progress-step')
+@themable
+export class RhProgressStep extends LitElement {
+  static styles = styles;
```
bennypowers:
```suggestion
  static styles = [styles];
```

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,192 @@
...
+    --stepLineColor: var(--rh-color-border-subtle);
+  }
+
+  --stepLineColor: var(--rh-color-text-primary);
+
+  & rh-icon { border-color: var(--rh-color-text-primary); }
+
+  & strong {
+    /* color: var(--rh-color-text-secondary, #151515); */
+
```
bennypowers:
```suggestion
```

bennypowers (changes requested):
Needs a minor changeset

elements/rh-progress-stepper/demo/compact-vertical.html
KamiQuasi: The label attribute does render visible text, but can also be overridden by a `label` slotted element.

elements/rh-progress-stepper/demo/rh-progress-stepper.html
KamiQuasi: Yes and no. We could do that, if we could know which one is actually current. Based on discussion (see May 7th Slack thread) about the behavior being somewhat nondeterministic I don't know that we can accurately say what is current or not. My original thought had been to do something with a `next` function to actually progress the steps, but the lack of explicit success/fail state makes that more inflexible than what I think we're trying to do. Happy to take some suggestions on this though.

elements/rh-progress-stepper/demo/vertical-stepper.html
KamiQuasi: Love it! Will get that in.

elements/rh-progress-stepper/demo/rh-progress-stepper.html
bennypowers: @adamjohnson thoughts?

elements/rh-progress-stepper/rh-progress-step.ts
adamjohnson: Is there anything preventing us from having default values for `label` and `description`? Doing so may help accessibility. I realize those values would need to change based on the current `state` attribute.

elements/rh-progress-stepper/demo/rh-progress-stepper.html
adamjohnson: Handling `aria-current` internally would probably require adding a `<div id="container">` to the shadowdom to attach `aria-current` to.

It might make the API less verbose too, should we decide to implement [default values](https://github.com/RedHat-UX/red-hat-design-system/pull/2449#discussion_r2219633820) for the various properties. 

That said, I don't know if people are writing custom messages for each progress stepper in userland.

Ideally, `aria-current="step"` is only applied to one step—the currently active step. Implementors will ultimately have to decide what that means in their environment.

elements/rh-progress-stepper/rh-progress-step.ts
KamiQuasi: @adamjohnson we can definitely add defaults per state, but I definitely don't know what the values would be based on actual (or projected) usage.

adamjohnson: Screenshot of the progress stepper showing completed, warning, and failure

Maybe just add defaults for `label`. Eg: Completed, Warning, Failure.

adamjohnson:
Screenshot showing progress stepper in edge and safari. In safari, the labels are showing when they should not show.

Heads up: in Safari the compact demos have visible labels.

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,186 @@
...
+  margin: 0;
+  overflow: auto;
+  padding: 0;
+  position: relative;
+  white-space: normal;
+
+  /* End Visually Showing */
+}
+
+:host-context([compact]) {
```
adamjohnson: @KamiQuasi this selector is why the compact version is not compact in Safari. `:host-context()` is [not yet supported](https://caniuse.com/mdn-css_selectors_host-context) in Safari and Firefox.

You'll probably have to use Lit Context for this. I recently learned Lit Context and wrote up some [instructions](https://gist.github.com/adamjohnson/b42615a53d36a236c3512a8a4f6f8e2b) for implementing it for my future self. Maybe you'll find them useful.

Also, TIL `:host-context()`! Can't wait till that replaces JS / library solutions.

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,186 @@
...
+          light-dark(var(--rh-color-text-secondary-on-light, #4d4d4d),
+          var(--rh-color-text-secondary-on-dark, #c7c7c7)))
+        1px;
+      text-decoration-skip-ink: none; /* 9.375% */
+      text-underline-offset: 5px; /* 4.8px */
+    }
+  }
+}
+
+:host-context([orientation='horizontal']) {
```
adamjohnson: You'll have to replace `:host-context()` here too.

IIRC, Benny mentioned something in Slack about SSR issues with multiple contexts. I can't seem to find the convo at the moment but consider checking in with him about it if you use Lit Context.

elements/rh-progress-stepper/rh-progress-step.css
adamjohnson: Whelp, sounds like Safari and Firefox don't plan to implement `:host-context()` (since 2017). :sadpanda: 

[More backstory](https://www.sarahgebauer.com/post/host-context-and-browser-ecosystem-disappointment/).

elements/rh-progress-stepper/rh-progress-step.ts
markcaron: I can see defaults for `label`, but `description` is an optional feature.

markcaron:
Image on Overview page needs to be updated as well.

coreyvickery:
@KamiQuasi Does this element support Popover? My design mimics [this version](https://www.patternfly.org/components/progress-stepper#with-help-popover) from PatternFly. I can remove the underline styling from Figma and the docs if it doesn't support Popover right now.

coreyvickery:
> it doesn't have that interface at the moment, but we can wire that up once popover is available

@KamiQuasi Great, let's get rid of the underline styling for now then.

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  /**
+   * Sets the description text for the progress step
+   */
+  @property({ reflect: true }) description = '';
+
+  /**
+   * Sets a custom icon name when the state is 'custom'
+   */
+  @property({ reflect: true, attribute: 'custom-icon' }) customIcon = '';
```
bennypowers: maybe this (and set) should just be `icon` and `icon-set`, and it's understood that if this is set, use it (because it's customized), otherwise, use the icon for the state.

markcaron: @bennypowers, @zeroedin and I did some work on this already. Can you double check my recent commit to see if it meets your expectations here?

bennypowers (approved):
linear gradient towards mastery

adamjohnson:
@coreyvickery Docs have been coded up. PTAL, TY.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -0,0 +1,196 @@
...
+    });
+  }
+
+  render(): TemplateResult<1> {
+    const compact = this.compact ?? false;
+    const vertical = this.orientation === 'vertical' || this.mobile;
+
+    return html`
+    <div id="container" class="${classMap({ compact, vertical })}">
+      <strong id="current-step">${this.current}</strong>
```
adamjohnson:
```suggestion
      <strong ?hidden="${!compact}" id="current-step">${this.current}</strong>
```

This will make sure the "Active step" is hidden from AT when the labels are displayed.

This element will not be hidden when users implement the `compact` stepper.

marionnegp:
Mind tagging me in for a design review when this is ready?

elements/rh-progress-stepper/rh-progress-stepper.ts
markcaron: @adamjohnson can you double check that this is resolved?

elements/rh-progress-stepper/rh-progress-step.ts
markcaron: @bennypowers can you double check that this is resolved? I believe it is.

markcaron:
Ok, I think this element is ready to go.

- [ ] @marionnegp Should be ready for a final design review of the element.
- [ ] @coreyvickery Can you do a design review on the Docs?
- [ ] @adamjohnson and @bennypowers can you each double check / re-review and mark your requests as resolved?

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: Resolved in [ffaad23](https://github.com/RedHat-UX/red-hat-design-system/pull/2449/commits/ffaad234b6559d860e6a4f383d7bd106f3ba240a).

elements/rh-progress-stepper/README.md
```diff
@@ -0,0 +1,114 @@
...
+    custom-icon="user-check" 
+    custom-icon-set="ui">
+    Custom Step
+  </rh-progress-step>
+</rh-progress-stepper>
+```
+
+## Accessibility
+
+- Use `aria-current="step"` on the currently active step
```
adamjohnson:
```suggestion
- Use `state="active"` on the currently active step to denote the current step
```

elements/rh-progress-stepper/rh-progress-stepper.css
```diff
@@ -0,0 +1,155 @@
...
+  --_current-step-color: var(--rh-color-status-note);
+}
+
+:host([state='complete']) {
+  --_current-step-color: var(--rh-color-status-success);
+}
+
+:host([state='warn']) {
+  --_current-step-color: light-dark(var(--rh-color-yellow-50), var(--rh-color-yellow-30));
+}
```
adamjohnson:
```suggestion
:host([state='warn']) {
  --_current-step-color:
    light-dark(
        var(--rh-color-yellow-50, #b98412),
        var(--rh-color-yellow-30, #ffcc17)
      );
}
```

Good to include fallbacks

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,148 @@
...
+}
+
+/* Complete state styling */
+:host([state='complete']) {
+  --_step-color: var(--rh-color-status-success);
+}
+
+/* Warning state styling */
+:host([state='warn']) {
+  --_step-color: light-dark(var(--rh-color-yellow-50), var(--rh-color-yellow-30));
```
adamjohnson:
```suggestion
  --_step-color:
    light-dark(
        var(--rh-color-yellow-50, #b98412),
        var(--rh-color-yellow-30, #ffcc17)
      );
```

Good to have fallbacks.

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,148 @@
...
+  grid-template-columns: var(--_step-template-columns);
+  grid-template-rows: var(--_step-template-rows, 24px) auto;
+  grid-template-areas: var(--_step-template-areas);
+  gap: var(--_step-row-gap, 16px) var(--_step-column-gap, 16px);
+  justify-self: var(--_step-justify-self, start);
+  align-items: var(--_step-align-items);
+  padding-block-end: var(--_step-compact-vertical-padding);
+  padding-inline-end: var(--_step-compact-horizontal-padding);
+  width: 100%;
+  height: 100%;
```
adamjohnson:
```suggestion
  inline-size: 100%;
  block-size: 100%;
```

Logical properties FTW.

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -0,0 +1,148 @@
...
+    padding-inline-end: var(--_step-compact-horizontal-padding, var(--rh-length-5xl, 80px));
+  }
+}
+
+/* Icon styling */
+:is(rh-icon, .no-icon) {
+  grid-area: icon;
+  place-self: center var(--_step-icon-justify-self);
+  height: var(--rh-size-icon-02, 24px);
+  width: var(--rh-size-icon-02, 24px);
```
adamjohnson:
```suggestion
  block-size: var(--rh-size-icon-02, 24px);
  inline-size: var(--rh-size-icon-02, 24px);
```

adamjohnson (commented):
Couple easy edits

adamjohnson (approved):
Labeled Gradual Task Markers

marionnegp (changes requested):
Sorry I didn't get to this earlier today!

- [x] The line that runs through the stepper looks like it might be using `--rh-color-border-subtle` with every state. It should turn to `--rh-color-border-strong` in between completed steps.
- [x] The icon size is correct, but the border around the icon is positioned outside in the mocks. This should end up adding a little more whitespace between the border and the icon.
- [x] The label for the warning state should be `--rh-color-yellow-60`. Not enough contrast otherwise.
- [x] In the Figma file, it looks like there should be a surface color (`--rh-color-surface-lightest` or `--rh-color-surface-darkest`) for each circle. Would be helpful to keep if someone tries to use this against a non-surface color.
- [x] In the compact horizontal variant, the mocks show 16px in between steps. In the demo, it looks like there's some extra space after the icon that's adding to the 16px of padding.
Screenshot 2025-08-04 at 9 57 32 PM

- [x] I'm not 100% sure how spacing in the default, horizontal variant works, but it looks bigger than the 80px (`--rh-space-5xl`) in the mocks. Is it possible to match the mockup's spacing? Here's what I see when I inspect one step. I added an 80px spacer to the screenshot.
Screenshot 2025-08-04 at 10 07 38 PM

marionnegp (approved):

PR #2455, opened 2025-06-17
chore: adopt `cem`

bennypowers:
## What I did

1. [x] replace `@custom-elements-manifest/analyzer` with `@pwrs/cem`
2. [ ] ~move part and slot docs into templates~ let's do this separately
3. [ ] ~move css prop docs into css files~ let's do this separately


adamjohnson (changes requested):
Getting `return is not defined` after running `npm run dev` and loading localhost:8000. 8000 doesn't load / is a white page in the browser. 

Here's my terminal output: https://app.warp.dev/block/DD60n1cKaVvHaLzC8MOwnR

bennypowers:
I think the test failures result from the upgraded playwright version

elements/rh-accordion/rh-accordion-header.ts
```diff
@@ -92,8 +83,19 @@ export class RhAccordionHeader extends LitElement {
...
         <span id="header-container" class="${classMap({ [accents ?? '']: !!accents })}">
-          <span id="header-text" part="text"><slot></slot></span>
-          <span part="accents"><slot name="accents"></slot></span>
+          <!-- summary: inline element containing the heading text or slotted heading content -->
+          <span id="header-text" part="text">
+            <!-- summary: accordion toggle content -->
+            <slot></slot>
+          </span>
+          <!-- summary: container for accents within the header -->
+          <span part="accents">
```
adamjohnson, line 92: Looking at the DP, these comments are not showing up on the [Code page](https://deploy-preview-2455--red-hat-design-system.netlify.app/elements/accordion/code/#rh-accordion-header-apis) in the docs for `rh-accordion-header`.

I know you said we were going to handle this in #2482, but, based on this diff, should this functionality be working here?

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -31,6 +31,9 @@ export type NavigationPrimaryPalette = Extract<ColorPalette, (
  * Primary navigation helps users orient themselves and move through websites and domains.
  *
  * @summary       Primary navigation
+ *
+ * @alias Navigation (primary)
```
adamjohnson, line 35: Should this be `navigation-primary`?

elements/rh-navigation-secondary/rh-navigation-secondary.ts
```diff
@@ -48,7 +48,11 @@ function focusableChildElements(parent: HTMLElement): NodeListOf<HTMLElement> {
 
 /**
  * The Secondary navigation is used to connect a series of pages together. It displays wayfinding content and links relevant to the page it is placed on. It should be used in conjunction with the [primary navigation](../navigation-primary).
+ *
  * @summary Propagates related content across a series of pages
+ *
+ * @alias Navigation (secondary)
```
adamjohnson, line 54: Should this be `navigation-secondary`?

adamjohnson (changes requested):
Some more things I found:

  * On each components [Overview page](https://deploy-preview-2455--red-hat-design-system.netlify.app/elements/accordion/) in the docs, the knobs demo takes the first demo listed under the [Demos page](https://deploy-preview-2455--red-hat-design-system.netlify.app/elements/accordion/demos/). This seems to never be the default `rh-{component}.html` demo—rather, some variant. The Overview knobs demo should show each component's default demo.
  * On the [Demos page](https://deploy-preview-2455--red-hat-design-system.netlify.app/elements/accordion/demos/) (eg: `/elements/accordion/demos/`), the default demo is never first. It should appear first in the list.
  * If I clone this PR and run it locally on the dev server, clicking the image for each demo does not include the trailing `/` for the base demo's permalink. 👇
Red boxes outlining the photos on localhost 8000

  * Running this locally and hitting the correct base demo URL (eg: `localhost:8000/elements/accordion/demo/`) does not show the accordion demo, but the All elements overview (same view as `localhost:8000`).
      * Hitting the other variants of a component _does_ work
  * When hitting localhost:8000, there is a link ("Accordion") in the first `<pf-card>` and, instead of having the header say "Accordion", it says "Expanded Index". 
Red arrow pointing to the first link of accordion demos on localhost 8000

  * When hitting `localhost:8000`, Jump Links and Tabs have duplicate links to their various demos.
  * When hitting `localhost:8000`, not all Footer demos are listed under the Footer component.

elements/rh-navigation-primary/rh-navigation-primary.ts
bennypowers: it gets slugified in cem. this is the pretty name for the menu

elements/rh-accordion/rh-accordion-header.ts
bennypowers: 8511f229

elements/rh-navigation-secondary/rh-navigation-secondary.ts
bennypowers: see above

PR #2463, opened 2025-06-27
feat(navigation-primary): bimodal logo

### LINKED ISSUE #2312: [feat] `<rh-navigation-primary>` should use hat only logo at smallest breakpoints
### Description

The 320px breakpoint can use the hat only Red Hat logo. Can this be updated?

### Suggested solution

[Figma file](https://www.figma.com/design/Ob5kpQip4qdSFJ7gTXSP2r/Unified-Navigation?node-id=1025-19329)

Image
Image

### Screenshots

_No response_

### Example API

```HTML

```

### Additional context

_No response_

zeroedin:
## What I did

1. The default logo is now bi-modal, automatically adjusting the text color to a light or dark scheme.
2. ~~Added `logo-href` which sets the `href` on the `<a>` wrapping the logo, defaults to `/`.~~ removed for [future implementation](https://github.com/RedHat-UX/red-hat-design-system/issues/2472)
3. ~~Added `logo-text-hide-at` sets the breakpoint at which the logo text is hidden~~ removed a default value of 1440px will apply everywhere, see note to reviewers.
    - ~~breakpoints available 'sm' | 'md' | 'lg' | 'xl' | '2xl' defaulting to 'md' (992px)~~
4. If you are slotting in a logo, these options will not take effect.

Closes #2312 

## Testing Instructions

1. View deploy preview

## Notes to Reviewers

 - @marionnegp Should we default to `xs` for `logo-text-hide-at`? That is what Figma seems to portray but is contrary to some of the images in the docs PR  https://deploy-preview-2395--red-hat-design-system.netlify.app/elements/navigation-primary/style/.
 
 
image

~~Update: as discussed with @marionnegp we are setting the default to what is currently being used on RHDC to 1440px. This larger setting was made to accommodate translations before this PR added the configuration option. In the future, translation pages can utilize independent settings, and we will retain this default setting.~~

Second update: after further consideration it was decided that 1440px instead will be the default everywhere.


elements/rh-navigation-primary/rh-navigation-primary.css
```diff
@@ -375,4 +380,46 @@
       }
     }
   }
+
+  &.logo-text-hide-at-xs #logo svg {
```
bennypowers: While ordinarily we should prefer shadow classes, because of SSR, we should find an alternative in the case

Ideally, this would all be done in CSS

zeroedin: @bennypowers I'm not sure I follow can you reword what you are asking?

elements/rh-accordion/rh-accordion-header.css
```diff
@@ -5,6 +5,7 @@
   --_padding-inline-start: var(--rh-space-xl, 24px);
   --_after-background-color: transparent;
   --_expanded-background-color: var(--rh-color-accent-brand);
+  --rh-this-is-not-a-known-token: 100px;
```
bennypowers: Sure isn't

marionnegp (approved):
Looks good!

elements/rh-navigation-primary/rh-navigation-primary.css
bennypowers: This new attribute,

If it changes the shadow root

RHDC engineers will have to calculate that change in their template

Until they're able to run lit SSR on each page

When we spoke with them about nav1 being simple enough to fake lit SSR by running it on your laptop with the only two, string-forwarding attributes given twig {{expression}} values,..

bennypowers: This sorta throws a 🐒 🔧 in that plan

So let's try to do it with CSS only, since anyways it's a presentational thing

zeroedin: After more discussion, we have removed the configurable attribute. Now there is only the 1440px breakpoint for hiding the logo text. Below 1440px, the nav will always hide the text of the default logo.

zeroedin:
> do we really really really need logo href? also this needs a docs update

@bennypowers In order to take advantage of not having to slot the image yourself, handle bimodal svg etc yes. @zhawkins is currently overriding the whole slot to do this, but then misses out on the default handling of color-scheme. IMHO its a small add for a big gain.  The links for him need to point at i18n based URLs, which will be common.

bennypowers (changes requested):
just need to update demos and changeset and we're g2g

PR #2539, opened 2025-07-28
docs: correct nesting, incorrect placement of styles

zeroedin:
## What I did

1.  Corrects broken CSS styles, nesting and incorrect placement of typography rules.
2. Fixed weird transition animation on links.  Should have been placed only on underline rule.
3. ~~Proposal of view transitions to smooth page loads. (we can remove from this PR and propose in another if desired)~~ moving to its own PR.

## Testing Instructions

1.

### DIFF
--- a/docs/styles/styles.css
+++ b/docs/styles/styles.css
@@ -79,20 +79,6 @@
     white-space: nowrap;
     border: 0;
   }
-
-  @view-transition {
-    navigation: auto;
-  }
-
-  ::view-transition-old(main),
-  ::view-transition-new(main) {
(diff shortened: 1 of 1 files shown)


bennypowers (commented):
would it be ok if we saved the view transition stuff for a separate pr? I'd like to evaluate that on its own merits

zeroedin:
> would it be ok if we saved the view transition stuff for a separate pr? I'd like to evaluate that on its own merits

yup no worries, with the broken CSS I couldn't apply it yet, so just figured why not drop it here. Will wait till this merges to create a new PR with the view transitions.

Issue #2409, opened 2025-05-29
[epic][docs] Revise Element JSDocs for clarity


bennypowers:
### Description

Let's go over *each of our element files*, and in each one:
- review the JSDoc for the exported element class, ensuring it *fully describes* the element, it's uses, guidelines, and limitations
- review each of the element's public apis (slots, properties, attributes, methods, events, etc) and for each one
  - ensure it *fully describes* that particular API, it's uses, limitations, guidelines, etc.
  - Make use of the new HTML and CSS comments feature in CEM to document slots, parts, and css props *inline*

## Guidelines for writing prose
a thing can be an element, attribute, token, event, method, etc:

- what is this thing? does it have special meaning in the design system?
- when is it appropriate to use this thing?
- where in a page or app should this be found?
- how can this thing be composed with other things? are there things that this MUST be composed with? e.g. there can only be one nav2 on a page, nav2 dropdowns can only be used in nav2, universal footer is always either at the bottom of the page, or composed into a footer?
- why does this thing exist in the design system?
- who are the intended downstream devs and end users of this thing?

## Notes for AI assistants:
- See PR #2351 for an example of how to do this
- Read [CEM generate command](https://bennypowers.dev/cem/docs/commands/generate/) for information on the supported comment syntax, and [Writing effective description for AI](https://bennypowers.dev/cem/docs/mcp/writing-descriptions/) for instructions on writing effective descriptions.

zeroedin:
Various missing items from the review of JSDoc/cem PR #2482  that can be converted into sub issues here.

### rh-accordion
- `accents` [attr] missing description
- `hasChanged()` method missing description

### rh-accordion-header
- `default slot` missing description

### rh-accordion-panel
- `expanded` [attr] missing description
- `container` [part] missing description

### rh-announcement
- `close()` method missing description

### rh-audio-player:
-  `--rh-audio-player-text-color` missing description
-  `--rh-audio-player-range-progress-color` missing description
-  `--box-fill` missing description, should be deprecated and renamed to either a private `--_box-fill` or exposed as a public `--rh-audio-player-box-fill`
-  `--rh-icon-size` missing description 

### rh-audio-player-about
- `scrollText()` method missing description
- `--rh-avatar-size` missing description

### rh-audio-player-scrolling-text-overflow
- `firstUpdated()` method missing description, should be ignored by cem?
- `stopScrolling()` method missing description
- `startScrolling()` method missing description
[...]

PR #2548, opened 2025-08-07
docs(progress-stepper): update docs

marionnegp:
## What I did

1.

## Testing Instructions

1.


adamjohnson (approved):
Linked Guidelines Thoroughly Modified

Impeccable work as usual!

bennypowers (changes requested):
Please add content guidelines to inline docs, in property and slot positions

adamjohnson:
I think Benny is talking about https://github.com/RedHat-UX/red-hat-design-system/issues/2409 more or less. To help LLMs give more accurate answers, adding content guidelines _inline_ to the `rh-progress-stepper.ts`, ...et all, files. 

```diff
  /**
   * Sets the description text for the progress step
   * Overridden by the `description` slot.
+  * Use only use a few words for the description.
   */
  @property({ reflect: true }) description?: string;
```

This could be in addition to whats in `/docs`. Or perhaps a short description that would include a URL that points to (for example) [Writing Content](https://ux.redhat.com/elements/progress-stepper/guidelines/#writing-content) for further info.

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -110,7 +111,9 @@ export class RhProgressStep extends LitElement {
     const compact = this.compact ?? false;
     const ariaCurrent = this.currentStep === this ? 'step' : undefined;
     const labelSlot = html`
-      <!-- A short title for the step, which also serves as the step's accessible name -->
+      <!-- A short title for the step, which also serves as the step's accessible name.
+          A title can also be
```
adamjohnson: > A title can also be

Looks like this sentence needs finished.

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -6,35 +6,49 @@
 }
 
 :host {
-  --_step-line-style:
+  --_step-line-style: /* @summary Styles connecting lines */
```
bennypowers: css props beginning with `--_` should _not_ be documented

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -6,35 +6,49 @@
...
+  justify-self: var(--_step-justify-self, start); /* @summary Horizontal step alignment */
+  align-items: var(--_step-align-items); /* @summary Vertical content alignment */
+  padding-block-end: /* @summary Bottom padding for compact variant in vertical layout */
+    var(--_step-compact-vertical-padding);
+  padding-inline-end: /* @summary Horizontal padding for compact variant */
+    var(--_step-compact-horizontal-padding);
   inline-size: 100%;
   block-size: 100%;
-  color: var(--rh-color-text-secondary); /* sets description color */
+  color: /* @summary A step's description color */
```
bennypowers: this is an interesting case. we have here a system token - which should have it's own documentation here: https://github.com/RedHat-UX/red-hat-design-tokens/blob/338a97ddb1dab3cbb6b7e2db34ca2cd4e792959c/tokens/color/text.yaml#L51-L71

@adamjohnson @zeroedin and Marionne, WDYT should be the expected behaviour when cem finds a token that has inline documentation? append/prepend the inline docs? ignore them? replace the token docs? print inline first, followed by "This is a design system token, global documentation follows:"?

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -71,7 +93,8 @@
   grid-area: label;
   align-self: center;
   color: var(--_step-label-color);
-  font-weight: var(--_current-step-weight, var(--rh-font-weight-body-text-regular, 400));
+  font-weight: /* @summary Font weight for labels of steps */
+    var(--_current-step-weight, var(--rh-font-weight-body-text-regular, 400));
```
bennypowers: see above comment re system tokens. a note on syntax: when there are two tokens in one expression, as in this example, the comment will be assumed to belong to the outermost, in this case `--_current-step-weight`. Did you mean this instead? 
```suggestion
  font-weight: 
    var(--_current-step-weight,
    /* @summary Font weight for labels of steps */ var(--rh-font-weight-body-text-regular, 400));
```

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -72,7 +72,8 @@ export class RhProgressStep extends LitElement {
   @property({ reflect: true }) state?: ProgressStepState;
 
   /**
-   * Sets the description text for the progress step
+   * Sets the description text for the progress step if more context is needed.
+   * Descriptions are secondary to titles.
```
bennypowers, line 76: description guidelines:
>   - Use between 1 - 3 words for the title and 

Should be mentioned here.

Maybe this should be separated by markdown headings

~~~md
The description text for the progress step, which is 
a short piece of descriptive text, secondary to the step's title.

### Guidelines
Add a description if a step requires more context than just it's title and optional custom icon.
Descriptions should be around 40 characters (or a max of 2 lines at the 768px breakpoint).

### Code
This attribute is overridden by the `description` slot.
~~~

This would already be helpful for mcp, and down the line we could use this to build the guidelines/code pages from the manifest.

cc @coreyvickery @adamjohnson @zeroedin @markcaron :point_up: PTAL and vet the idea

elements/rh-progress-stepper/rh-progress-step.ts
bennypowers: like description above, the guidelines for titles (ie 1-3 words) should be stated here

elements/rh-progress-stepper/rh-progress-step.ts
```diff
@@ -139,10 +142,12 @@ export class RhProgressStep extends LitElement {
...
   /**
+   * Icons for each step indicates the status of a process or task.
+   * Icons change as users progress.
    * Computes the icon for the step:
-   * always use the prescribed warn or fail icons
-   * otherwise, use the custom user icon,
-   * or fall back to the default active/inactive icon
+   * Always use the prescribed warn or fail icons.
+   * Otherwise, use the custom user icon, if available,
+   * or fall back to the default active/inactive icon.
```
bennypowers: It appears that this belongs on the icon `@property`, rather than on this `protected` (which means private to this class and it's subclasses, i.e. not meant for public use) method. see also notes about guidelines / code docs

elements/rh-progress-stepper/rh-progress-stepper.css
```diff
@@ -6,29 +6,47 @@
...
-  --_step-line-color: var(--rh-color-border-strong);
-  --_current-step-color: var(--_rh-color-status-disabled, #6a6e73);
-  --_current-step-label-color: var(--rh-color-text-primary);
-  --_step-line-angle: -180deg; /* sets the line direction between steps to horizontal */
-  --_step-line-position-block: 0 100%;
-  --_step-line-position-inline: calc(calc(100% - var(--_step-icon-width)) * -1) 100%;
-  --_step-line-size-inline: calc(100% - var(--_step-icon-width));
-  --_step-line-size-block: 100%;
-  --_step-template-areas:
+  --_step-line-color: /* @summary Color of connecting lines between steps */
```
bennypowers: see previous comments re: not documenting `--_` props, documenting system tokens. A note about syntax: IIRC, cem will associate comments with the prop on the next line, not the one for which the comment is placed at the end of the line

```suggestion
  /* @summary Color of connecting lines between steps */
  --_step-line-color: 
```

if on the other hand you intended this comment to apply to `--rh-color-border-strong`, then you did it correctly. Note however that on this line for example, the comment is misplaced

```css
  --_step-line-angle: -180deg; /* Line direction between steps to horizontal */
```

elements/rh-progress-stepper/rh-progress-step.css
marionnegp: If I'm asking about progress stepper, I'd expect to see the inline documentation first to see what the property is used for in the element. Then cem can tell me about the token and maybe explain why this one is used.

elements/rh-progress-stepper/rh-progress-step.css
marionnegp: The comment was intended to explain what the variable + its fallback defined, but is this unexpected placement?

elements/rh-progress-stepper/rh-progress-step.ts
adamjohnson: I do like the idea of having the docs in one place and pushing them out everywhere. My only worry would be complexity. Oftentimes the simplest solution is the right one. 

Pros:
- Write once, publish everywhere (every web worker's utopian goal)

Cons:
- Complexity
    - There are probably weird edge cases we can't yet see
- Conflating docs and code (these are two different things)
- Our `*.ts` files would get quite long

The way we automate our Code docs does seem to work. I could see this working. Might have to make a POC / MVP to see how it will shake out.

elements/rh-progress-stepper/rh-progress-step.ts
marionnegp: I moved some of the info from here to the icon `@property`. Since the `@property` talks more about using custom icons, I'm not sure that's where general icon info would go. Should we remove this because `protected` isn't public?

elements/rh-progress-stepper/rh-progress-step.ts
marionnegp: Thanks! I ended up removing comments on the private variables.

elements/rh-progress-stepper/rh-progress-step.css
bennypowers: if the comment describes `--_current-step-weight`, then it should be removed altogether (rule: don't document private apis)

if the comment describes `--rh-font-weight-body-text-regular`, then it should be moved, as per my suggestion. rule: [CEM associates](https://github.com/bennypowers/cem/blob/f1189d1aaafee9bcd55b34f875bcaac457157c87/queries/css/cssCustomProperties.scm#L35-L50) comments with the closest syntax node.

elements/rh-progress-stepper/rh-progress-step.ts
bennypowers: protected members are relevant to anyone who's going to extend progress step - something I don't see happening.

I'd expect this method to either be simple and self-documenting enough that a comment isn't necessary, or to have a brief comment describing why it's necessary and any special considerations, side effects, etc.

in other words, not relevant to this PR

elements/rh-progress-stepper/rh-progress-step.ts
bennypowers: I'm happy to defer discussion of docsgen to a later date, but I'd still like to see guidelines mentioned here

elements/rh-progress-stepper/rh-progress-step.css
bennypowers: thinking this over, I think the fastest route to done here is:

1. cem will first add local comment docs, then append design tokens docs. I believe this is current behaviour, but i'll check if it's tested
2. we'll simply remove the (known in-advance) tokens-side documentation from the description, and extract that into a new column in the tokens table

pseudocode:
```ts
import { tokens } from '@rhds/tokens';
for (const token of tokensInElementManifestEntry) {
  const upstreamDescription = tokens.get(token.name)?.$description;
  const userDescription = (token.description??'').replace(upstreamDescription, '')
  // same for summary?
  // ... append userDescription to tokens table 
  //     in a third column "Usage in this element"
}
```

@zeroedin WDYT?

bennypowers: https://github.com/bennypowers/cem/pull/103 PTAL @adamjohnson @marionnegp

elements/rh-progress-stepper/rh-progress-step.css
bennypowers: [cem v0.4.6](https://github.com/bennypowers/cem/releases/tag/v0.4.6) is out with the feature that enables user comments on design tokens

adamjohnson: Updated CEM to v0.4.6 in #2606.

elements/rh-progress-stepper/rh-progress-step.ts
marionnegp: I asked Cursor to use content from this DP's progress stepper ux dot documentation, and check out how it formatted this one:
Screenshot 2025-09-12 at 3 22 37 PM

Still worth thinking through how to avoid bloating the element files.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -122,6 +166,11 @@ export class RhProgressStepper extends LitElement {
     }, 100);
   });
 
+  /**
+   * Initializes responsive behavior on first update.
+   * Sets mobile state based on element width,
+   * ensuring the stepper displays correctly on initial load.
+   */
```
marionnegp, line 173: These lines should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -179,6 +241,7 @@ export class RhProgressStepper extends LitElement {
       this.currentState = this.currentStep.state || '';
       this.#contentString = '';
       // Use childNodes instead of children to access both Element and Text nodes
+      // This ensures we capture all text content for accessibility
```
marionnegp, line 244: This line should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -189,6 +252,10 @@ export class RhProgressStepper extends LitElement {
     }
   }
 
+  /**
+   * Handles changes to the verticalAt property.
+   * Updates the breakpoint threshold for responsive vertical orientation switching.
+   */
```
marionnegp, line 258: These lines should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -95,14 +134,19 @@ export class RhProgressStepper extends LitElement {
 
   #maxWidth = 768;
 
-  /** normalized string content of the current step */
+  /**
+   * Normalized string content of the current step
+   * Extracts text content from the current step's title and description
+   * for screen reader accessibility and visual display.
```
marionnegp, line 140: These lines should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -95,14 +134,19 @@ export class RhProgressStepper extends LitElement {
...
+   */
   #contentString = '';
 
   #resizeTimeoutId?: number;
 
   /**
+   * ResizeObserver for responsive behavior.
+   * This callback is debounced with a simple timeout to prevent excessive updates.
```
marionnegp, line 148: These lines should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -147,23 +197,35 @@ export class RhProgressStepper extends LitElement {
 
     return html`
       <div id="container" class="${classMap({ compact, vertical, [currentState]: true })}">
+        <!-- "Current step" label for screen readers and compact display -->
+        <!-- Visually hidden except in compact mode -->
```
marionnegp, line 201: These lines should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -147,23 +197,35 @@ export class RhProgressStepper extends LitElement {
...
       </div>
     `;
   }
 
+  /**
+   * Handles change events from progress steps.
+   * Updates the current step state for accessibility and visual feedback.
+   *
+   * @param event - The change event from a progress step
+   */
```
marionnegp, line 217: These lines should be reviewed for accuracy.

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM.

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM

elements/rh-progress-stepper/rh-progress-stepper.ts
adamjohnson: LGTM

elements/rh-progress-stepper/rh-progress-step.ts
bennypowers: i don't think this is "bloat" it'll gzip up just fine, and doesn't cost much to parse at runtime. We should err on the side of better documentation, and worry about shaving bytes if and when that becomes a problem

elements/rh-progress-stepper/rh-progress-step.css
```diff
@@ -9,8 +9,10 @@
   --_step-line-style:
     linear-gradient(
         var(--_step-line-angle),
+        /* @summary Sets the angle direction for connecting line gradient */
```
bennypowers: we'll need to remove the inline comments for --_private-vars

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -47,6 +52,17 @@ const BREAKPOINTS = new Map(Object.entries({
  * - fail (failed to occur)
  * Or a custom state, set using the `icon` attribute.
  *
+ * ## Usage guidelines
+ * - Use 3-5 steps maximum to reduce cognitive load
+ * - Designed to complement standard previous/next navigation, not as navigation itself
```
bennypowers:
```suggestion
 * - Designed to complement standard previous/next navigation. Avoid using as the only navigation
```

elements/rh-progress-stepper/rh-progress-stepper.ts
```diff
@@ -147,23 +197,35 @@ export class RhProgressStepper extends LitElement {
 
     return html`
       <div id="container" class="${classMap({ compact, vertical, [currentState]: true })}">
+        <!-- "Current step" label for screen readers and compact display -->
+        <!-- Visually hidden except in compact mode -->
         <strong id="current-step"
                 class="visually-hidden"
                 ?hidden="${!compact}">${this.#contentString}</strong>
         <!-- Use this slot for \`<rh-progress-step>\` items -->
+        <!-- Each step should include title and optional description -->
```
bennypowers:
```suggestion
        <!-- Use this slot for \`<rh-progress-step>\` items
             Each step should include title and optional description -->
```

bennypowers (changes requested):

bennypowers (approved):

PR #2552, opened 2025-08-08
fix(footer): not defined color contrast

### LINKED ISSUE #2483: [bug] Intersection-observer-loaded rh-footer loads with dark text on dark background
Text in our footer at RHDC (and possibly elsewhere) is currently raising accessible contrast issues in several tools, including Lumar and WAVE:

rh-footer at RHDC showing several contrast errors in Webaim Wave

Zack H identified the issue and has an RHDC fix pending:

- [Jira ticket](https://issues.redhat.com/browse/DXFED-2784) 
- [MR code](https://gitlab.cee.redhat.com/it-marketing/rhdc/-/merge_requests/6058/diffs#33e9397becaeca6d2619323640911d974e6798ad_24_26)

> The JavaScript that powers the footer is loaded only when the user scrolls towards the bottom of the page... [Lumar and WAVE are] seeing the black text on a black background.

> I found that the default color-scheme for the page is set to light (as expected) but the footer light-dom CSS doesn't provide a color-scheme so it's getting a weird mix of light and dark (before the JS upgrades).

> I'm providing a default color AND setting color-scheme: dark to avoid this issue.

To reproduce at RHDC, you can block lazy-load-esmodule.js in the network panel and reload the homepage.

adamjohnson:
## What I did

1. Fixes #2483, which relates to color contrast of links, headings, and `<p>` tags in the footer before the component upgrades.

## Testing Instructions

1. First, load the [default footer demo](https://ux.redhat.com/elements/footer/demo/) on uxdot.
2. Open dev tools.
3. Open the command palette by hitting CMD + Shift + P.
4. Type in "Disable JavaScript"
5. Refresh the page.
6. Wait more than 5 seconds.
7. Note the nearly invisible links and text.
8. Now open the [DP demo](https://deploy-preview-2552--red-hat-design-system.netlify.app/elements/footer/demo/).
9. Repeat steps 2-6.
10. Observe much improved color contrast.
11. Re-enable JavaScript.
12. Make sure the fully-upgraded footer matches what's on uxdot.
13. Confirm these changes work for other footer variants.
14. Testers should also scan the DP with Lumar, Wave, and possibly Axe to ensure this fixes the issue flagged in those tools.

## Notes to Reviewers

Lumar, Wave, and possibly Axe were throwing a11y color contrast errors—even though, when the component upgraded, the color contrast passes. 

Somehow, these tools were scanning the page before the `deopacitize-footer` animation completed _and_ before the component upgraded. In doing this, these tools saw browser default links, h3's, and paragraph tags on a black background. These changes fix the `:not(:defined)` experience to pass a11y color contrast guidelines.


elements/rh-footer/rh-footer-lightdom.css
```diff
@@ -141,6 +142,20 @@ rh-footer:not(:defined) {
   min-height: var(--rh-footer-nojs-min-height, 750px);
 }
 
+:is(rh-footer, rh-footer-universal):not(:defined) {
+  p {
+    color: var(--rh-color-text-secondary);
```
zeroedin: You'll want to load the fallback value here for the color, unless the theming tokens are loaded directly on the page, which is unlikely. If the footer is the lone component and undefined, the theming tokens won't be available till after the JS loads.

```Suggestion
    color: 
      var(--rh-color-text-secondary, 
        light-dark(
            var(--rh-color-text-secondary-on-light, #4d4d4d),
            var(--rh-color-text-secondary-on-dark, #c7c7c7)
        ));
```

PR #2549, opened 2025-08-07
chore: rename demo index files

bennypowers:
## What I did

1. rename elements/{tagname}/demo/{tagname}.html to elements/*/demo/index.html
2. remove cem postprocess script

## Testing Instructions

1. run dev server and eleventy server (or check dp): ensure demos are properly ordered

## Notes to Reviewers
this requires the as-yet-unreleased new version of cem: https://github.com/bennypowers/cem/pull/88

PR #2555, opened 2025-08-11
docs(alert): register toast demo

bennypowers:
## What I did

1. add demo metadata to the rh-alert demos for toast variants

## Testing Instructions

1. check the dp: alert demos page should show "toast".
2. so should dev server


bennypowers:
Closing in favour of #2549

Issue #2355, opened 2025-05-05
[feat] Dark color scheme `Inline/Toast alerts`


coreyvickery:
### Description

Requested by Wes in [this discussion](https://github.com/orgs/RedHat-UX/discussions/2353).

A design is ready, but needs approval.

Note: The current Toast alert drop shadow is a bit hard to see on dark backgrounds (`gray-95` is the background color in the image below), but it is there. See for yourself in [the demo](https://ux.redhat.com/elements/alert/demos/#demo-toast).

Options to remedy:

- Create a darker box shadow token for components like dark scheme `Toast alerts`, `Tooltip`, etc. (I believe PatternFly 6 is doing this)
- Change an existing box shadow token value
- Customize the opacity without a token for dark scheme only

@marionnegp @markcaron Any other thoughts I may be missing?

### Acceptance Criteria

- [x] Design done
- [ ] Development done

### Images

<img width="1480" alt="Image" src="https://github.com/user-attachments/assets/a778fcd5-f851-48fd-a0dd-e1d3e2bc8a88" />

### Design file

[Figma link](https://www.figma.com/design/oFpobEGrlAtgMipbOICcZL/rh-alert-dark-scheme?node-id=0-1&m=dev&t=BdF602ZbkNc8gxjX-1)

markcaron:
@coreyvickery Yeah, I think @marionnegp is right. We should make a token for the darker box shadow color, and make that value darker (or as needed).

I don't believe option #3 isn't possible ATM, since the `color-scheme` and `light-dark()` won't toggle just the opacity. It'd just be simpler to do #1.

coreyvickery:
@marionnegp @markcaron Do one of you want to create this token?

marionnegp:
@coreyvickery, I can work on adding these new box shadow tokens: ([#166](https://github.com/RedHat-UX/red-hat-design-tokens/issues/166) in tokens repo)

coreyvickery:
Update 7/11: Arathy is working on this.

PR #2559, opened 2025-08-12
fix(alert): dark color scheme Inline/Toast alerts

ArathyKumar:
## What I did

1. Fixes https://github.com/RedHat-UX/red-hat-design-system/issues/2355

## Testing Instructions

In the current implementation, alerts don't adjust their background color when dark mode is enabled. As a result, they appear much lighter than the rest of the interface, unintentionally drawing attention and looking more prominent than intended.

To address this, dark theme support has now been applied to the alert component as well. Please refer to the [Figma design](https://www.figma.com/design/oFpobEGrlAtgMipbOICcZL/rh-alert-dark-scheme?node-id=1-254&m=dev) to verify that the alert's color adjusts appropriately in dark mode, as specified in the design.


elements/rh-alert/demo/toast-variant.html
bennypowers, line 1: the toast.html demo works now so i think we can remove this

ArathyKumar: Removed `toast-variant.html` file.

ArathyKumar: Added changeset and updated docs.

.changeset/brave-pumas-poke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-alert>`: add dark color theme styling
```
bennypowers:
```suggestion
`<rh-alert>`: added support for dark color schemes
```

bennypowers:
low-gamut textual messages

coreyvickery (approved):

PR #2560, opened 2025-08-12
fix(footer): reduce CLS before component upgrades or if JavaScript fails to load

adamjohnson:
## What I did

1. Reduce CLS of footer.
2. Also, subesequently, improves the layout of the footer when JS fails to load.

## Testing Instructions

1.

### DIFF
--- a/.changeset/lucky-coins-refuse.md
+++ b/.changeset/lucky-coins-refuse.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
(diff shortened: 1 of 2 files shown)


adamjohnson:
FYI this branch contains a patch for PFE Tools v5.0.4 to get lightdom rendering working on the dev server (`8000`). Once https://github.com/patternfly/patternfly-elements/pull/2944 merges, we should be able to remove this patch and merge this PR.

PR #2563, opened 2025-08-13
feat(icon): add icon gradient API

markcaron:
## What I did

1.

## Testing Instructions

1.


markcaron:
Still debating the API, since this isn't a pure CSS theme approach—which we were hoping to take with the AI styling.

PR #2582, opened 2025-08-19
fix: ai derived dev server routing corrections

zeroedin:
## What I did

**NOTE: AI helped create this code; use caution in review, and more detailed testing is warranted.**

1. Removed need for elementsDir in `pfe.config.json`
2. Fixed routing of lightdom(-shim).css files in dev server via a patch for pfe-tools@5.0.4
3. Added helper route to deal with missing `/` on common paths `/elements/footer` -> `/elements/footer/`. 
4. Added helper route to deal with missing `/demo/` after element `/elements/footer/` -> `/elements/demo/footer`

## Testing Instructions

1. Run the dev server, ensure all demos function correctly, including lightdom(-shim).css files work.  

## Notes to Reviewers

Review the patch package for pfe-tools@5.0.4. This change has been made upstream as well and verified to work on PFE https://github.com/patternfly/patternfly-elements/pull/2944.  We can merge that, remove it here, and then update the dependencies again here.


zeroedin:
Converted back to draft. The back port is not yet working. I am digging in to find a working solution.

zeroedin:
I believe I have verified this fix upstream:

Upstream fix: https://github.com/patternfly/patternfly-elements/pull/2944 

We can hold on to this PR, pulling out the patch later after updating to pfe-tools@5.0.5 if we like this approach.

Issue #2475, opened 2025-07-15
[bug] `<rh-code-block>` Show more button behaviour when there is no 'more' to show


widescreenBob:
### Describe the bug

If the show more button is present and the content isn't larger than the original window, when clicked it actually makes the  code block less tall.

![Image](https://github.com/user-attachments/assets/1e3a0e8f-de58-488a-b6f5-b6abaf022c1c)

### RHDS version number

3.0.3

### Steps to reproduce

https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/virtualization/support#virt-using-virt-must-gather_virt-collecting-virt-data

The first codeblock gets smaller when you click show more.

### Expected behaviour

The min height should likely be maintained so it doesn't shrink.

### Element HTML

```HTML

```

### Operating System (OS)

macOS

zeroedin:
I believe this was closed by https://github.com/RedHat-UX/red-hat-design-system/pull/2670. If it needs to be reopened, please reopen a new issue.

PR #2601, opened 2025-09-08
fix(code-block):  show more button bug 

Michal7290:
Resolves: https://github.com/RedHat-UX/red-hat-design-system/issues/2475 
## What I did

We saw that the problem starts when the attribute "line-numbers" has the value "hidden",
that is, the function that counts the number of lines exits as soon as it sees that the value in the attribute is "hidden", which causes the button to not show.
The button depends on the function that counts the lines to know whether to show.
Therefore, we omitted the part in the condition that causes this.
Then we checked that the change we made does not interfere with other behaviors. (Hopefully we are aware of everything)

## Testing Instructions

1. Navigate to rh-code-block where the "Show More" button is used.
2. Click on the "Show More" button.
3. Verify that additional content/items are displayed correctly.
4. Click again (if applicable) to ensure collapse works as expected.
5. Check for any UI glitches or layout issues.

## Notes to Reviewers

- Removed a condition in the "Show More" logic; behavior has been adjusted accordingly.
- Please verify that expanding/collapsing content works as expected.

### DIFF
--- a/elements/rh-code-block/demo/prerendered-prism-highlighting.html
+++ b/elements/rh-code-block/demo/prerendered-prism-highlighting.html
@@ -28,7 +28,7 @@
 <span class="token tag"><span class="token tag"><span class="token punctuation">&lt;/</span>html</span><span class="token punctuation">&gt;</span></span></code></pre>
   </rh-code-block>
 
-  <rh-code-block highlighting="prerendered" actions="copy wrap" >
+  <rh-code-block highlighting="prerendered" actions="copy wrap">
   <span slot="action-label-copy">Copy to Clipboard</span>
   <span slot="action-label-copy" hidden data-code-block-state="active">Copied!</span>
   <span slot="action-label-wrap">Toggle word wrap</span>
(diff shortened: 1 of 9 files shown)


elements/rh-code-block/rh-code-block.ts
```diff
@@ -335,7 +335,7 @@ export class RhCodeBlock extends LitElement {
    * Portions copyright prism.js authors (MIT license)
    */
   async #computeLineNumbers() {
-    if (this.lineNumbers === 'hidden' || !this.#isIntersecting) {
+    if ( !this.#isIntersecting) {
```
bennypowers: Interesting... The reason we check for 'this.lineNumber === hidden' first is an optimization for large pages. We found that this check reduces first render times on pages with hundreds of code blocks from multiple minutes to several seconds.

That being said, your investigatory instincts here are impressive - please continue to look into this, perhaps there's a faster way to do the initial check, e.g. if line numbers are hidden, then split the code string by `\n` and count the length to get the initial number of lines.

elements/rh-code-block/demo/show-more.html
bennypowers, line 1: Please add more code blocks to this demo. There should be 4 or 5 code blocks here, with 1, 2, 3, 4, and 5 lines, respectively

elements/rh-code-block/demo/prerendered-prism-highlighting.html
bennypowers, line 1: Please revert whitespace changes

Michal7290:
Closed because changes are moved to feature/my-changes PR.

PR #2604, opened 2025-09-09
feat(tabs): add `icon` and `icon-set` attributes to rh-tab

### LINKED ISSUE #2337: [feat]: `<rh-tab icon="...">`
Let's consider adding an `icon`/`icon-set` attr to rh-tab, to pair with the slot. That would release 3.1.0, so we can follow up in a later PR

_Originally posted by @bennypowers in https://github.com/RedHat-UX/red-hat-design-system/pull/2332#pullrequestreview-2780149066_

Gili-Greenberger:
Fixes #2337

## What I did
1. Updated the existing demo page (`icons-and-text.html`) for the rh-tabs component with icon support.
2. Updated `rh-tab.ts` to handle tab icons and ensure proper rendering.

## Testing Instructions
1. Open the new demo page: `elements/rh-tabs/demo/rh-tab-icon.html`.
2. Verify that:
   - Tabs with icons render correctly.
   - Existing tab functionality remains intact.

## Notes to Reviewers
- Only two files were modified in this PR.

### DIFF
--- a/elements/rh-tabs/rh-tab.css
+++ b/elements/rh-tabs/rh-tab.css
@@ -14,7 +14,8 @@
   --rh-icon-size: var(--rh-size-icon-01, 16px);
 }
 
-::slotted(rh-icon) {
+rh-icon,
+::slotted(rh-icon[slot='icon']) {
   inline-size: var(--rh-icon-size);
   block-size: var(--rh-icon-size);
   flex: none;


elements/rh-tabs/demo/rh-tab-icon.html
zeroedin, line 1: Instead of creating an additional demo, let's reuse `demo/icons-and-text.html`, and add this feature there.

.changeset/petite-garlics-roll.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+Add support for icons in rh-tab component, with a new demo page
```
zeroedin:
```Suggestion
`<rh-tab>`: added support for `icon` and `icon-set` attributes 
```

elements/rh-tabs/rh-tab.ts
```diff
@@ -79,15 +91,20 @@ export class RhTab extends LitElement {
...
-        <!-- Can contain an \`<svg>\` or \`<rh-icon>\` -->
-        <slot name="icon"
-              part="icon"></slot>
+        <slot name="icon" @slotchange=${this.#onIconSlotChange}>
+          <span part="icon">
+          ${showAttrIcon ? html`
+          <rh-icon icon=${this.icon!} set=${this.iconSet}></rh-icon>
+           ` : null}
+          </span>
+        </slot>
```
zeroedin: User-slotted content will always override the default slot content. If you add the attrs `icon` and `icon-slot` to a `rh-tab` but then slot in an icon in the same tab, the slotted content will take precedence. This is how we'd want this to function. Your code above overrides that expected functionality. Instead, let's use an ifDefined here on the attributes, hide the element completely if `icon` isn't set, and remove the need for the slotchange event.

```Suggestion
        <slot name="icon"
              part="icon">
          <rh-icon ?hidden="${!this.icon}" icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>
        </slot>
```

zeroedin: Additionally, we'll now need to consider how to approach styling the `rh-icon` now that it is in the shadow root, as it previously only existed as slotted light DOM content. We will now need to provide a default style.  

If we consult the Figma files for tabs we see that `rh-icon` was never a specification of rh-tabs. A brief history is that the icons were added as part of the alignment between `pf-tabs` and `rh-tabs` when we used an inheritance model.
 
My suggestion at this point would be to take into consideration the `1rem (16px)` font size of the rh-tab's and pick a size that looks correct. `--rh-size-icon-02, 24px` would be what I would suggest.  We'd want to include @coreyvickery or @marionnegp as a review here though to ensure we are doing the correct thing.

To override the icon size, in lightdom we can still do:

```
  rh-tab {
    --rh-icon-size: var(--rh-size-icon-03);
  }
```

marionnegp: I'd keep the icon size as `--rh-size-icon-01, 16px` (although I agree that the icon looks nicer when you use `--rh-size-icon-02, 24px`). PatternFly uses an icon SVG size that matches the text's font size, and I think the larger size gives too much focus to the icon.

elements/rh-tabs/rh-tab.ts
```diff
@@ -46,6 +49,9 @@ export class RhTab extends LitElement {
   /** True when the tab is disabled */
   @property({ reflect: true, type: Boolean }) disabled = false;
 
+  @property({ reflect: true }) icon?: IconNameFor<IconSetName>;
+  @property({ attribute: 'icon-set' }) iconSet: IconSetName = 'ui';
```
bennypowers, line 57: these require jsdoc

elements/rh-tabs/rh-tab.ts
```diff
@@ -21,6 +22,8 @@ import {
 import { themable } from '@rhds/elements/lib/themable.js';
 
 import styles from './rh-tab.css';
+import type { IconNameFor, IconSetName } from '@rhds/icons';
```
bennypowers: prefer to put type imports at the top of the file, and library imports like lit directives above local imports like styles. see other files for examples

elements/rh-tabs/rh-tab.ts
```diff
@@ -85,9 +91,10 @@ export class RhTab extends LitElement {
            part="button"
            ?disabled="${this.disabled}"
            class="${classMap({ active, box, vertical, first, last })}">
-        <!-- Can contain an \`<svg>\` or \`<rh-icon>\` -->
-        <slot name="icon"
-              part="icon"></slot>
+  <slot name="icon"
+              part="icon">
+          <rh-icon ?hidden="${!this.icon}" icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>
+        </slot>
```
bennypowers:
```suggestion
        <slot name="icon"
              part="icon">
          <rh-icon ?hidden="${!this.icon}" icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>
        </slot>    
```

whitespace

.changeset/petite-garlics-roll.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+<rh-tab>: added support for `icon` and `icon-set` attributes
```
bennypowers:
```suggestion
`<rh-tabs>`: added support for `icon` and `icon-set` attributes to the `<rh-tab>` element
```

elements/rh-tabs/rh-tab.ts
```diff
@@ -85,9 +95,10 @@ export class RhTab extends LitElement {
            part="button"
            ?disabled="${this.disabled}"
            class="${classMap({ active, box, vertical, first, last })}">
-        <!-- Can contain an \`<svg>\` or \`<rh-icon>\` -->
         <slot name="icon"
-              part="icon"></slot>
+              part="icon">
+          <rh-icon ?hidden="${!this.icon}" icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>
```
bennypowers, line 100: add something like 

```ts
@observes('icon')
protected iconChanged() {
  if (this.icon) {
    import('@rhds/elements/rh-icon/rh-icon.js');
  }
}
```

to lazy load the icon dependency when icon is set

bennypowers (changes requested):

marionnegp:
> should I update the demo to use the default icon size (--rh-size-icon-01) and add an example showing the larger one, or is it fine to leave it as is?

@Gili-Greenberger, yep, let's change the default icon size for every tab but the last one in the demo. Maybe something like this, with the tabs labeled more clearly?
Screenshot 2025-10-23 at 2 32 25 PM

Also, should I be seeing an icon in the tab labeled "Attrs + size override (icon-03)"? This is what I see.
Screenshot 2025-10-23 at 2 33 38 PM

marionnegp:
> should I update the demo to use the default icon size (--rh-size-icon-01) and add an example showing the larger one, or is it fine to leave it as is?

@Gili-Greenberger, yep, let's change the default icon size for every tab but the last one in the demo. Maybe something like this, with the tabs labeled more clearly?
Screenshot 2025-10-23 at 2 32 25 PM

Also, should I be seeing an icon in the tab labeled "Attrs + size override (icon-03)"? This is what I see.
Screenshot 2025-10-23 at 2 33 38 PM

Gili-Greenberger:
@marionnegp, thanks!
I updated and renamed the branch

elements/rh-tabs/rh-tab.css
```diff
@@ -11,6 +11,13 @@
     var(--rh-border-width-sm, 1px)
     solid
     var(--rh-color-surface);
+  --rh-icon-size: var(--rh-size-icon-01, 16px);
+}
+
+slot[name='icon'] > rh-icon {
+  inline-size: var(--rh-icon-size);
```
bennypowers, line 19: FYI: this won't style the slotted content, only the fallback content. if that's your intent, then resolve this comment

Gili-Greenberger: Updated to use ::slotted(rh-icon) so the style applies to user-provided icons rather than only fallback content.

bennypowers: you might want

```css
rh-icon,
::slotted(rh-icon[slot=icon]) {
```

just to be safe. test it both ways

Gili-Greenberger: Applied your suggestion to support both fallback and slotted icons. Seems to work well in both cases.

bennypowers (approved):

PR #2612, opened 2025-09-12
fix(code-block): accessible labels for word wrap + copy action buttons

### LINKED ISSUE #2607: [bug][a11y] `rh-code-block` toggle/copy buttons require labels
code block screencap

Flagged by Lumar: [example issue details](https://analyze.lumar.io/accounts/78957/projects/437738/crawls/6986049/reports/accessibility_issue_instances_basic/resources/10fa0701694c838b676f0bd042fd6597?source=accessibility_issues&token=c49d0142e49fbb32fd9f8ca331b07b2ebc3f757e1932231927614018d730d52a17fd548c01c8c9bba01a956102bf56f9ea519ba2216e76fc09933f41918cabe5)

[Code block demo page](https://ux.redhat.com/elements/code-block/demo/actions/)

Code block's action-copy and action-wrap buttons currently lack accessible labels. Each such label could be put on the `<button>` or on the `<svg>` within.

Notes:

- The tooltip text (e.g., "Copy to clipboard") _is_ currently read by screen readers, thanks to an `aria-live` region. So, we'll want to investigate whether labeling the button will cause this text to be read twice--and if that can be avoided.

- Using `aria-labelledby` to point to the tooltip may offer a solution. However, cross-root ARIA could rear its ugly head.

adamjohnson:
## What I did

1. Fixes #2607.
    * Adds the `silent` boolean attribute to `<rh-tooltip>` to not add content to the ARIA live region
    * Adds a `silent.html` demo
    * Use the `silent` `<rh-tooltip>` attribute in `<rh-code-block>`
    * Renames the variable `x` to `task` in `<rh-code-block>`
    * Adds two changesets (a minor and a patch)

## Testing Instructions

1. Open the [Silent demo](https://deploy-preview-2612--red-hat-design-system.netlify.app/elements/tooltip/demo/silent/) in the DP.
2. Test with your screen reader of choice (Safari + VO, et all).
3. Ensure the tooltip content is not added to the ARIA live region when the tooltip is activated
    * You may have to change the content inside the tooltip to verify this!
5. Now that you've verified `<rh-tooltip>`, go to the [Actions demo](https://deploy-preview-2612--red-hat-design-system.netlify.app/elements/code-block/demo/actions/) of Code Block.
6. Run an Axe DevTools scan, make sure the error "Buttons must have discernible text" does not show up in the results.
7. Open the actions demo in your screen reader + browser combo of choice.
8. Tab to the Word Wrap and Copy to Clipboard buttons.
9. Ensure the screen reader reads out these interactive item labels properly. 

## Notes to Reviewers

> Could you just add `aria-hidden` to the `<span>` elements inside `<rh-code-block>`/`<rh-tooltip>` and not add the `silent` attribute to `<rh-tooltip>`?

My testing with VO + Safari with this configuration lead to mixed results. Sometimes the label would be announced, sometimes it wouldn't. The solution in this PR ensures consistent behavior from AT.

### DIFF
--- a/.changeset/dark-papayas-attack.md
+++ b/.changeset/dark-papayas-attack.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-accordion>`: fix axe accessibility error flagged in `<rh-accordion-header>`.
--- a/elements/rh-accordion/rh-accordion-header.ts
+++ b/elements/rh-accordion/rh-accordion-header.ts
@@ -70,8 +70,11 @@ export class RhAccordionHeader extends LitElement {
         this.#internals.ariaLevel = heading.localName.replace('h', '');
         heading.replaceWith(this);
       } else {
-        this.#internals.ariaLevel = Math.max(2, this.#heading.level).toString();
+        if (!this.#internals.ariaLevel) {
+          this.#internals.ariaLevel = Math.max(2, this.#heading.level).toString();
+        }
       }
+      this.removeAttribute('role');
     }
   }
 


elements/rh-tooltip/rh-tooltip.ts
```diff
@@ -99,6 +99,9 @@ export class RhTooltip extends LitElement {
   /** Tooltip content. Overridden by the content slot */
   @property() content?: string;
 
+  /** When true, disables screen reader announcements for tooltip content. Only use when another accessible label is provided. */
+  @property({ type: Boolean, reflect: true }) silent = false;
```
zeroedin, line 103: Given this renders the tooltip inaccessible without other means, I suggest that we also add a warning in the documentation under guidelines. Perhaps specifically showing a code snippet on how to re-provide accessibility if it is used.  Maybe another note in the accessibility section as well?

adamjohnson: Gave this a go via [5b16972](https://github.com/RedHat-UX/red-hat-design-system/pull/2612/commits/5b1697231f0e0512c07774a72c243ed1c0724cb9). The a11y page seemed like the best location for this content.

elements/rh-tooltip/rh-tooltip.ts
bennypowers, line 1: this is a minor changeset. is there another way we could do this, without adding public features?

bennypowers (changes requested):
My gut reaction is to balk at any new APIs. But consider also #2677, which needs new APIs for when copy is cancelled (or fails) so that we can do l10n, similar to copy-label.

PR #2613, opened 2025-09-14
fix(tile): allow any color-palette and add demo page

### LINKED ISSUE #2467: [bug] `<rh-tile>` should support all color palettes
### Describe the bug

Tile should revet to supporting all color palette options again.

If this is a relatively easy fix, I'm happy to take it on, @bennypowers.

### Screenshots

_No response_

### Steps to reproduce

View current tile demo and try to use `darker` or `lighter` as a color palette. It will only show `lightest` or `darkest`.

### Expected behaviour

Tile's `color-palette` attribute should support all surface colors (except for the deprecated `dark-alt`).

### Element HTML

```HTML

```

### Operating System (OS)

macOS

### Browser

_No response_

_No response_

Gili-Greenberger:
Fixes #2467

## What I did
- Fixed a bug that caused the `<rh-tile>` component to display only two color palettes instead of all six
- Replaced `<rh-surface>` with `<rh-tile>` in `elements/rh-tile/demo/color-palettes.html`
- Added `slot="headline"` to `<h2>` elements so titles render correctly in `<rh-tile>`
- Removed the `light-dark()` default from `background-color` in the CSS to allow all six color palettes to apply properly
- **Fixed a CSS Bug:** The `background-color` property in the `rh-tile` CSS was using the `light-dark()` function as its default, which overrode the specific CSS rules for the six color palettes. I removed the `light-dark()` function and added explicit rules using the design system's CSS variables so each `color-palette` attribute applies correctly.
- Added comments in the CSS file explaining the new color-palette rules
- Verified that the fix also resolves the bug in `color-context.html`
- Chose a patch version bump because this is a bug fix restoring intended functionality, not a new feature.

## Current Status
- The `<rh-tile>` component now correctly displays all six color palettes

## Testing Instructions
1. Open the `color-palettes.html` demo page
2. Verify that all six color palettes render correctly on `<rh-tile>`
3. Check if the issue in `color-context.html` is resolved or still reproducible

## Notes to Reviewers
- Please confirm whether the `color-context.html` bug is fully resolved
- Let me know if additional fixes or refactoring are needed

Thanks

### DIFF
--- a/.changeset/breezy-gifts-report.md
+++ b/.changeset/breezy-gifts-report.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-navigation-secondary>`: adds support for `<rh-navigation-link>`
+  
--- a/.changeset/light-feet-enter.md
+++ b/.changeset/light-feet-enter.md
@@ -1,0 +1,15 @@
+---
+"@rhds/elements": minor
+---
+
+✨ Added `<rh-navigation-link>`.
+
+A styled link that is used as a slotted child of navigation elements.  Can be used in the primary, secondary, subnav, and vertical navigation elements.
+
+```html
+<rh-navigation-primary>
+  <rh-navigation-link href="/" current-page>Home</rh-navigation-link>
+  <rh-navigation-link href="/about">About</rh-navigation-link>
+  <rh-navigation-link href="/contact">Contact</rh-navigation-link>
+</rh-navigation-primary>
+```
--- a/.changeset/neat-ducks-stick.md
+++ b/.changeset/neat-ducks-stick.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-subnav>`: adds support for `<rh-navigation-link>`
+  
--- a/.changeset/tough-cows-hide.md
+++ b/.changeset/tough-cows-hide.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-pagination>`: corrected behavior using `aria-current="page"` attributes or fragment identifier URLs
+  
--- a/.changeset/yummy-banks-hug.md
+++ b/.changeset/yummy-banks-hug.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-navigation-primary>`: adds support for `<rh-navigation-link>`
+  
--- a/elements/rh-audio-player/test/rh-audio-player.spec.ts
+++ b/elements/rh-audio-player/test/rh-audio-player.spec.ts
@@ -387,6 +387,7 @@ describe('<rh-audio-player>', function() {
         });
       });
 
+
       describe('clicking rate stepdown', function() {
         beforeEach(waitForCanplaythrough);
         beforeEach(async function() {
--- a/elements/rh-navigation-link/README.md
+++ b/elements/rh-navigation-link/README.md
@@ -1,0 +1,45 @@
+# Navigation Link
+A styled link that can be used as a child of navigation elements.
+
+## Usage
+
+### Primary navigation
+
+```html
+<rh-navigation-primary>
+  <rh-navigation-link href="/" current-page>Home</rh-navigation-link>
+  <rh-navigation-link href="/about">About</rh-navigation-link>
+  <rh-navigation-link href="/contact">Contact</rh-navigation-link>
+</rh-navigation-primary>
+```
+
+### Secondary navigation
+```html
+<rh-navigation-secondary>
+  <ul slot="nav">
+    <li><rh-navigation-link href="/" current-page>Home</rh-navigation-link></li>
+    <li><rh-navigation-link href="/about">About</rh-navigation-link></li>
+    <li><rh-navigation-link href="/contact">Contact</rh-navigation-link></li>
+  </ul>
+</rh-navigation-secondary>
+```
+
+### Subnav navigation
+
+```html
+<rh-subnav>
+  <rh-navigation-link href="/" current-page>Home</rh-navigation-link>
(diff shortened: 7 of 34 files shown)


bennypowers (changes requested):
Very nice, a few notes.

In the PR description, in the what I did section, it's important to be more specific about the changes that you made, so it's not enough to just say I fixed the code according to what I thought was right. You have to say what changes you made.

> I took away the options from the colour palette decorator to make it use the default that way any colour palette would be allowed. I also added a new demo to demonstrate the feature.

In some ways, this is even more important than the actual code, because it tells your teammates what to look for when they do the review.

With regard to the title of the PR, I think we have to make some slight adjustments. Instead of feat(colour-palettes), we probably should call this something like fix(tile). For the reasons why and an explanation of what the different parts of our commit messages mean, please take a look at the wiki.

Lastly, since this PR doesn't just change the demo files, it actually changes the code of our elements that we'll be shipping to NPM. We'll need to add a change set at the patch level, which describes what this PR accomplishes, in terms that are relevant to end users.

Aside from that, this is a well-prepared PR, very focused and clean. It's clear what the PR does, and it doesn't make any unnecessary changes. Nice job.

cc @zeroedin any other notes? Do you think this is maybe a minor release instead of a patch?

elements/rh-tile/demo/color-palettes.html
mguetta1, line 1: @Gili-Greenberger Thanks for the PR!
Please use `rh-tile` for the demo

mguetta1 (changes requested):

elements/rh-tile/rh-tile.css
```diff
@@ -345,3 +335,74 @@ svg {
...
+:host([color-palette='lightest']) {
+  background-color: var(--rh-color-surface-lightest);
+  color: var(--rh-color-text-primary-on-light);
+
+  --_interactive-color: var(--rh-color-border-interactive-on-light);
+  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-light);
+  --_text-color-secondary: var(--rh-color-text-secondary-on-light);
+  --_disabled-background-color: var(--rh-color-surface-light);
+  --_border-color: var(--rh-color-border-subtle-on-light);
+}
```
zeroedin: We can minify the amount of code in these selectors by using an attribute selector and the [prefix selector ^](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors#attrvalue_4)

```Suggestion
:host([color-palette^='dark']) {
  --_interactive-color: var(--rh-color-border-interactive-on-dark);
  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-dark);
  --_text-color-secondary: var(--rh-color-text-secondary-on-dark);
  --_disabled-background-color: var(--rh-color-surface-dark);
  --_border-color: var(--rh-color-border-subtle-on-dark);

  color: var(--rh-color-text-primary-on-dark);
}

:host([color-palette^='light']) {
  --_interactive-color: var(--rh-color-border-interactive-on-light);
  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-light);
  --_text-color-secondary: var(--rh-color-text-secondary-on-light);
  --_disabled-background-color: var(--rh-color-surface-light);
  --_border-color: var(--rh-color-border-subtle-on-light);

  color: var(--rh-color-text-primary-on-light);
}

:host([color-palette='darkest']) {
  background-color: var(--rh-color-surface-darkest);
}

:host([color-palette='darker']) {
  background-color: var(--rh-color-surface-darker);
}

:host([color-palette='dark']) {
  background-color: var(--rh-color-surface-dark);
}

:host([color-palette='light']) {
  background-color: var(--rh-color-surface-light);
}

:host([color-palette='lighter']) {
  background-color: var(--rh-color-surface-lighter);
}

:host([color-palette='lightest']) {
  background-color: var(--rh-color-surface-lightest);
}
```

.changeset/witty-toys-sell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
```
zeroedin: I would suggest this is a `minor` as we are modifying the expected api and adding additional features.

bennypowers: should probably rebase this to doduo

Gili-Greenberger: I changed the changeset from patch to minor,
and updated the PR to target the doduo branch instead of main.

.changeset/witty-toys-sell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+fix: The <rh-tile> component now correctly displays all six color palettes.
```
bennypowers: see CHANGELOG.md for examples of minor changesets, and make sure to wrap tagnames in backticks

elements/rh-tile/demo/color-palettes.html
```diff
@@ -0,0 +1,34 @@
+<rh-tile color-palette="darkest">
+  <h2 slot="headline">Darkest</h2>
+  <rh-cta><a href="#">Call to Action</a></rh-cta>
```
bennypowers:
```suggestion
  <rh-cta href="#">Call to Action</rh-cta>
```

etc

bennypowers (changes requested):
a few small changes

.changeset/witty-toys-sell.md
```diff
@@ -2,5 +2,5 @@
 "@rhds/elements": minor
 ---
 
-fix: The <rh-tile> component now correctly displays all six color palettes.
+fix: The `<rh-tile>` component now correctly displays all six color palettes.
```
zeroedin:
```Suggestion
`<rh-tile>`:  allows tile to accept all six color palettes
```

PR #2617, opened 2025-09-18
fix(code-block): show more button bug

Michal7290:
**Resolves**: https://github.com/RedHat-UX/red-hat-design-system/issues/2475

**What I did**

We saw that the problem starts when the attribute line-numbers has the value hidden. Originally, the function that counts the number of lines exited as soon as it saw hidden, which caused the "Show More" button not to appear. The button depends on the function that counts the lines to know whether to show.
Instead of removing this part of the condition, we modified it so that it now returns an array according to the number of lines even when the attribute is hidden. This way, the function still knows the correct number of lines and the button behaves correctly.
We then verified that this change does not interfere with other behaviors.

**Testing Instructions**

Navigate to rh-code-block where the "Show More" button is used.
Click on the "Show More" button.
Verify that additional content/items are displayed correctly.
Click again (if applicable) to ensure collapse works as expected.
Check for any UI glitches or layout issues.

**Notes to Reviewers**

Adjusted the "Show More" logic to correctly handle line-numbers="hidden" without removing the condition.
Please verify that expanding/collapsing content works as expected.

### DIFF
--- a/.changeset/wise-rings-yell.md
+++ b/.changeset/wise-rings-yell.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+Fix "Show More" button not appearing when line-numbers="hidden"
+  
--- a/elements/rh-code-block/demo/actions-i18n.html
+++ b/elements/rh-code-block/demo/actions-i18n.html
@@ -7,15 +7,6 @@
   display: block;
   background-color: var(--rh-color-surface-lighter, #f2f2f2);
   border: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle-on-light, #c7c7c7);
-  border-block-start-width: var( --rh-code-block-border-block-start-width, var(--rh-border-width-sm, 1px));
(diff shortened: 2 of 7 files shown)


.changeset/wise-rings-yell.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+Fix "Show More" button not appearing when line-numbers="hidden"
```
bennypowers, line 5: see the format for changesets from previous prs
```suggestion
`<rh-code-block>`: ensure "Show More" buttons appears when `line-numbers="hidden"`
```

elements/rh-code-block/demo/actions-i18n.html
bennypowers, line 1: instead of altering the existing demos, we should add a new demo which illustrates the problem. since we already have one in this PR, these changes should be reverted.

bennypowers (changes requested):
We have some lint errors here. be sure to add eslint to your editor

PR #2624, opened 2025-09-22
feat: add `<rh-navigation-link>`

zeroedin:
## What I did

1. Added `<rh-navigation-link>`
2. Added support for `<rh-navigation-link>` in `<rh-navigation-primary>`
3. Added support for `<rh-navigation-link>` in `<rh-navigation-secondary>`
4. Added support for `<rh-navigation-link>` in `<rh-subnav>`

## Testing Instructions

1.  View Deploy preview

## Notes to Reviewers

This PR needs reviewed and merged before `<rh-navigation-vertical>` as navigation vertical depends on `<rh-navigation-link>`


elements/rh-audio-player/test/rh-audio-player.spec.ts
bennypowers, line 1: revert these

elements/rh-navigation-link/README.md
bennypowers, line 1: needs docs

elements/rh-navigation-link/demo/in-nav-secondary.html
```diff
@@ -0,0 +1,18 @@
+<rh-navigation-secondary>
+   <ul slot="nav" aria-labelledby="logo-id">
+    <li><rh-navigation-link href="#link" slot="links">AI</rh-navigation-link></li>
+    <li><rh-navigation-link href="#link-2" slot="links">Hybrid Cloud</rh-navigation-link></li>
+    <li><rh-navigation-link href="#link-3" slot="links">Products</rh-navigation-link></li>
+    <li><rh-navigation-link slot="links"><a href="#link-4" aria-current="page">Learn</a></rh-navigation-link></li>
+    <li><a href="#">Link</a></li>
+  </ul>
```
bennypowers, line 8: I'm just spitballing here, but what if...
```html
<rh-navigation-secondary>
   <rh-navigation-link slot="nav" href="#link">AI</rh-navigation-link>
   <rh-navigation-link slot="nav" href="#link-2">Hybrid Cloud</rh-navigation-link>
   <rh-navigation-link slot="nav" href="#link-3">Products</rh-navigation-link>
   <rh-navigation-link slot="nav" href="#link-4" current>Learn</rh-navigation-link>
   <a slot=nav href="#">Link</a>
```

maybe there are x-root aria issues here? maybe we couldn't do it without breaking backwards...

elements/rh-navigation-link/test/rh-navigation-link.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      list = await fixture(html`<ul>${element}</ul>`);
+    });
+    beforeEach(async () => await element.updateComplete);
+
+    it('is accessible', async function() {
+      await expect(list).to.be.accessible();
+    });
+
+    it('should have internals role of listitem', async function() {
+      expect(element.role).to.equal('listitem');
```
bennypowers: this won't test internals. element.role is public. this.#internals.role is the default

elements/rh-navigation-primary/rh-navigation-primary-lightdom.css
```diff
@@ -55,4 +57,56 @@ rh-navigation-primary {
       }
     }
   }
+
+  & rh-navigation-link {
+    &[slot='links'] {
+      /*
```
bennypowers: remove comments

elements/rh-navigation-secondary/rh-navigation-secondary-lightdom.css
```diff
@@ -189,6 +194,37 @@ rh-navigation-secondary-menu-section > [slot='header'] {
   margin: 0 0 var(--rh-space-lg, 16px);
 }
 
+rh-navigation-secondary > [slot='nav'] > li > rh-navigation-link {
+  --_navigation-link-container-align-items: center;
```
bennypowers, line 198: this is the way :+1:

elements/rh-navigation-link/rh-navigation-link.ts
```diff
@@ -0,0 +1,100 @@
...
+  @property({ attribute: 'icon-set' }) iconSet?: IconSetName;
+
+  /** The URL to navigate to when the link is clicked */
+  @property({ reflect: true }) href?: string;
+
+  /**
+   * Optional current-page attribute that, highlights the item as the current page.
+   * Used to pass the aria-current attribute to the anchor element when the <a> is not slotted.
+   */
+  @property({ attribute: 'current-page', type: Boolean }) currentPage? = false;
```
bennypowers, line 45:
```suggestion
  @property({ type: Boolean }) current = false;
```

elements/rh-navigation-link/rh-navigation-link.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  /** Icon set for the `icon` property - 'ui' by default */
+  @property({ attribute: 'icon-set' }) iconSet?: IconSetName;
+
+  /** The URL to navigate to when the link is clicked */
+  @property({ reflect: true }) href?: string;
+
+  /**
+   * Optional current-page attribute that, highlights the item as the current page.
+   * Used to pass the aria-current attribute to the anchor element when the <a> is not slotted.
```
bennypowers:
```suggestion
   * Indicates that this link refers to the current page.
   * Used it's important to set this attribute for accessibility purposes
```

elements/rh-navigation-link/rh-navigation-link.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  firstUpdated(): void {
+    /**
+     * SSR Adds the role, but then removes when ElementInternals is hydrated
+     * However, axe-dev tools then complains as it doesn't handle Internals correctly
+     * So.... lets readd it for brevity, then when axe decides to fix their stuff,
+     * we can remove at a later date.
+     */
+    this.role = 'listitem';
+  }
```
bennypowers: better to fix this out of band and come back here if we really need to.

elements/rh-navigation-link/rh-navigation-link.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  override async scheduleUpdate() {
+    if (this.icon) {
+      await import('@rhds/elements/rh-icon/rh-icon.js');
```
bennypowers, line 49: :man_cook: :kissing_heart:

elements/rh-navigation-link/rh-navigation-link.ts
```diff
@@ -0,0 +1,100 @@
...
+          <a href="${ifDefined(this.href)}" aria-current="${ifDefined(isCurrentPage)}">
+            <!--
+              Use this slot when the \`icon\` and \`icon-set\` attributes are not set. 
+              Can contain a rh-icon, svg, or img tag.
+            -->
+            <slot name="icon">
+              ${this.icon ?
+                html`<rh-icon icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>`
+                : html``
+              }
```
bennypowers:
```suggestion
              ${!this.icon ? '' : html`
              <rh-icon icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>
              `}
```

nit: reads better when the false case is first

elements/rh-navigation-link/rh-navigation-link.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+    super.scheduleUpdate();
+  }
+
+  render(): TemplateResult<1> {
+    const isCurrentPage = this.currentPage ? 'page' : undefined;
+
+    return html`
+      <div id="container">
+        ${this.href ? html`
```
bennypowers, line 59: we might encounter ssr hydration problems with this conditional template. have you tested it?

bennypowers (changes requested):
codewise this looks pretty much solid. a few nitpicks here and there
Docs are important here, though

elements/rh-navigation-link/demo/in-nav-secondary.html
zeroedin: > maybe we couldn't do it without breaking backwards...

I assumed this would be the next step. With secondary nav currently being redesigned, it's likely already a breaking change, and we could remodel the API in the form of primary nav, as you are suggesting here.

elements/rh-navigation-primary/rh-navigation-primary-lightdom.css
zeroedin: There are still some missing styles I need to resolve here. Thanks for pointing this back out. The hamburger's mobile link styles aren't quite right yet.

elements/rh-navigation-link/test/rh-navigation-link.spec.ts
zeroedin: Correct, but because the element is readding the value back this does work.  This is incorrect behavior on both fronts related to the lit ssr-client bug.  This test will be skipped (due to snapshot not also being able to read the role).

elements/rh-navigation-link/rh-navigation-link.ts
zeroedin: This was a workaround due to the lit ssr-client bug; this code will be removed.

elements/rh-navigation-link/rh-navigation-link.ts
zeroedin: Good call out, i'll give that a deeper dive.

elements/rh-navigation-link/rh-navigation-link.ts
zeroedin: forgot to split my commit before pushing, this change happened here: https://github.com/RedHat-UX/red-hat-design-system/pull/2624/commits/0520f23d898bda7c2a974e52d07f736ccaea5cf2

elements/rh-navigation-link/rh-navigation-link.ts
zeroedin: As far as I can tell, after turning on SSR, this doesn't seem to pose a problem.

elements/rh-navigation-link/docs/00-overview.md
adamjohnson, line 1: Docs needed here.

elements/rh-navigation-link/demo/index.html
adamjohnson, line 1: Since these are list items, should we consider wrapping these in a `<ul>` or `<ol>` here?

elements/rh-navigation-link/demo/index.html
```diff
@@ -0,0 +1,6 @@
+<rh-navigation-link href="#link">Link</rh-navigation-link>
+<rh-navigation-link><a href="#link">Slotted a href Link</a></rh-navigation-link>
+<rh-navigation-link icon="auto-light-dark-mode-fill" icon-set="ui" href="#ai">Containers</rh-navigation-link>
```
adamjohnson: Should this display an icon? If we look at [the demo](https://deploy-preview-2624--red-hat-design-system.netlify.app/elements/navigation-link/demo/), it does not currently.

elements/rh-navigation-link/demo/in-nav-primary.html
```diff
@@ -0,0 +1,427 @@
...
+      </div>
+    </div>
+  </rh-navigation-primary-item>
+
+  <rh-navigation-primary-item>
+    <a href="#top-level">Top Level</a>
+  </rh-navigation-primary-item>
+
+  <rh-navigation-link href="#top-level">Top Level</rh-navigation-link>
+  <rh-navigation-link><a href="#top-level-2">Top Level 2</a></rh-navigation-link>
```
adamjohnson, line 199: It appears that `<rh-navigation-primary-item>` on load adds `role="listitem"`.

Axe is complaining "Certain ARIA roles contain particular children". 

This is probably related to [this talk](https://github.com/w3c/aria/issues/2663#issuecomment-3423952742), but, seeing as we're adding `role` attrs to `<rh-navigation-primary-item>`, it seems appropriate we should add them to `<rh-navigation-link>` as well—if only to satisfy Axe / Lumar / et all.

.changeset/breezy-gifts-report.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-navigation-secondary>`: adds support for rh-navigation-link
```
adamjohnson:
```suggestion
`<rh-navigation-secondary>`: adds support for `<rh-navigation-link>`
```

If you accept this change, apply it to the other changesets too.

elements/rh-navigation-link/demo/index.html
zeroedin: It does locally in dev.  I'll have to look into this deeper tomorrow.

elements/rh-navigation-link/demo/in-nav-primary.html
zeroedin: Yup.  I haven't removed it yet until we decide the way forward.

elements/rh-navigation-link/demo/index.html
zeroedin: No, the parent (navigations) element are role="list" internally in the shadowroot, and the rh-navigation-links are list items. This is the purpose of the element internals and the custom element acting as that api surface to accessible technology; they should look and function the same. The idea is to limit the implementation details for the user, which reduces the need for lightdom.css (no deeply nested content) and overall less surface area to get wrong while maintaining the correct accessibility semantics. This is the same way items/dropdowns work in rh-navigation-primary.

elements/rh-navigation-link/demo/index.html
zeroedin: Ended up being an incorrect icon name, none exists with that name and `-fill`

This was the breaking change from `@rhds/icons@2.0.0`

>Removed auto-light-dark-mode-fill UI icon. Please use auto-light-dark-mode instead.

elements/rh-navigation-link/demo/in-nav-primary.html
zeroedin: So, to explain in more detail, `<rh-navigation-primary-item>` does, in fact, still add a role on firstUpdate. Unless we change rh-navigation-primary to fix that, it is doing the very thing we are trying not to repeat here in `<rh-navigation-link>`, namely reapplying those roles.

https://github.com/RedHat-UX/red-hat-design-system/blob/355d53926cfacb3e013efaa86e5b3008994505ea/elements/rh-navigation-primary/rh-navigation-primary-item.ts#L68-L87

This was done due to the lit ssr-client bug and axe-core issues all coming together and adding confusion. I added a comment on the reasoning on line 80. Now, in retrospect, as we learn more, we find out it is a much deeper issue on that end.

So, if we go the route we are going as is done in this PR. We only add Element Internals and do not re-add the role. I realize this causes issues with Lumar / local automation testing and the CI test suite. In the test suite, we are now using the ignoreRules option:

https://github.com/RedHat-UX/red-hat-design-system/blob/c000cec2c5313d8783c43889c707e40ba67f53ab/elements/rh-navigation-link/test/rh-navigation-link.spec.ts#L47-L52

For Lumar, though, I don't know what's possible to mute or reduce the error level when it's associated with a custom element. @hellogreg, let me know if there is anything I could help with there.  

For localized testing in the Axe web extension, what else can we do besides manually testing and notifying people about the false positive?  Unfortunately, this isn't "our" issue per se, but it's definitely exposed in our stuff.

zeroedin: @bennypowers do you think we should open an issue on `rh-navigation-primary` to remove the re-adding of roles at this time?

hellogreg: Though we can't mute/omit issues in Lumar from the general dashboard, I _can_ filter them out for the reports I use and share, for now. (I met with Lumar this week to ask for some improved crawling/reporting/filtering options. But that's all TBD.)

As for anything else axe-core (e.g., the browser extension), yeah, we may have to let people know--if we're not going to be consistent in the application of role attributes to both parents and children (as Adam mentioned above).

bennypowers: sounds like either 

- nav1 should go all in and use internals exclusively for roles
- nav1 should use a `<nav>` in shadow root (if that's otherwise kosher)
- we remove elementinternals roles from everyone

I'm partial to the first option, but it does require downstream to ignore those errors until axe provides an escape hatch.


given that adam's approved, let's merge this PR

PR #2631, opened 2025-09-26
feat(disclosure) add `open` and `compact` variants

### LINKED ISSUE #2574: [feat] `<rh-disclosure>` needs an `open` variant
### Description

We need an `open` variant of `<rh-disclosure>` that is a simple expression of the pattern without the border and panel.

### Screenshots

For reference, [here's PF's](https://www.patternfly.org/components/expandable-section):

Image

### Example API

```HTML
<rh-disclosure variant="open"></rh-disclosure>
```

### Additional context

_No response_

adamjohnson:
## What I did

1. Closes #2574.

## Testing Instructions

1. Forthcoming

### DIFF
--- a/elements/rh-disclosure/rh-disclosure.ts
+++ b/elements/rh-disclosure/rh-disclosure.ts
@@ -68,9 +68,14 @@ export class RhDisclosure extends LitElement {
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
 
   /**
-   * Sets the disclosure to be in its open state
+   * Sets the disclosure to be in its open (expanded) state
    */
   @property({ type: Boolean, reflect: true }) open = false;
+
+  /** Removes the outer and left border from the disclosure.
+   * The background is `surface-light`/`surface-dark` when expanded.
+   */
+  @property({ reflect: true }) variant?: 'open';
 
   /**
    * Sets the disclosure title via an attribute
--- a/elements/rh-disclosure/rh-disclosure.ts
+++ b/elements/rh-disclosure/rh-disclosure.ts
@@ -72,10 +72,11 @@ export class RhDisclosure extends LitElement {
    */
   @property({ type: Boolean, reflect: true }) open = false;
 
-  /** Removes the outer and left border from the disclosure.
+  /** **Open:** Removes the outer and left border from the disclosure.
    * The background is `surface-light`/`surface-dark` when expanded.
+   * **Compact:** decreases disclosure padding.
    */
-  @property({ reflect: true }) variant?: 'open';
+  @property({ reflect: true }) variant?: 'open' | 'compact';
 
   /**
    * Sets the disclosure title via an attribute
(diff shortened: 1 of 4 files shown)


marionnegp (changes requested):
Thanks for getting a demo ready so quickly!

**Compact, box/default disclosure**
- [x] The font size for `summary` should actually be `--rh-font-size-body-text-sm`. That was my mistake in the Figma file!
- [x] Caret icons should be 14px. (I don't think we have an icon size token that small yet.)

**Open disclosure (both sizes)**
- [x] Font color for `summary` slot should be `--rh-color-interactive-primary-default` in default state and `--rh-color-interactive-primary-hover` for all other states.
- [x] Focus ring should be offset from edge of the `summary` by 2px.
- [ ] `Border-radius` should be `--rh-border-radius-default`
- [ ] For dark scheme, if the disclosure is collapsed, the hover state background color should be `--rh-color-surface-darker`.

**Open, compact disclosure**
- [x] Caret icons should be 14px
- [x] The font size for `summary` should be `--rh-font-size-body-text-sm`

marionnegp (approved):
Looks great!

elements/rh-disclosure/rh-disclosure.ts
```diff
@@ -68,10 +68,16 @@ export class RhDisclosure extends LitElement {
...
-   * Sets the disclosure to be in its open state
+   * Sets the disclosure to be in its open (expanded) state
    */
   @property({ type: Boolean, reflect: true }) open = false;
 
+  /** **Open:** Removes the outer and left border from the disclosure.
+   * The background is `surface-light`/`surface-dark` when expanded.
+   * **Compact:** decreases disclosure padding.
+   */
+  @property({ reflect: true }) variant?: 'open' | 'compact';
```
zeroedin: Because we already have an open attribute I worry that "open" variant is confusing for the user.  Could we instead rename this to `variant="borderless"`

adamjohnson: This is a good point. I will make this change in code and update this PR. @marionnegp WDYT?

adamjohnson: Implemented in [49af914](https://github.com/RedHat-UX/red-hat-design-system/pull/2631/commits/49af91442c6e7e342cc706387925540a33c82a26).

marionnegp: Darn. I'm guessing that changing the current `open` attribute would be a breaking change. `Borderless` is fine if that's the case.

adamjohnson: Yes, changing the current `open` boolean attribute would be a breaking change. That `open` attribute also exists in (almost) all of the navigation components. It's rooted in the `<details open>` element as a [browser API](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details#attributes). Keeping that naming convention congruent is a good idea.

Mark pointed out that we should try to keep the naming convention for this "borderless" style congruent. `rh-pagination` is the only other component that has `variant="open"`. Since we're changing this to `borderless`, we should change that API too. In the case of pagination, we might be able to alias `open` to `borderless` and deprecate the `open` variant at the next major version as a breaking change. I will open an issue to track that.

adamjohnson: Issue #2637 has been opened for `<rh-pagination>`. I'm going to resolve this conversation here.

marionnegp:
> Mark noted the rounded corners on the borderless variant versus the default <rh-disclosure> does not have rounded corners. Can you confirm that this is intented and the design style we want moving forward? (it very well may be)

Ah, good point. Can we go ahead and make the border-radius for all of them use `--rh-border-radius-default`, if it's not a breaking change?

PR #2635, opened 2025-09-30
docs: stable gutter sidenav

zeroedin:
## What I did

1. After the change to the overflow auto in #2630, the sidenav contents bounce when a item dropdown is opened causing the menu to overflow.  Corrected this by adding stable to scrollbar-gutter.
2. Correct border placement for sidenav. It has always been a mistake to have the border on the main element instead of the sidenav itself. This issue is now more noticeable in dark mode than it was in light mode. If this is deemed a wrong assumption, I will revert.

## Testing Instructions

1. [View the Deploy preview](https://deploy-preview-2635--red-hat-design-system.netlify.app)https://deploy-preview-2635--red-hat-design-system.netlify.app/

### DIFF
--- a/docs/elements/demo.html
+++ b/docs/elements/demo.html
@@ -24,10 +24,7 @@
     <script async src="https://ga.jspm.io/npm:es-module-shims@1.8.1/dist/es-module-shims.js"></script>
   </head>
   <body>{% if demo.filePath %}
-    <main>
-      {%- include demo.filePath -%}
-    </main>{%- else -%}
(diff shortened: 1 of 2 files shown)


marionnegp (changes requested):
If my Mac is set to always show scroll bars, I see a gap between the gray background of an active nav item and the side nav border, before the side nav overflows.
Screenshot 2025-10-07 at 12 45 07 PM
Screenshot 2025-10-07 at 12 45 02 PM

Otherwise, the updates look good to me. Nice catch with moving the border to the side nav.

PR #2636, opened 2025-10-01
fix(pagination): input on change, aria-current functionality

zeroedin:
## What I did

1.  Correctly setting the`event.target` value for input onChange.
2. Added type="number" to input.  Can discuss if this is something we want to do. I think it does given the use of min and max values are only valid if the `type="number"` is present. 
> HTML attribute: min 
It is valid for the input types including: [date](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/date), [month](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/month), [week](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/week), [time](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/time), [datetime-local](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/datetime-local), [number](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number) and [range](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range) types, and the [&lt;meter&gt;](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meter) element.
3. Updated demos to be a bit more helpful.
4. Correctly use `aria-current="page"` for setting `#currentLink` on initial load.

### TODO:

- [x] Fix Docs issue around `#`
- [x] Add Changeset

## Testing Instructions

1. View [Deploy Preview demos](https://deploy-preview-2636--red-hat-design-system.netlify.app/elements/pagination/demos/)

## Notes to Reviewers

I have questions about listening and updating to hashchange for all paginations.  If you were to have multiple paginations or a pagination and a jump links on the same page you could have incorrect behavior.   Also hash should not have been  a requirement of pagination as I understand it, this was a mistake in the documentation.  I would think supporting hashchange would have to be handled as an integration piece allowing you to target the specific pagination you want to update.

This PR also attempts to correct the `aria-current="page"` usage.  It is of my opinion that when the page load the server is the authority.  If it set `aria-current="page"` on a slotted link that link should become the selected link.  However the component was actively removing this server set value on load (in updated) and clearing that preference.  This appears to have been done in order to again support the above mentioned `hashchange`.

### DIFF
--- a/elements/rh-pagination/docs/30-code.md
+++ b/elements/rh-pagination/docs/30-code.md
@@ -2,7 +2,7 @@
 
 ### Basic Structure
   * `<rh-pagination>` must have a nested `<ol>` element with at least one nested `<li>` and `<a>` element.
-  * Each `<a>` element should have an `href` attribute that links to the corresponding page.
+  * Each `<a>` element must have an `href` attribute that links to the corresponding page.
 
 ### URL Structure Options
 You can implement pagination URLs using any of these approaches:


adamjohnson:
Once #2644 merges, we should add demo descriptions for Pagination to this PR to guide users about `aria-current`/`active` not changing without a page refresh on some demos. That'll probably ward off any "pagination is broken" future issues.

elements/rh-pagination/demo/vue.html
```diff
@@ -0,0 +1,61 @@
...
+      </li>
+    </ol>
+  </rh-pagination>
+
+  <span v-text="message"></span> <span v-text="currentPage"></span>
+</div>
+
+<script>
+  // eslint-disable-next-line no-undef
+  const { createApp } = Vue;
```
bennypowers: where is this loaded from? can you add it to the import map instead and explicitly import it here (which would also clear the lint error)?

zeroedin: fixed in [7308bff](https://github.com/RedHat-UX/red-hat-design-system/pull/2636/commits/7308bff8a722634ee01170c535bf361fb2b04e88)

bennypowers (changes requested):
Needs a changeset

package.json
```diff
@@ -355,6 +355,7 @@
     "tsx": "^4.19.3",
     "typescript": "^5.8.2",
     "typescript-transform-lit-css": "^2.0.2",
+    "vue": "^3.5.22",
```
bennypowers: this is ok for now if we just need to merge quickly

when we get some breathing room, it would be better to remove this from devDeps and use a CDN link for the demo

markcaron:
@zeroedin Looks good!

Lets Get This Merged!

PR #2646, opened 2025-10-06
fix(navigation-primary): hide `role=list` elements when slots are empty

zeroedin:
## What I did

1. Hide `role="list"` elements when their slots are empty.

## Testing Instructions

1. View demo [No Secondary Items](https://deploy-preview-2646--red-hat-design-system.netlify.app/elements/navigation-primary/demos/#demo-no-secondary-items)

## Notes to Reviewers

The default slot, although it is also a role="list" element with a slot, is not intentionally hidden when empty.  It was decided that this slot should never be empty, but instead, it would be an implementation error. This may be the case for a future `rh-masthead` element that abstracts the "bar" and what is slotted inside that bar to a parent component.

### DIFF
--- a/elements/rh-navigation-primary/rh-navigation-primary.ts
+++ b/elements/rh-navigation-primary/rh-navigation-primary.ts
@@ -81,6 +81,10 @@ export class RhNavigationPrimary extends LitElement {
   @query('summary')
   private _hamburgerSummary!: HTMLElement;
 
+  @query('#title')
+  private _title!: HTMLHeadingElement;
+
+
   /**
    * Sets the mobile toggle (hamburger) text, used for translations, defaults to 'Menu'
    */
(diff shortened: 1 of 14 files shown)


elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -158,6 +158,10 @@ export class RhNavigationPrimary extends LitElement {
       compact,
       dehydrated: !this.#hydrated,
     };
+    const hasEvent = this.#slots.hasSlotted('event');
```
adamjohnson, line 170: Looks like `event` needs added to the SlotController array on line 54.

zeroedin: good catch

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -185,25 +190,24 @@ export class RhNavigationPrimary extends LitElement {
               <!-- Use this slot for \`<rh-primary-navigation-item>\` hamburger menu links and dropdowns -->
               <slot></slot>
             </div>
-
           </details>
           <div id="secondary">
             <!--
               Use this slot for event promotion.  Images such as SVGs and links are most often slotted here.
```
zeroedin: Adjust the documentation to include that anything slotted in event, links, dropdowns must be of `role="listitem` `rh-navigation-primary-item` already set themselves to role="listitem".  However, if you slot something other than that component, it will be incorrect without an explicit `listitem` role.

PR #2648, opened 2025-10-06
Fix/navigation primary/hidden title slot

zeroedin:
## What I did

1. Correct accessible label usage by implementing an internal hidden header with aria-labelledby through element internals.

## Testing Instructions

1.


elements/rh-navigation-primary/rh-navigation-primary.css
```diff
@@ -107,6 +107,19 @@
   }
 }
 
+.visually-hidden {
+  position: absolute !important;
+  height: 1px;
+  width: 1px;
```
adamjohnson:
```suggestion
  block-size: 1px;
  inline-size: 1px;
```

Also consider alphabetizing these properties.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -175,6 +183,7 @@ export class RhNavigationPrimary extends LitElement {
               </a>
             </slot>
           </div>
+          <h2 id="title" class="visually-hidden">${this.accessibleLabel}</h2>
```
adamjohnson: Since this is visually hidden, would you be opposed to moving this `<h2>` between `#container` and `#bar` (lines 170-171)?

We're doing similar things on uxdot, keeping the headings referenced by `aria-labelledby` close to top level in the shadowdom.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -89,10 +93,10 @@ export class RhNavigationPrimary extends LitElement {
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: NavigationPrimaryPalette;
 
   /**
-   * Customize the default `aria-label` on the `<nav>` container.
-   * Defaults to "main" if no attribute/property is set.
+   * Customize the default label for the navigation.
+   * Defaults to "Navigation" if no value is set.
    */
-  @property({ attribute: 'accessible-label' }) accessibleLabel = 'main';
+  @property({ attribute: 'accessible-label' }) accessibleLabel = 'Navigation';
```
adamjohnson:
```suggestion
  @property({ attribute: 'accessible-label' }) accessibleLabel = 'Main navigation';
```

Screen readers already read "navigation" due to the `<nav>` / ARIA internals. That can cause readings like "Navigation navigation".

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -89,10 +93,10 @@ export class RhNavigationPrimary extends LitElement {
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: NavigationPrimaryPalette;
 
   /**
-   * Customize the default `aria-label` on the `<nav>` container.
-   * Defaults to "main" if no attribute/property is set.
+   * Customize the default label for the navigation.
+   * Defaults to "Navigation" if no value is set.
```
adamjohnson: (see comment below first)

If you like what's below:

```suggestion
   * Customize the default label for the navigation.
   * Defaults to "Main navigation" if no value is set.
```

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -80,6 +80,10 @@ export class RhNavigationPrimary extends LitElement {
   @query('summary')
   private _hamburgerSummary!: HTMLElement;
 
+  @query('#title')
+  private _title!: HTMLElement;
```
adamjohnson: Would `HTMLHeadingElement` work / be more appropriate here?

[MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/HTMLHeadingElement)

zeroedin: Good catch at first this was on a div allowing user to change the heading level, decided to static the heading level then didn't change the type.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: How about before `#container`, as technically it labels the `:host`?

adamjohnson: Works for me.

PR #2626, opened 2025-09-23
feat(menu-dropdown): rh-menu-dropdown element

ArathyKumar:
## What I did

Developed the `rh-menu-dropdown` component to provide an accessible, customizable dropdown menu, enhancing navigation and user interaction.

screenshot

## Testing Instructions
Please help review the following:

[Menu Dropdown Deploy Preview](https://deploy-preview-2626--red-hat-design-system.netlify.app/elements/menu-dropdown/)

**UI Validation (Figma)**
 Compare the component with the[ Figma design](https://www.figma.com/design/gKhpTnRlHByqa92H6AcGjP/rh-dropdown?node-id=2006-2254&m=dev)
- [ ] Check spacing, padding, font sizes, colors, borders, and icon alignment
- [ ] Validate all visual states / variants:
  - [ ] Default
  - [ ] Compact
  - [ ] Boxed
  - [ ] With icons
  - [ ] Disabled items
  - [ ] With Description
  - [ ] Hover, focus, and active states
  
**Interaction & Behavior**
- [x] Open and close the dropdown using mouse clicks
- [x] Test keyboard interactions:
  - [x] Enter / Space/ Arrow Down to open
  - [x] Escape to close
  - [x] Arrow keys to navigate items
- [x] Dropdown should close when clicking outside

**Accessibility**
- [x] Confirm keyboard navigation works correctly
- [x] Check focus states on all menu items
- [x]  Validate ARIA roles and attributes (aria-expanded, aria-haspopup, role="menu", etc.)
- [x] Screen reader should correctly announce the dropdown and menu items

## Notes to Reviewers

**In Progress:**

- [x] Implementing dropdown variant with group headings
- [x] Resolving accessibility issues related to aria-posinset and aria-setsize
- [x] Writing test cases
- [x] Updating documentation
- [x] Cleaning up CSS (adding appropriate rh-tokens)
- [x] Use FloatingUI API

PR #2650, opened 2025-10-08
feat(menu-dropdown): draft PR for POC

ArathyKumar:
Draft PR for preview


markcaron:
@ArathyKumar I tend to like the composability of the structure in this draft PR.

```html
<rh-menu-dropdown>
    <rh-menu>
        <rh-menu-item></rh-menu-item>
    </rh-menu>
</rh-menu-dropdown>
```

It seems to function the same and sound the same in VO as [the other PR](https://github.com/RedHat-UX/red-hat-design-system/pull/2626).

I did notice, however, that in both PRs the first item doesn't get announced when initially opened and auto-focused. Maybe @adamjohnson could ensure that's correct or advise on how to fix.

ArathyKumar:
> @ArathyKumar I tend to like the composability of the structure in this draft PR.
> 
> ```
> <rh-menu-dropdown>
>     <rh-menu>
>         <rh-menu-item></rh-menu-item>
>     </rh-menu>
> </rh-menu-dropdown>
> ```
> 
> It seems to function the same and sound the same in VO as [the other PR](https://github.com/RedHat-UX/red-hat-design-system/pull/2626).
> 
> I did notice, however, that in both PRs the first item doesn't get announced when initially opened and auto-focused. Maybe @adamjohnson could ensure that's correct or advise on how to fix.

Hey @markcaron 

I noticed this issue earlier and had tried to address it as well. However, when the dropdown is opened, VoiceOver announces that the dropdown has expanded, which overrides the reading of the first menu item.

I also checked the example on the W3C website:
https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-actions/

It shows the same behavior there as well.

adamjohnson:
I like where this is going — reusing `<rh-menu>` inside `<rh-menu-dropdown>` makes a lot of sense at first glance, especially since it already handles list rendering, focus management, and item states.

## Planning

That said, I want to double-check what interaction model we’re targeting here. Right now, `<rh-menu>` is built around the [ARIA menu / menuitem pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/) — basically the same thing you’d see in a desktop app’s “File → Edit → View” menu bar. That pattern expects `role="menu"` and `role="menuitem"`, plus arrow key navigation, Escape to close, and so on. It’s great for app-like command menus, but not necessarily for dropdowns or filter menus in a website UI.

If `<rh-menu-dropdown>` is meant to behave more like a menu button, disclosure, or even a select/listbox pattern (per the APG), then we might run into accessibility mismatches — for example, screen readers will announce it as an application menu, and users might expect desktop-style keyboard behavior.

## What is it?

Curious what your intent is for `<rh-menu-dropdown>` from an interaction perspective (ala [previous comment](https://github.com/RedHat-UX/red-hat-design-system/pull/2626#issuecomment-3325459576)). Should it behave like a:

* Website dropdown
  * Click/space toggles visibility
  * Up/down arrow moves focus within items
  * ESC closes
* A select / listbox
* Or like a full menu bar menu?

If we do want more than one option, maybe we could add an attribute — something like `behavior="menu"` (naming is hard, please help with this) — so `<rh-menu>` (and its items) can adapt roles and keyboard handling accordingly. That way we preserve reuse while keeping semantics correct for each context. If we go that route, I would suggest using `<rh-menu>` be the default (sans-attribute) and, if people want `<rh-menu>` to act as an application menu (`role="menu"`), end users must add `behavior="menu"` or whatever it ends up being.

That said, I am **very** leery of premature abstractions. More often than not, they just add complexity. See the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

## Alternatively

Alternatively, we could draw a line in the sand.

Keep `<rh-menu>` and `<rh-menu-item>` together, relating _specfiically_ to application menus. Then, have `<rh-menu-dropdown>` (or, perhaps, `<rh-dropdown>`/`<rh-select>`) + their respective child elements relating specifically to `<rh-menu-dropdown>`. 

Happy to help sort out which ARIA pattern fits best here — I just want to make sure we’re not unintentionally signaling “application menu” semantics where a simple dropdown, select, or disclosure would be more appropriate.

hellogreg:
Just wanted to give a quick +1 to everything Adam says above.

PR #2664, opened 2025-10-14
docs: patch lit ssr client to correctly remove attributes

zeroedin:
## What I did

1. Patching the Lit SSR client to remove attributes on hydration correctly.

See https://github.com/lit/lit/pull/5115 for more info on the fix.  This upstream PR still needs tests added before it is accepted.

## Testing Instructions

1. Open the [Deploy Preview](https://deploy-preview-2664--red-hat-design-system.netlify.app/)
2. Inspect any element that has been SSR'd and hydrated. It should remove any `hydrate-internals-*` attributes and the corresponding `aria-` attributes on hydration.
3. Elements should remain accessible before and after hydration. 

## Notes to Reviewers

There may be elements where extra code was added to address issues that might have arisen due to this problem.  

https://github.com/RedHat-UX/red-hat-design-system/blob/75d19c6eee4dbbacef5cc8f811c1a5648fb69166/elements/rh-navigation-primary/rh-navigation-primary.ts#L142

https://github.com/RedHat-UX/red-hat-design-system/blob/83609884a79b70aee2d2e48040ed0e520e6b74a2/elements/rh-navigation-link/rh-navigation-link.ts#L48-L54

The comment in `navigation-link` is now an incorrect assessment, as it wasn't that the axe doesn't handle the internals correctly; instead, this bug in the lit SSR client is likely confusing axe.


hellogreg (approved):
let's get that mutation

bennypowers:
let's give it a few days to see if lit can release the fix

Issue #2623, opened 2025-09-22
![bug] classification of `-lightdom` and `-lightdom-shim` CSS file consistency


zeroedin:
We've made the distinction of `-lightdom.css` files being for deeply nested slotted content and `-lightdom-shim` files being more optional and for `:not(:defined)` files and other CLS helpers.  However we have not combed back over the repository and applied this consistency.

Known components that have an intermixing of `:not(:defined)` and deeply nested content in a single `-lightdom.css` file:

- rh-footer
- rh-subnav
- ... [ add others here as found ]

### Note:

Could have potential breaking changes.  Maybe we need a deprecation of these files to split?  Requirements for some implementation maybe to include the `-lightdom-shim` when they previously loded just `-lightdom` in such instances as `rh-subnav` where the `-lightdom` file includes only `:not(:defined)` styles (as of release 3.1.1)

PR #2665, opened 2025-10-15
fix!: lightdom shim styles should be in their own files, and not required. 

zeroedin:
## What I did

1. Places `:not(:defined)` styles in the correct stylesheet `-lightdom-shim.css`.  Reducing overhead of `-lightdom.css` for early developed elements.
2. ~~Reverted subnav lightdom change as they are happening in 2683~~ Will need to reimplement this after #2683 merges

### TODO
- [ ] Add changeset
- [ ] Reimplement the lightdom split for subnav
- [ ] Take a look at subnav with rh-navigation-link in docs.

https://github.com/RedHat-UX/red-hat-design-system/issues/2623 

## Testing Instructions

1.

## Notes to Reviewers

My opinion:  If we don't split these files, what is the purpose of any of the `-shim` files to begin with?  If we have shim files at all, they should be consistent across the elements in terms of what they and `-lightdom.css` provide, even if that means a breaking change to align them.


zeroedin:
Closing this, would still like to tackle this in a future PR.

PR #2666, opened 2025-10-16
perf(code-block): set content-visibility on code-blocks

bennypowers:
## What I did

1. set `content-visibility: auto` on codeblocks which don't intersect the viewport
2. set `content-visibility: visible` on codeblocks which do intersect the viewport
3. added a demo containing 1000 code blocks (appears to load just fine in chromium and firefox) 

## Testing Instructions

1. check out the "Thousands" code block demo

## Notes to Reviewers

Related to docs-dot performance issues


zeroedin (approved):
Lazily Getting Text Makeover

PR #2670, opened 2025-10-16
fix(code block): expandable hide/show

zeroedin:
## What I did

1. Removed the early return if hiding line numbers, still needed to expand (hide or show the element) even if the `lineNumbers="hidden"` is used

## Testing Instructions

1.

## Notes to Reviewers

It's worth mentioning that we could likely optimize the "#computeLineNumbers()" function to compute the total lines in the slotted content without the additional work of showing line numbers when they are turned off. This can be achieved by splitting that method into two separate functions.  I have not done this yet in this PR, but I want to.

Also of note is that @kylebuch8 and I are going to generate a patch file and test it instead of quickly releasing it as a patch to RHDS. This PR, although cut from the main branch, can be moved to staging/doduo later, when we work on more optimizations.

Another possible optimization is to calculate individual line heights only if the code is wrapping and lineNumbers != hidden. We can skip that calculation if wrap is not turned on.

### DIFF
--- a/.changeset/solid-teeth-allow.md
+++ b/.changeset/solid-teeth-allow.md
@@ -2,5 +2,5 @@
 "@rhds/elements": patch
 ---
 
-`<rh-code-block>`: fixed ability to expand hide/show when hiding line numbers'
+`<rh-code-block>`: show expand button ("Show more") even when line numbers are hidden
   


.changeset/solid-teeth-allow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-code-block>`: fixed ability to expand hide/show when hiding line numbers'
```
bennypowers:
```suggestion
`<rh-code-block>`: show expand button ("Show more") even when line numbers are hidden
```

PR #2673, opened 2025-10-17
fix(progress-stepper): token css fallback missing 

zeroedin:
## What I did

1. Corrects missing fallback token values if tokens aren't loaded at the document level.

Thanks @zhawkins for reporting.

## Testing Instructions

1.


zhawkins:
Thank you!

bennypowers (approved):
solid fix. thanks for the report @zhawkins

PR #2674, opened 2025-10-17
docs: add copy permalink on slots, attrs, methods, events, parts

### LINKED ISSUE #1604: docs: linki

zeroedin:
## What I did

1. Added a copy permalink for direct linking to slots, attrs, methods, events, and parts on element code pages.
2. Did not add a permalink to CSS custom properties, as it resulted in a very long URL pattern.  We can reassess that if necessary.

Closes #1604 

## Testing Instructions

1. Visit [Navigation Primary Dropdowns Slot](https://deploy-preview-2674--red-hat-design-system.netlify.app/elements/navigation-primary/code/#rh-navigation-primary-slots-dropdowns)

## Notes to Reviewers
Note, this does not create a pattern where we auto-open a parent accordion when following a hash link, as the ux.redhat.com code page accordions are open by default to accommodate cmd+F.  If we want that, IMHO, it is a new pattern and it is a new issue that will need to be created.

### DIFF
--- a/docs/styles/styles.css
+++ b/docs/styles/styles.css
@@ -365,13 +365,8 @@ rh-alert code {
 
 @layer overrides {
   /**
-     * Overrides layer for overriding component and site styles
-     */
-
-  /* TODO: Document this bug table overflows vertically causing a scrollbar when it shouldn't */
-  rh-table {
-    overflow-y: unset;
-  }
+   * Overrides layer for overriding component and site styles
+   */
 
   /* Compact tile + Uxdot Example reduce overall spacing */
   rh-tile[compact] uxdot-example::part(container) {
@@ -382,5 +377,21 @@ rh-alert code {
     /* TODO: Document this bug shouldn't have to force the font size
                the component is trying to set internally on #description */
     font-size: var(--rh-font-size-body-text-md) !important;
+  }
+
+  rh-table {
+    /* TODO: Document this bug table overflows vertically causing a scrollbar when it shouldn't */
+    overflow-y: unset;
+
+    /* Hover state for copy permalink button when in a table row */
+    & tr {
+      &:hover {
+        --uxdot-copy-permalink-button-background:
+          light-dark(
+              var(--rh-color-surface-light),
+              var(--rh-color-surface-darker)
+            );
+      }
+    }
   }
 }
--- a/uxdot/uxdot-copy-permalink.css
+++ b/uxdot/uxdot-copy-permalink.css
@@ -18,7 +18,13 @@
 }
 
 #button:is(:hover, :focus) {
-  background: light-dark(var(--rh-color-surface-light), var(--rh-color-surface-darker));
+  background:
+    var(--uxdot-copy-permalink-button-background,
+      light-dark(
+        var(--rh-color-surface-light),
+        var(--rh-color-surface-dark)
+      )
+    );
   color: var(--rh-color-text-primary);
 }
 


adamjohnson (changes requested):
Screenshot of the default slot with an arrow pointing to the copy permalink icon

Default slots do not have a slot name and cause accessibility issues to be flagged in axe: "Links must have discernable text".

I have actually thought this was confusing for some time. IMO we should add some sort of text for the default slot. `{Default slot}` or something. That would solve this a11y issue too. 

Yes, I could see how this could be misconstrued in the other direction. But, hopefully, we can come up with something that is clear for our RH developers.

docs/_includes/layouts/pages/element.11ty.ts
```diff
@@ -419,7 +419,7 @@ export default class ElementsPage extends Renderer<Context> {
                   </uxdot-copy-permalink></td>
                   <td>${await this.#innerMD(slot.summary)}</td>
                   <td>
-                   ${slot.name === '' ? await this.#innerMD(`<strong>Note:</strong> [default] <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot#attributes">unnamed slots</a> do not have a slot="name" attribute. ${slot.description}`) : await this.#innerMD(slot.description)}
+                   ${slot.name === '' ? await this.#innerMD(`<strong>Note:</strong> [default] <a href="https://github.com/RedHat-UX/red-hat-design-system/wiki/Shadow-DOM-Style-Guide#slot-names">unnamed slots</a> do not have a slot="name" attribute. ${slot.description}`) : await this.#innerMD(slot.description)}
```
bennypowers: why this change?

zeroedin: @adamjohnson suggested maybe linking to somewhere else other then mdn.  I agreed that it didn't do a great job of explaining the lack of a slot/name attribute.  In fact, if you read the previous link, it talks about `name=""` rather than just `<slot>` sans the name attribute, so it was a bit confusing.  We decided to describe that in our own wiki.

zeroedin: Let me know if you have a better idea/preference?

adamjohnson: I don't kow if the Wiki is a good place to link to. Generally, we use the Wiki for people who are creating RHDS components—it's our style guide.

The link above also lists:

```html
<slot></slot>
```

vs what page authors will use:

```html
<rh-cta>Content</rh-cta>
```

Perhaps it would be better to link to a new section we create under Get Started > Developers > [Usage](https://deploy-preview-2674--red-hat-design-system.netlify.app/get-started/developers/usage/#other-resources)? Or to this bit of [Lit documentation](https://lit.dev/docs/components/shadow-dom/#using-the-slot-element)?

docs/_includes/layouts/pages/element.11ty.ts
```diff
@@ -410,9 +410,17 @@ export default class ElementsPage extends Renderer<Context> {
...
+                    <span id="${tagName}-slots-${this.slugify(slot.name)}">
+                      <a href="#${tagName}-slots-${this.slugify(slot.name)}">
+                        <code>${slot.name === '' ? '[default]' : slot.name}</code>
+                      </a>
+                    </span>
+                  </uxdot-copy-permalink></td>
                   <td>${await this.#innerMD(slot.summary)}</td>
-                  <td>${await this.#innerMD(slot.description)}</td>
+                  <td>
+                   ${slot.name === '' ? await this.#innerMD(`<strong>Note:</strong> [default] <a href="https://github.com/RedHat-UX/red-hat-design-system/wiki/Shadow-DOM-Style-Guide#slot-names">unnamed slots</a> do not have a slot="name" attribute. ${slot.description}`) : await this.#innerMD(slot.description)}
```
adamjohnson: Screenshot of the default slot content for rh-navigation-primary with some design changes described below

Can we change the design so it looks more like this?

1. Move the description to be before the "Note".
2. Add a `<br>` tag after the description.
3. Wrap the "Note" text in a span and add a class to make the text smaller (or inline style).

zeroedin: Reformatting done.

adamjohnson: Resolved in [77c1b51](https://github.com/RedHat-UX/red-hat-design-system/pull/2674/commits/77c1b516c981548a0e562cdf00c7fb2fb6fbc464). 🎉

docs/_includes/layouts/pages/element.11ty.ts
zeroedin: I had also suggested the lit documentation URL as a possible place to link to. I agree that the endpoint would be best if we controlled it in some way, which is why I suggested the wiki. It is our implementation detail.

bennypowers: we're talking about anonymous slots. this is a platform thing, independent of framework (lit) or our library. we should link to mdn or to the html spec.

zeroedin: The language MDN uses talks about `name=""`, which, in my opinion, is not intuitive to how we implement it—sans attribute at all. `<slot>` vs `<slot name="">`.  This is what I was refering to as an implementation detail, not to get it crossed with the platform idea, because yes it is true of the platform as well, just not elegantly in spec.

>Unnamed slots have the name default to the empty string. 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot#attributes

Would be better if mdn, then followed that up with:  "OR the slot attribute not present / undefined"

The HTML spec doesn't really talk about anonymous slots at all in the same context we are trying to convey either: 

>The name content attribute may contain any string value. It represents a [slot](https://dom.spec.whatwg.org/#concept-slot)'s [name](https://dom.spec.whatwg.org/#slot-name).
https://html.spec.whatwg.org/multipage/scripting.html#the-slot-element

Also, not saying anything about the absence, stretching the word "may" a bit, I believe.

We are trying to get the user to understand that we can't represent an empty string in our table, so we have replaced it with `[default]`. Maybe `[default]` needs to be `[anonymous]`.  AND what this means for them is if they don't include `slot="name"` on an element, it will end up in the default slot.  

If you feel strongly that one of these links does a better job of that than the other, I'm good with whatever you all decide.

bennypowers: make a pr to mdn? but for "how slots work" we shouldn't be linking to our own docs, i don't think. comprehensive how-to-html docs are out of scope for now

bennypowers:
maybe out of scope but the copy permalinks aren't legible when hovered in dark scheme

bennypowers:
also, for events, the permalink copy button appears above the title, not at inline-end

zeroedin:
> maybe out of scope but the copy permalinks aren't legible when hovered in dark scheme

Added a quick `light-dark` var to uxdot-copy-permalink that allows us to override this given table's hover effect, which is the same color as what the copy buttons' hover is.

> also, for events, the permalink copy button appears above the title, not at inline-end

There was an extra `</uxdot-copy-permalink>` close tag where it shouldn't have been... thanks, autocomplete.

bennypowers (approved):
lads, 'gone to market

PR #2675, opened 2025-10-17
fix(navigation-primary): visibly hide details contents for cmd/ctrl + f

zeroedin:
## What I did

1. Visually hide content in hamburger and nav item dropdowns when `cmd/ctrl + f` is used on the page unless that menu is open.

## Testing Instructions

1.  Visit a [DP demo](https://deploy-preview-2675--red-hat-design-system.netlify.app/elements/navigation-primary/demo/) press cmd+F or ctrl+F and search for a term that would appear inside a dropdown.  Dropdown menus should not open.  
2. If a menu is open cmd+F should find the term.


zhawkins:
Awesome. Thanks for doing this Steven!

hellogreg (approved):
locate guiding text manually

PR #2677, opened 2025-10-19
feat(code-block): mutable `copy` event

### LINKED ISSUE #2672: [feat]: `<rh-code-block>`: ability to add 'line-start' prompts hidden from copy function
Ability to add a line start character `$` for bash scripts that are hidden from the copy function without the need to duplicate the `<pre>`.  Doing so causes other issues, such as possible line number miscalculation by injecting a duplicate `<pre>` tag.

bennypowers:
Closes #2672

## What I did

1. Added a new `copy` event to rh-code-block
2. Added two demos: 
    1. [modifying](https://deploy-preview-2677--red-hat-design-system.netlify.app/elements/code-block/demo/copy-modify/) the copy text (i.e. removing prompt)
    2. [cancelling](https://deploy-preview-2677--red-hat-design-system.netlify.app/elements/code-block/demo/copy-cancel/) the copy event (prevent copy altogether)

## Testing Instructions

1. check out the demos

## Notes to Reviewers

~~I added a simple shake-back-and-forth animation to the tooltip when the copy action is cancelled, although it's not clear to me that users will understand that the animation means that copying failed.~~ we could make it explicit with a new attr/slot combo for the cancel text, perhaps.

### DIFF
--- a/docs/_includes/layouts/pages/element.11ty.ts
+++ b/docs/_includes/layouts/pages/element.11ty.ts
@@ -410,9 +410,17 @@ export default class ElementsPage extends Renderer<Context> {
               <tbody>
                 ${(await Promise.all(slots.map(async slot => html`
                 <tr>
-                  <td><code>${slot.name}</code></td>
+                  <td><uxdot-copy-permalink>
+                    <span id="${tagName}-slots-${this.slugify(slot.name)}">
+                      <a href="#${tagName}-slots-${this.slugify(slot.name)}">
+                        <code>${slot.name === '' ? '[default]' : slot.name}</code>
+                      </a>
+                    </span>
+                  </uxdot-copy-permalink></td>
                   <td>${await this.#innerMD(slot.summary)}</td>
-                  <td>${await this.#innerMD(slot.description)}</td>
+                  <td>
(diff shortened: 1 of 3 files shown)


elements/rh-code-block/rh-code-block.ts
```diff
@@ -455,28 +467,37 @@ export class RhCodeBlock extends LitElement {
...
+        }
       }
+      this.requestUpdate();
+      tooltip?.show();
+    } else if (tooltip) {
+      this.classList.add('copy-error');
+      this.addEventListener('animationend', () => {
+        this.classList.remove('copy-error');
+      }, { once: true });
     }
```
zeroedin: I think we would need to announce the error to screen readers here as well.

bennypowers: so then we need the slot/attr pair afterall, and should maybe lose the animation (but we could keep the colorization until tooltip closes, if @marionnegp and/or @coreyvickery  agrees)

bennypowers:
@kylebuch8 validated the approach here, so now it's just code review

elements/rh-code-block/rh-code-block.ts
bennypowers: added in 86eb52b

zeroedin (approved):
Logic Guards Twin Messages

adamjohnson (approved):

Issue #2638, opened 2025-10-02
[docs][bug] Side Nav disclosure close issue


markcaron:
### Description

When first landing on a subitem/page within a main category, that main category disclosure cannot be closed on the initial click. It has to be clicked twice.

### Screenshots

<img width="375" height="192" alt="Image" src="https://github.com/user-attachments/assets/fae7484a-82b1-44c9-8b74-83f9376f8546" />

### Steps to reproduce

1. Go to [ux.redhat.com](https://ux.redhat.com/)
2. Click to open `Tokens` category
3. Click on `Overview`
4. Once the tokens overview page loads, click on `Tokens` again to try to close it **
5. Notice it doesn't close?
6. Click a 2nd time, and watch it close

** Whether you click on the word `Tokens` or the arrow, it doesn't work.

### Expected behaviour

The disclosure should toggle properly with each click (open/close)

### Operating System (OS)

macOS

### Browser

Chrome

PR #2678, opened 2025-10-20
docs: correct uxdot-sidenav details toggle state

Gili-Greenberger:
## What I did

1. Fixed issue where the side navigation dropdown required two clicks to collapse.

2. Synced the expanded state with the details open attribute and active child items.

3. Updated toggle logic so dropdown closes on the first click.

## Testing Instructions

1. Open the side navigation and expand a dropdown.

2. Click again on the same dropdown - it should collapse immediately on the first click (not require two).

3. Navigate to a page that matches a child link - the dropdown should initialize expanded automatically.

## Notes to Reviewers

This resolves issue #2638


.changeset/wicked-times-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+Fix navigation dropdown that required two clicks to close. Now it collapses with a single click.
```
bennypowers: Since this PR only affects docs (we don't publish the uxdot-* elements), it doesn't need a changeset. Please delete this file.

uxdot/uxdot-sidenav.ts
```diff
@@ -189,9 +201,9 @@ export class UxdotSideNavDropdown extends LitElement {
   async #onClick(event: Event) {
     if (!event.composedPath().some(node => node instanceof HTMLAnchorElement)) {
       event.preventDefault();
+      this.#details = this.#details || this.querySelector('details');
```
bennypowers: make `#details` a getter then, so that we only have this logic in one place

.changeset/wicked-times-rest.md
zeroedin: On the same note, maybe change the title of the PR to be:

"docs: correct uxdot-sidenav details toggle state"

uxdot/uxdot-sidenav.ts
Gili-Greenberger: Updated  #details is now implemented as a getter so the logic only lives in one place

uxdot/uxdot-sidenav.ts
```diff
@@ -186,12 +196,15 @@ export class UxdotSideNavDropdown extends LitElement {
     `;
   }
 
+  get #details(): HTMLDetailsElement | null {
+    return this.querySelector('details');
+  }
```
zeroedin: Move this above the lifecycle methods

https://github.com/RedHat-UX/red-hat-design-system/wiki/Custom-Elements-API-Style-Guide#ordering

bennypowers (approved):
LGTM!

PR #2683, opened 2025-10-21
fix(subnav):  improve accessibility with nested list when `<rh-navigation-link>` is slotted

zeroedin:
## What I did

1. Updated subnav to handle slotted `<rh-navigation-link>` automatically adds role="list" to container if a `<rh-navigation-link>` is slotted.
2. deprecated slotted `<a href>` implementation.
3. Corrected and deprecated `color-palette` attribute. Design has stated subnav should be transparent and react to the parent color-scheme.
4. ~~Correctly split `lightdom.css` and `lightdom-shim.css` styles into their correct files.~~ Maintained the lightdom as a single file, we will need to make this as a breaking change later.  https://github.com/RedHat-UX/red-hat-design-system/pull/2665
5. Implemented real decorated properties for the scroll labels so they are correctly documented.
6. Corrected scroll button bottom borders
7. Updated ux. docs integration with subnav: created a theme override to provide subnav with the correct background color.

before
```html
<rh-subnav>
  <a href="#">Link</a>
  ...
</rh-subnav>
```
after with rh-navigation-link
```html
<rh-subnav>
  <rh-navigation-link href="#">Link</rh-navigation-link>
  ...
</rh-subnav>
```
### Todo

- [x] docs
- [x] add changeset  

## Testing Instructions

1. View [Deploy Preview](https://deploy-preview-2683--red-hat-design-system.netlify.app/)

### DIFF
--- a/.changeset/twenty-shirts-carry.md
+++ b/.changeset/twenty-shirts-carry.md
@@ -2,5 +2,4 @@
 "@rhds/elements": patch
 ---
 
-`<rh-subnav>`: updated `active` to `aria-current="page"` on slotted links so the current page is
- announced to assistive technology
+`<rh-subnav>`: corrected styling to reflect proper `aria-current="page"` usage for the active page


elements/rh-subnav/rh-subnav.ts
```diff
@@ -55,14 +55,14 @@ export class RhSubnav extends LitElement {
   /**
    * Label for the scroll back button
    */
-  @property({ reflect: true, attribute: 'label-scroll-back' })
-  labelScrollBack = 'Scroll back';
+  @property({ reflect: true, attribute: 'label-scroll-left' })
+  labelScrollLeft = 'Scroll left';
```
bennypowers: why this change? rtl?

zeroedin: The change reverted it to the previously "non-documented" way. Limiting the real changes if someone was already using it in that way. Personally, I would prefer it as `back` and `forward`, but reverted for less surface area of change. Instead, the change is that the property becomes a real reactive property and is correctly documented.

https://github.com/RedHat-UX/red-hat-design-system/blob/59dde97864cdc3c523764cd9f975dedca8edce36/elements/rh-subnav/rh-subnav.ts#L109

The default value should have remained `Scroll back` now that I look at the entire previous state that was a mistake, the change should have only been to the attribute name itself.

elements/rh-subnav/rh-subnav.ts
```diff
@@ -97,28 +108,30 @@ export class RhSubnav extends LitElement {
...
-      <nav part="container" aria-label="${this.accessibleLabel}">
+      <nav part="container" 
+           aria-label="${this.accessibleLabel}">
         ${!this.#overflow.showScrollButtons ? '' : html`
           <button id="previous"
                   tabindex="-1"
                   data-direction="start"
-                  aria-label="${this.getAttribute('label-scroll-left') ?? 'Scroll back'}"
+                  aria-label="${this.labelScrollLeft}"
                   ?disabled="${!this.#overflow.overflowLeft}"
```
adamjohnson, line 121:
```suggestion
                  ?aria-disabled="${!this.#overflow.overflowLeft}"
```

Instead of `disabled`, could we use `aria-disabled` here so that this element remains focusable, but still disabled?

Same goes for the `disabled` attribute on line ~135.

elements/rh-subnav/rh-subnav.css
```diff
@@ -15,6 +19,52 @@
...
+::slotted(a):after {
+  content: '' !important;
+  position: absolute !important;
+  inset: 0 !important;
+  width: 100% !important;
+  border-block-end: var(--rh-border-width-lg, 3px) solid transparent !important;
+}
+
+::slotted(a:hover),
+::slotted(a:active),
```
adamjohnson, line 40: When visiting the [basic demo](https://deploy-preview-2683--red-hat-design-system.netlify.app/elements/subnavigation/demo/) on the DP, we have `<a href="#" active="">Servers</a>`. 

When using AT, `active` has no semantic meaning. VoiceOver does not announce "Current page" when tabbing over this link. Would it be worthwhile to detect instances of `active` in lightdom `<a>` tags and automatically add `aria-current="page"`?

elements/rh-subnav/rh-subnav.ts
```diff
@@ -97,28 +108,30 @@ export class RhSubnav extends LitElement {
   override render() {
     return html`
       <!-- container, \`<div>\` element -->
-      <nav part="container" aria-label="${this.accessibleLabel}">
+      <nav part="container" 
+           aria-label="${this.accessibleLabel}">
         ${!this.#overflow.showScrollButtons ? '' : html`
```
adamjohnson, line 116: If I view the [basic demo](https://deploy-preview-2683--red-hat-design-system.netlify.app/elements/subnavigation/demo/) on the DP and shrink my viewport down to a size where I should see left and right scroll `<button>` elements, they do not appear.

Curiously, this also happens [on uxdot](https://ux.redhat.com/elements/subnavigation/demo/).

elements/rh-subnav/rh-subnav.ts
```diff
@@ -97,28 +108,30 @@ export class RhSubnav extends LitElement {
...
         <!--
           slot:
-            description: Navigation links, expects collection of \`<a>\` elements
+            description: Sub navigation links, expects collection of \`<a>\` or \`<rh-navigation-link>\` elements
+            Slotting a \`<a>\` element is deprecated and will be removed in a future release. Use \`<rh-navigation-link>\` instead.
           part:
             description: the anonymous slot
         -->
-        <slot part="links" @slotchange="${this.#onSlotchange}"></slot>
+        <div role="${ifDefined(this.hasNavigationLinks ? 'list' : undefined)}" part="links"><slot @slotchange="${this.#onSlotchange}"></slot></div>
```
adamjohnson: Is changing the location of `part="links"` to the containing `<div>` vs the `<slot>` element going to impact current users?

I'm thinking about styling nested elements with `part::()` (or rather the limitations surrounding this API).

.changeset/slimy-clouds-fall.md
adamjohnson, line 1: Would it be worthwhile to add a "Before vs After" example here?

See the [Changelog](https://github.com/RedHat-UX/red-hat-design-system/blob/main/CHANGELOG.md) for examples.

docs/styles/styles.css
adamjohnson, line 1: Screenshot of subnav on uxdot with js disabled.

With JS disabled / before upgrading, subnav looks like this on uxdot. It may be worthwhile to add some styles either to uxdot or to the lightdom (-shim) to avoid this FOUC.

elements/rh-subnav/rh-subnav.css
adamjohnson: This may require changes to the [Dynamic](https://deploy-preview-2683--red-hat-design-system.netlify.app/elements/subnavigation/demo/dynamic/) demo, should we explore this option.

adamjohnson (changes requested):
Many of my comments here may not be directly tied to this PR and could be broken off into separate issues if necessary.

That said, the issue regarding "Scroll forward/back" buttons not showing on the basic demo should be addressed.

elements/rh-subnav/rh-subnav.css
zeroedin: > Would it be worthwhile to detect instances of active in lightdom <a> tags and automatically add aria-current="page"?

Ooofff, yes, this was another accessibility issue fixed with the breaking change I forgot about documenting thanks for bringing it back to light.  Yes, we need to improve this experience. `active` should never have been a thing here. `aria-current="page"` should have been set directly by the user.

elements/rh-subnav/rh-subnav.ts
zeroedin: This is a problem with the demo styles and is the reason I created the [overflow-scroll demo](https://deploy-preview-2683--red-hat-design-system.netlify.app/elements/subnavigation/demo/overflow-scroll/).

elements/rh-subnav/rh-subnav.ts
zeroedin: Yes this was a mistake, good catch that should have been moved back after reverting the breaking change.

zeroedin: fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/2683/commits/73e2f3a59f3c1b59966aedd5e85fff532d160863

.changeset/slimy-clouds-fall.md
zeroedin: There really isn't a before and after here. This isn't a breaking change.   The only thing this change does is fix when you add `color-palette="darker"` that it actually works and doesn't present a broken state.  The `color-palette` attribute on ux.redhat.com does not correctly apply the background style.

Technically, by design, the `color-palette` attribute should be removed entirely, as this component should be a consumer only.  This was part of the breaking change.

docs/styles/styles.css
zeroedin: So this is because docs got updated to use `<rh-navigation-link>`.  Likely this will have to happen on docs styles as a specfic implmentation of lightdom for rh-nav link in subnav.

elements/rh-subnav/rh-subnav.ts
zeroedin: We are going to leave this alone because users can tab through links.

elements/rh-subnav/rh-subnav.ts
zeroedin: This should be fixed now.

docs/styles/styles.css
zeroedin: Going to handle this at the docs layer until we can implement https://github.com/RedHat-UX/red-hat-design-system/pull/2665

elements/rh-subnav/rh-subnav.css
adamjohnson: Do you think people will have been keying off `a[active]` in their CSS? 

This change removes `active` and adds `aria-current="page"`. Wondering if it should keep the `active` attribute too until we issue a breaking change.

zeroedin: Recap chat: `active` was never a real property, it was a mistake that should have been caught earlier in subnav's lifespan.  The fix here was the retain the styles of `[active]` because it was proliferated by the demos, but remove it from the demos moving forward, replacing with proper `[aria-current="page"]`.  

We also now have a `a:is([active], [aria-current="page"])` `:is` selector that will style both the improper usage (backward compat) and the forward progress usage, which is aria-current.  Nothing is breaking instead the suggestion is to properly fix accessibility by using aria-current="page" as a way to mark the current page with the style, but no attribute was created nor destroyed.

adamjohnson (approved):
Links Get Thoroughly Modernized

PR #2684, opened 2025-10-21
docs: theming for pf-select

### LINKED ISSUE #2681: [docs] light-dark support in edit element properties
Text gets lost in dark mode for dropdowns using `pf-select`, likely a token use issue, as the select dropdown backgrounds themselves do not chnage.

Image

bennypowers:
## What I did

1. add initial css for pf-select theming

## Testing Instructions

1. see any (e.g. icon) knobs widget

## Notes to Reviewers

pf-select's css is broken, and doesn't actually support theming

Closes #2681


uxdot/uxdot-knob-attribute.css
```diff
@@ -1,3 +1,32 @@
...
+  --pf-c-select__menu--BackgroundColor: var(--rh-color-surface);
+  --pf-c-select__menu-item--Color: var(--rh-color-text-primary);
+
+  /* doesn't work */
+  --pf-c-select__menu-item--hover--BackgroundColor: var(--rh-color-surface-lighter);
+
+  /* doesn't work */
+  --pf-c-select__menu-item--focus--BackgroundColor: var(--rh-color-surface-lighter);
+}
+
```
zeroedin, line 29: if we update to @patternfly/elements@4.2.0 we can do this nasty hack:

```css
  pf-option {
    --rh-color-gray-20:
      light-dark(
          var(--rh-color-surface-lighter),
          var(--rh-color-surface-dark)
        ) !important;
    --_pf-option-selected-background-color:
      light-dark(
          var(--rh-color-surface-lighter),
          var(--rh-color-surface-dark)
        ) !important;
    --_svg-color: var(--rh-color-interactive-primary-default);
  }
}
```

zeroedin: also there is a rouge `--rh-color-gray-20` in PFE that shouldn't be there 🙃

bennypowers: @dali327519294 can you continue this PR, instead of https://github.com/patternfly/patternfly-elements/pull/2952?

PR #2695, opened 2025-10-30
docs(card): card title typography in asset and title-bar patterns

### LINKED ISSUE #2164: [bug] title style fixes
### Describe the bug

Heading styles from a constructed stylesheet are preventing titles in some card patterns from being consistent with what's in Figma. Some of the computed values end up being correct, but since these titles use `--rh-font-family-body-text`, the following css uses `body-text` tokens for the other text properties too. (I think that's what we typically do.)

- [ ] Asset card titles should be using:
```
font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif);
font-weight: var(--rh-font-weight-body-text-regular, 400);
color: var(--rh-color-text-brand-on-light, #ee0000);
```

- [ ] Title bar cards should be using 
```
font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif);
font-weight: var(--rh-font-weight-body-text-regular, 400);
```

<hr>

Unrelated to the heading style issue in card:

- [ ] Titles in `<rh-stat>` should be using `--rh-font-family-body-text`

### Screenshots

Image

### Steps to reproduce

Check demos for asset cards, title bar cards, and stats.

### Expected behaviour

Expected behavior explain the bug description

### Element HTML

```HTML

```

### Operating System (OS)

macOS

### Browser

_No response_

_No response_

HadassahYelenik:
Fixes #2164 by updating card pattern headings to match the Figma design using body-text typography tokens.

## Changes
Updated three card patterns (title-bar, asset-title-and-link, asset-title-and-link-top) to override heading typography using CSS custom properties:

- Font family: body text instead of heading

- Font weight: regular (400) instead of light (300)

- Font size: pattern-specific (md or xl from body-text scale)

- Line height: body-text line height

- Color (asset patterns only): light-dark() with brand red in light scheme and white in dark scheme for proper WCAG contrast

This approach uses the component’s existing custom property API to override typography at the pattern level, keeping the rh-card component unchanged and preserving its !important declarations for general use.

## Reviewer Notes

- Component CSS (rh-card.css) remains unchanged—!important is preserved

- All typography changes are made via CSS custom property overrides in the pattern files

- The light-dark() color approach ensures accessibility requirements are met in both color schemes without increasing font weight

### DIFF
--- a/.changeset/wide-bugs-feel.md
+++ b/.changeset/wide-bugs-feel.md
@@ -1,6 +1,0 @@
----
-"@rhds/elements": patch
----
-
-`<rh-card>`: fix card title typography in asset and title-bar patterns to match design specs
-  


elements/rh-card/rh-card.css
bennypowers, line 1: I'm not sure this is the correct approach. did you try setting the --rh-font-family-heading from the outside instead?

HadassahYelenik: That makes sense! I updated the implementation to override the heading-related custom properties on each pattern instead of removing the !important flags. This keeps the design system structure intact while letting the patterns apply the correct body-text tokens. Much cleaner than what I had before.

docs/patterns/card/patterns/asset-title-and-link-top.html
```diff
@@ -20,9 +29,7 @@ <h4>Title, Lg</h4>
       margin-block: 0 var(--rh-space-lg, 16px);
       display: flex;
       gap: var(--rh-space-lg, 16px);
-      font-weight: var(--rh-font-weight-heading-regular, 300);
-      font-size: var(--rh-font-size-body-text-md, 1rem);
-      font-family: var(--rh-font-family-body-text);
+      color: var(--rh-color-text-brand-on-light, #ee0000);
```
zeroedin: @marionnegp can you confirm that we'd need to use the theming token `light-dark()` variation here: `--rh-color-text-brand`.

```css
var(--rh-color-brand-red, 
  light-dark(
    var(--rh-color-brand-red-on-light, #ee0000), 
    var(--rh-color-brand-red-on-dark, #ee0000)
  )
);
```

docs/patterns/card/patterns/asset-title-and-link.html
```diff
@@ -36,9 +45,7 @@ <h4>
       margin-block: 0 var(--rh-space-lg, 16px);
       display: flex;
       gap: var(--rh-space-lg, 16px);
-      font-weight: var(--rh-font-weight-heading-regular, 300);
-      font-size: var(--rh-font-size-body-text-md, 1rem);
-      font-family: var(--rh-font-family-body-text);
+      color: var(--rh-color-text-brand-on-light, #ee0000);
```
zeroedin: `--rh-color-text-brand` likely here as well @marionnegp

elements/rh-card/rh-card.css
bennypowers: Lovely. you can change the commit topic from fix to docs then, since this won't affect the npm package

docs/patterns/card/patterns/asset-title-and-link-top.html
marionnegp: See [my reply](https://github.com/RedHat-UX/red-hat-design-system/pull/2695#discussion_r2490869864) to the other comment.

docs/patterns/card/patterns/asset-title-and-link.html
marionnegp: Yep! I think this code predates the switch to `light-dark`. However, we need to adjust the font weight for the brand red text to have enough color contrast at this font size. I'll make a separate comment for that.

marionnegp: Actually, is it possible to use `--rh-color-text-brand` for the title text in light scheme and `--rh-color-text-primary` in dark scheme? 

A medium font weight still isn't bold enough to satisfy the color contrast requirements, and I don't think we want these card headings to be bold or much bigger. I think switching the color to white in dark scheme is our best bet.

HadassahYelenik: Thank you @zeroedin and @marionnegp!
Updated to use `light-dark()` with brand red on light and primary white on dark for better contrast.

.changeset/wide-bugs-feel.md
bennypowers, line 1: shouldn't need a changeset

PR #2696, opened 2025-10-30
docs(navigation-link): add coming soon to docs

zeroedin:
## What I did

1.  Adds coming soon alert to docs

## Testing Instructions

1. View deploy preview, overview, styles, and guidelines for rh-navigation-link

### DIFF
--- a/elements/rh-navigation-link/docs/10-style.md
+++ b/elements/rh-navigation-link/docs/10-style.md
@@ -6,7 +6,3 @@
     message: 'This element is available to use, but the rest of the documentation is in progress.',
   });
 </script>
-
-## When to use
-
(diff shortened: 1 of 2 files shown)


markcaron:
@zeroedin Alerts look good, but maybe in another PR, we need to tweak the Overview description some, since there's a typo and could use a few tweaks.

**Current** 

```
Navigation Link is a link that is used as a child of the primary, secondary, subnav, and vertical navigation elements. The link is intrinsically a list item and should not be used outside of navigation elements that define the parent list element.

```

**Suggested**

```
Navigation Link is a link element that is used as a child of the primary, secondary, subnav, and vertical navigation elements. Intrinsically, the Navigation Link is a list item and should not be used outside of navigation elements that define the parent list element.

```

PR #2700, opened 2025-10-31
chore: update @rhds/tokens dep

zeroedin:
## What I did

1. updated @rhds/tokens dep

## Testing Instructions

1. View Deploy preview, check `--rh-color-surface` output value


adamjohnson (approved):
Light Gets Tuned Masterfully

Issue #2685, opened 2025-10-21
[bug] lightdom.css linking in tooling from dev server to docs


zeroedin:
Correct tooling/redirects to ensure the ability to link lightdom.css files in the dev server and docs (ux.) across element demos/directories.

Would prefer not to hard link as was done here: https://github.com/RedHat-UX/red-hat-design-system/pull/2624/commits/27c20cc1ce1360fecd2755e0f4f92a7cb4b42cc7

zeroedin:
The challenge with this fix is removing the hard links in the demo HTML

ie: `rh-navigation-link/demo/in-nav-primary.html`
```html
<link rel="stylesheet" href="/assets/packages/@rhds/elements/elements/rh-navigation-primary/rh-navigation-primary-lightdom.css">
```

Replacing that with a relative link, such as:
```html
<link rel="stylesheet" href="/rh-navigation-primary/rh-navigation-primary-lightdom.css">
```

Such a path should work in both the dev-server and the docs.

This is likely a change to the element-demos.ts file in the demo-lightdom-css transform.

```js
  /**
   * Eleventy transform to rewrite demo lightdom CSS URLs in demo.html pages.
   *
   * This transform looks for all href attributes in the form of relative lightdom CSS
   * imports (i.e. href="./rh-button/rh-button-lightdom.css" or similar) within
   * the page located at "./docs/elements/demo.html". It rewrites those hrefs to use
   * the correct static asset path:
   *
   *   /assets/packages/@rhds/elements/elements/${tagName}/${filename}
   *
   * This ensures that component lightdom stylesheets are loaded correctly from the
   * built assets directory, rather than failing due to invalid relative paths in
   * generated output.
   *
   * Only runs for the "demo.html" Eleventy input file.
   */
  eleventyConfig.addTransform('demo-lightdom-css', async function(this, content) {
    const { inputPath } = this.page;

    if (inputPath === './docs/elements/demo.html') {
      // Replace all lightdom CSS hrefs with the correct paths
      content = content.replace(LIGHTDOM_HREF_RE, (match, pathname) => {
        // Extract tag name and filename from the path
        // e.g., "./rh-button/rh-button-lightdom.css" or "../rh-button/rh-button-lightdom-shim.css"
        const parts = pathname.split('/');
        const filename = parts.pop(); // e.g., "rh-button-lightdom.css"
        const tagName = parts.pop(); // e.g., "rh-button"

        // Build the new path
[...]

PR #2710, opened 2025-11-04
fix(tooling): lightdom CSS linking in demo pages

Gili-Greenberger:
Fixed bug https://github.com/RedHat-UX/red-hat-design-system/issues/2685
## What I did

1. Updated the `demo-lightdom-css` transform in `docs/_plugins/element-demos.ts` to handle both relative paths (e.g., `../rh-button-lightdom.css`) and absolute paths (e.g., `/rh-button/rh-button-lightdom.css`)
2. Modified the regex pattern `LIGHTDOM_HREF_RE` to match both path formats
3. Enhanced the path extraction logic to correctly identify element names and filenames from both path styles
4. Updated the web-dev-server middleware comment for clarity

## Testing Instructions

1. Verify that demo pages load and display correctly.
2. Confirm that Light DOM styles are applied as expected.
3. Ensure that both development and production builds behave consistently.

## Notes to Reviewers

This fix allows demo HTML files to use simpler, more maintainable paths instead of hard-coded `/assets/packages/...` URLs

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,8 @@
     "Corey Vickery <cvickery@redhat.com>",
     "Diwanshi Gadgil <dpandey@redhat.com>",
     "Em Nouveau <enouveau@redhat.com>",
+    "Gili Greenberger (https://github.com/Gili-Greenberger)",
+    "Hadassah Yelenik",
     "Ivana Rodriguez <ivana@redhat.com>",
     "Kelsey Swanson <kswanson@redhat.com>",
     "Luke Dary <ldary@redhat.com>",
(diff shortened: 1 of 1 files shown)


bennypowers:
the source files should reference their subresources relative to their location

so for example, let's imagine that we wanted to link in the lightdom css from navigation-primary into a demo for rh-navigation-link, at `elements/rh-navigation-link/demo/has-subresource.html`. the demo file should contain this snippet:

```html
<link rel="stylesheet" href="../../rh-navigation-primary/rh-navigation-primary-lightdom.css">
```

11ty (and the dev server plugins) will rewrite those URLs, so that the URL in the browser is `http://localhost:8000/elements/navigation-link/demo/has-subresource/`

What the dev server will need to do then is add another '..' to the url, so that it lines up with the files on disk

What 11ty needs to do is rewrite the subresource url so that it resolves to `/assets/packages/@rhds/elements/rh-navigation-primary/rh-navigation-primary-lightdom.css`

-----

The goal:
- when reading source files, everything is relative to the files on disk, as if we were serving with a very simple http server
- when viewing the demos in the dev server or on the 11ty site, URLs are rewritten to make them prettier (no `.html` suffix), so we have to rewrite the subresource paths

.changeset/poor-mangos-show.md
bennypowers, line 1: changesets are not needed for docs changes, please delete

Issue #2218, opened 2025-03-20
We should apply a `::selection {  background-color / color }` rule with a `light-dark()` switch. Dark mode selection is hard to read right now.


bennypowers:
We should apply a `::selection {  background-color / color }` rule with a `light-dark()` switch. Dark mode selection is hard to read right now.

_Originally posted by @zeroedin in https://github.com/RedHat-UX/red-hat-design-system/pull/2138#discussion_r1994470526_

coreyvickery:
@bennypowers @zeroedin @marionnegp Put together some options.

https://www.figma.com/proto/g5B0IE3yD7tG9mctlgv9T0/Text-selection-background-color?page-id=1%3A943&node-id=1-944&p=f&viewport=121%2C139%2C1&t=Y17sl9xVTsbJJqMv-1&scaling=min-zoom&content-scaling=fixed

saraDahanCode:
@bennypowers @zeroedin @marionnegp

Just to clarify before I start: should the ::selection styling be applied only to regular text on the page, or should it also cover text inside components with Shadow DOM?
Thanks!

zeroedin:
@saraDahanCode  `::selection` breaks through the shadowdom encapsulation. So it would be applied at the page layer. In this case, the base @layer in docs/styles/styles.css

coreyvickery:
@saraDahanCode If you still need access, please provide your e-mail.

zeroedin:
> [@saraDahanCode](https://github.com/saraDahanCode) If you still need access, please provide your e-mail.

@coreyvickery i shared the code that was needed just need a design review sign off on the PR now.

PR #2713, opened 2025-11-05
docs: add ::selection at page layer to investigate text selection issue

saraDahanCode:
This PR fix to issue #2218 
### What I did
Added the `::selection` styles at the `base` layer (`docs/styles/styles.css`) as suggested,  
so that text selection colors are applied at the page layer (outside the shadow DOM).  

### What I expected
The selection color should appear when selecting text on the documentation page.

### What actually happens
After adding the styles, there is no visible change — the page seems not to load the `::selection` styles at all.

### Request
Could you please take a look and let me know if I'm missing something?  
I followed the suggestion to define the colors in the base layer,  
but it seems the styles are not being applied to the page.
@bennypowers  @zeroedin  @marionnegp

### DIFF
--- a/.changeset/olive-geese-pump.md
+++ b/.changeset/olive-geese-pump.md
@@ -1,7 +1,0 @@
----
-"@rhds/elements": patch
----
-
-Added `::selection` styles to the base layer in `docs/styles/styles.css` to ensure consistent text selection color across pages.  
-Improved visibility and accessibility of selected text for users.
-


docs/styles/styles.css
```diff
@@ -79,6 +79,11 @@
     white-space: nowrap;
     border: 0;
   }
+
+  ::selection {
+    background-color: var(--rh-color-text-selection, currentColor);
+    color: var(--rh-color-text-selection-foreground, inherit);
+  }
```
zeroedin, line 89: Based on the figma file shared in the issue, the background-color for ::selection should be:

```Suggestion
  ::selection {
    background-color: 
      light-dark(
        var(--rh-color-blue-20, #b9dafc), 
        rgba(from var(--rh-color-blue-40, #4394e5) r g b / 50%)
      );
  }
```
We shouldn't have to modify the font color at this time.

saraDahanCode: I don't have access to the Figma file yet

zeroedin: Ok, no worries — move forward with the suggested code above.

saraDahanCode:
@zeroedin 
I've implemented the suggested changes, but I’m not seeing any effect.
 Could you clarify where I should expect to see the impact of this definition?

zeroedin:
> @zeroedin I've implemented the suggested changes, but I’m not seeing any effect. Could you clarify where I should expect to see the impact of this definition?

When you select text, which is what `::selection` is referring to https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/::selection

Screenshots taken from your deploy preview:
image

image

Also, you'll want to run `npm run lint`; it looks like there are some spacing issues. Likely my fault with the suggestion code.

saraDahanCode:
@zeroedin 
So, the changes are not visible locally when running npm run dev , understood.....
I’m currently working on fixing the linting issues mentioned.

zeroedin:
> @zeroedin So, the changes are not visible locally when running npm run dev, understood..... I’m currently working on fixing the linting issues mentioned.

Yes, this is a docs (ux.redhat.com) side fix only. localhost:8080 (docs) vs localhost:8000 (dev server).

If you want to explore adding it to the dev server side, add the same code to `docs/styles/demo/dev-server.css`. That file is included in the dev server by `.pfe.config.json` in the project's root.

zeroedin (changes requested):
Please delete the changeset,  documentation changes that do not change exported code do not need a change set.

.changeset/olive-geese-pump.md
bennypowers, line 1: Changes that only affect docs don't need a change set, please delete this file

Issue #2688, opened 2025-10-22
[bug] `rh-code-block`  Show more / Show less button not appearing when line numbers are hidden


tejas161:
### Describe the bug

It seems like [hiding line numbers](https://ux.redhat.com/elements/code-block/demos/#demo-hide-line-numbers) and  [Expand/Collapse](https://ux.redhat.com/elements/code-block/style/#expand-or-collapse) features aren’t syncing properly.
For my use-case, the codeblock has more than 5 lines, but the Show more / Show less button doesn’t appear initially. However, as soon as I click on any of the action buttons (like copy) the Show more / Show less button suddenly shows up.

### RHDS version number

3.0.3

### Steps to reproduce

1. Create a code block with more than 5 lines.
2. Hide line numbers.
3. Observe that the Show more / Show less button does not appear initially.
4. Click on any of the action buttons (e.g., Copy).
5. Notice that the Show more / Show less button suddenly appears.

### Expected behaviour

The Show more / Show less button should be visible initially (when applicable), regardless of whether line numbers are hidden.

### Element HTML

```HTML
<rh-code-block line-numbers="hidden" actions="wrap copy">
<script type="text/css">#content {
  display: block;
  background-color: var(--rh-color-surface-lighter, #f2f2f2);
  border: var(--rh-border-width-sm, 1px) solid var(--rh-color-border-subtle-on-light, #c7c7c7);
  border-block-start-width: var( --rh-code-block-border-block-start-width, var(--rh-border-width-sm, 1px));
  font-family: var(--rh-font-family-code, RedHatMono, "Red Hat Mono", "Courier New", Courier, monospace);
  color: var(--rh-color-text-primary-on-light, #151515);
  padding: var(--rh-space-xl, 24px);
  height: calc(100% - 2 * var(--rh-space-xl, 24px));
  border-radius: var(--rh-border-radius-default, 3px);
  max-width: 1000px;
  max-height: 640px;
  overflow-y: auto;
}</script>
</rh-code-block>
```

### Operating System (OS)

macOS

### Browser

Chrome

tejas161:
Initial Loading:-

<img width="1710" height="1107" alt="Image" src="https://github.com/user-attachments/assets/1b7287e7-70d1-4765-8e2c-c3db950da938" />




After clicking on copy button:-

<img width="1710" height="1107" alt="Image" src="https://github.com/user-attachments/assets/15994e17-0caa-41e4-b44f-9491f862d521" />

PR #2720, opened 2025-11-06
fix(code-block): show more button appears when line numbers are hidden

saraDahanCode:
## What I did

Fixed a bug where the "Show more" / "Show less" button was not appearing when line numbers were hidden in `<rh-code-block>`.

## Testing Instructions

1. Open the code block demo page.
2. Set the `line-numbers` attribute to `hidden`.
3. Verify that the "Show more" / "Show less" button is now visible and works correctly.

## Notes to Reviewers

- Related issue: [#2688](https://github.com/RedHat-UX/red-hat-design-system/issues/2688)

### DIFF
--- a/.changeset/strong-candies-wave.md
+++ b/.changeset/strong-candies-wave.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-code-block>`: fixed the visibility of the Show more/less button in `<rh-code-block>` when line numbers are hidden.
+  


.changeset/open-spiders-fail.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+Fixed the visibility of the Show more/less button in `<rh-code-block>` when `line-numbers="hidden"`.
```
bennypowers: see changelog.md for changeset format
```suggestion
`<rh-code-block>`: fixed the visibility of the Show more/less button in `<rh-code-block>` when line numbers are hidden.
```

elements/rh-code-block/rh-code-block.ts
```diff
@@ -158,6 +158,8 @@ export class RhCodeBlock extends LitElement {
 
   @state() private copyButtonState: 'default' | 'active' | 'failed' = 'default';
 
+  @state() lines: string[] = [];
```
bennypowers: let's keep this private. you can just call `requestUpdate`

elements/rh-code-block/rh-code-block.ts
```diff
@@ -158,6 +158,8 @@ export class RhCodeBlock extends LitElement {
 
   @state() private copyButtonState: 'default' | 'active' | 'failed' = 'default';
 
+  private lines: string[] = [];
```
bennypowers:
```suggestion
  #lines: string[] = [];
```

try this and see if it still works

saraDahanCode: locally it works, 
but I see that the checks failed

PR #2722, opened 2025-11-06
fix(footer): move landmark role to host

### LINKED ISSUE #2686: [bug] `<rh-footer>` not being exposed as a page landmark
**Edit: looks like this is being caused by RHDC lazy loading the footer upon scroll.**

[Here's why we still may want to address it.](https://github.com/RedHat-UX/red-hat-design-system/issues/2686#issuecomment-3429544807)

The following is what I thought was happening, initially:

------

Pages at redhat.com currently have no footer (aka contentInfo) landmark upon load. Screen reader users attempting to navigate the page via landmarks can only access the navigation and main landmarks--not the footer.

`<rh-footer>` currently has no role. Instead, it contains a `<footer>` (and its implicit contentInfo role) within it. ([Relevant source here](https://github.com/RedHat-UX/red-hat-design-system/blob/bbb5b3c37418620c0ce40fe5bb9462f804ef7ec6/elements/rh-footer/rh-footer.ts#L71))

@adamjohnson suggested we may want to set `role="contentInfo"` on `<rh-footer>` and remove the `<footer>` element from the shadow DOM. I'm not sure what knock-on effects this will have for styles, if any.

If you want to verify this issue for yourself, you can do so at redhat.com.

1. On a Mac, turn on VoiceOver.
2. Go to www.redhat.com in your browser (Safari preferred, but any will work).
3. Open the Rotor with `control-option-u`, and press the arrow keys until you see the page landmarks.
4. You will _not_ see a footer landmark listed.

VoiceOver screenshot of redhat.com showing no footer landmark

_h/t Jairus Mitchell for asking about footers at RHDC and getting this discussion started._

adamjohnson:
## What I did

1. Fixes #2686 
2. Removed `<footer>` elements from `<rh-footer>` and `<rh-footer-univeral>`'s shadowdom.
3. Using InternalsController to apply the [`contentinfo` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/contentinfo_role) to each of the above's host elements.
4. `<rh-footer-universal>` only gets the contentinfo role IF it's not contained in a parent `<footer>` or `<rh-footer>` element.

## Testing Instructions

1. Open [the DP](https://deploy-preview-2722--red-hat-design-system.netlify.app/elements/footer/).
2. Verify no `<footer>` elements exist in the shadowdom of `<rh-footer>` and `<rh-footer-universal>`.
3. On the [`<rh-footer>` demo](https://deploy-preview-2722--red-hat-design-system.netlify.app/elements/footer/demo/), verify that `<rh-footer>` has the `contentinfo` role via the a11y tree.
4. Then, on the same demo, verify that `<rh-footer-universal>` does NOT have the `contentinfo` role.
5. Now move to the [`<rh-footer-universal>` demo in the DP](https://deploy-preview-2722--red-hat-design-system.netlify.app/elements/footer/demo/footer-universal/).
6. Verify that `<rh-footer-universal>` DOES have the `contentinfo` role since it's not wrapped by a `<footer>` or `<rh-footer>` element.
7. Ensure rh-footer remains in spec for design.
8. Ensure there are no axe/lighthouse errors that shouldn't be there.
9. Check the formatting of the changeset (do we usually put sub-components in there?)

### DIFF
--- a/elements/rh-footer/rh-footer-universal.ts
+++ b/elements/rh-footer/rh-footer-universal.ts
@@ -52,9 +52,26 @@ export class RhFooterUniversal extends LitElement {
     'tertiary',
   );
 
+  #hasAncestorH2 = false;
+
   override connectedCallback() {
     super.connectedCallback();
     this.#updateRole();
+    this.#hasAncestorH2 = this.#detectAncestorH2();
+  }
+
+  /** Check if an h2 already exists in the parent context. */
+  #detectAncestorH2(): boolean {
+    let node: HTMLElement | null | undefined = this.parentElement;
+    while (node) {
+      if (node?.closest('h2')
+        || node?.querySelector('h2')
+        || node?.shadowRoot?.querySelector('h2')) {
+        return true;
+      }
+      node = node.parentElement;
+    }
+    return false;
   }
 
   /**
@@ -93,20 +110,9 @@ export class RhFooterUniversal extends LitElement {
   override render() {
     const hasTertiary = this.#slots.hasSlotted('tertiary');
 
-    // determine if h2 already exists in parent context
-    let node: HTMLElement | null | undefined = this.parentElement;
-    let h2: HTMLElement | null | undefined = null;
-    while (!!node && !h2) {
-      h2 = h2
-        || node?.closest('h2')
-        || node?.querySelector('h2')
-        || node?.shadowRoot?.querySelector('h2');
-      node = node.parentElement;
-    }
-
     return html`
       <div class="footer">
-        <h2 id="global-heading" ?hidden="${!!h2}">
+        <h2 id="global-heading" ?hidden="${this.#hasAncestorH2}">
           <!-- Visually-hidden heading announced by screen readers to identify the footer landmark. Expects inline text. Defaults to "Red Hat footer". -->
           <slot name="heading">Red Hat footer</slot>
         </h2>
--- a/elements/rh-footer/rh-footer.ts
+++ b/elements/rh-footer/rh-footer.ts
@@ -68,10 +68,38 @@ export class RhFooter extends LitElement {
 
   override connectedCallback() {
     super.connectedCallback();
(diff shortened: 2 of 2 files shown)


hellogreg (approved):
landmark gets total makeover

elements/rh-footer/rh-footer-universal.ts
```diff
@@ -31,26 +33,59 @@ export class RhFooterUniversal extends LitElement {
...
+        break;
+      }
+
+      node = node.parentElement;
+    }
+
+    if (!hasFooterAncestor) {
+      this.#internals.role = 'contentinfo';
+    }
+  }
```
zeroedin, line 108: We should use @lit/context here instead of doing this parentElement check.

bennypowers:
I generally approve and im sure @paceaux will too

Just want to make sure we aren't breaking backwards

elements/rh-footer/rh-footer-universal.ts
zeroedin: @bennypowers thoughts on this? I went ahead and mocked up the changes locally if we want to use @lit/context. If not I can revert and resolve the comment.

bennypowers: This is to avoid breaking backwards, right?

It's not going to be so easy for us to set every native footer as a context host on the page.

You could potentially conceive of a situation where there's like a custom footer element that uses element internals that we'd have to watch out for.

But I think maybe for most of our cases this is going to be plenty, although maybe we should leave a comment here that explains the limitations of the approach. as well as some documentation on the website.

zeroedin: Ok yes I follow.  

the lit context would only work in `<rh-footer><rh-footer-universal>` pairing.  Not `<footer><rh-footer-universal>`.    I'll resolve and revert my changes locally. Didn't think of that scenario.

PR #2726, opened 2025-11-09
fix(accordion): show content when JS is disabled

### LINKED ISSUE #2059: [bug] `<rh-accordion>` demo seems to hide content when JS doesn't load
### Describe the bug

The [demo for `<rh-accordion>`](https://ux.redhat.com/elements/accordion/demo/) seems to hide the content of the accordion panels when JS doesn't load.

IMO, accordion should degrade gracefully—showing content when JS fails.

### Screenshots

image

image

### Steps to reproduce

1. Go to [rh-accordion demo](https://ux.redhat.com/elements/accordion/demo/)
2. Disable JS
3. Refresh and see content is hidden and inaccessible

### Expected behaviour

Failing `open`

### Element HTML

_No response_

### Operating System (OS)

iOS

### Browser

Chrome

_No response_

Gili-Greenberger:
Fixes #2059

## What I did

1. Created a new `rh-accordion-lightdom-shim.css` file that provides fallback styling when JavaScript is disabled or not yet loaded:
2. Shows all accordion panels by default when elements are `:not(:defined)`
3. Uses `@media (scripting: enabled)` to hide non-expanded panels only when JavaScript is available
4. Linked the shim file in all accordion demo pages

## Testing Instructions

1. Verify that panels remain visible when JavaScript is disabled.

### DIFF
--- a/elements/rh-button-group/docs/40-accessibility.md
+++ b/elements/rh-button-group/docs/40-accessibility.md
@@ -1,6 +1,102 @@
+## Roles and semantics
+
+### role="group"
+
+`rh-button-group` exposes `role="group"` automatically via ElementInternals. The [group role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role) identifies a logical collection of related controls that is not intended to be included in a page summary or table of contents by assistive technologies. Since we use [ElementInternals](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals) to set `role="group"`, authors do not need to set it.
+
+#### Naming the group (optional)
+
+Naming `<rh-button-group>` is optional: the [group](https://w3c.github.io/aria/#group) role does not require an accessible name. Only add `aria-labelledby`/ `aria-label` when several groups on one page could be hard to distinguish, or when the collection’s purpose is not obvious from nearby visible text. Ensure the label describes the container only, not each button.
+
+If you wrap the group in a toolbar, put the label on the element with `role="toolbar"` (see [Toolbar labeling](#toolbar-labeling)).
+
+#### Example: two named groups
+
+```html rhcodeblock
+<h3 id="view-actions">View</h3>
+<rh-button-group aria-labelledby="view-actions">
+  <rh-button>List</rh-button>
+  <rh-button>Grid</rh-button>
+</rh-button-group>
+
+<h3 id="sort-actions">Sort</h3>
+<rh-button-group aria-labelledby="sort-actions">
+  <rh-button>Name</rh-button>
+  <rh-button>Date</rh-button>
+</rh-button-group>
+```
+
+### Optional role="toolbar"
+
+When the button group acts as a compact set of commonly used actions (a toolbar pattern), wrap it in an element with `role="toolbar"`. The [toolbar role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role) is intended for three or more controls. 
+
+Full toolbar behavior (single tab stop, arrow-key navigation between controls) may require additional focus management; this section covers markup and semantics only. See the ARIA Authoring Practices Guide's [Toolbar Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/toolbar/) and our docs on [Focus order and roving tabindex](/accessibility/development/#focus-order-and-roving-tabindex) for further information.
+
+## Toolbar labeling
+
+- **When to label**: 
+    - When the application has more than one toolbar, authors must supply a label on each toolbar.
+    - When there is only one toolbar, a label is optional but can improve clarity for assistive technologies.
+- **How**: Use `aria-label` or `aria-labelledby` on the element that has `role="toolbar"`.
+
+## aria-orientation with toolbar
+
+Elements with `role="toolbar"` have an implicit `aria-orientation="horizontal"`. When the toolbar is vertical, set `aria-orientation="vertical"` on the same element so assistive technologies and keyboard expectations (e.g. Up/Down for navigation) are correct. Omit it for horizontal toolbars.
+
+## Toolbar role examples
+
+### Single toolbar (horizontal)
+
+Optional `aria-label` when there is only one toolbar on the page:
+
+```html rhcodeblock
+<div role="toolbar" aria-label="Text formatting">
+  <rh-button-group>
+    <rh-button>Bold</rh-button>
+    <rh-button>Italic</rh-button>
+    <rh-button>Underline</rh-button>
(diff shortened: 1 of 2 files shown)


adamjohnson (changes requested):
This is trickier than what it may look like. Here's some further explanation:

1. This change is being applied in `rh-accordion-panel.css`. This file is only loaded when:
    1. JS is enabled
    1. The template is DSD'ed / SSR'ed

More than likely, we'd want to apply a fix like this inside of an [`rh-accordion-lightdom-shim.css` file](https://ux.redhat.com/get-started/developers/installation/#lightdom-css-shims). Once that happens, we will have to link it in each of the HTML files in the demo and test the output.

The `rh-accordion-lightdom-shim.css` file will likely need some `rh-accordion-{something}:not(:defined)` selectors alongside the scripting media query.

One this change has been made, we'll have to test the following cases for FOUC and if this feature works as intended in the following circumstances:

1. On the dev server (localhost:8000) with JS enabled. Aka no DSD/SSR.
1. On the dev server (localhost:8000) with JS disabled.
2. On the docs server / uxdot deploy previews with JS enabled. Aka DSD enabled.
2. On the docs server / uxdot deploy previews with JS disabled. Aka DSD enabled.

Benny makes a good point about FOUC. Someone with more institutional knowledge will have to decide if:

1. Having some FOUC is OK since this change may prove to be a net positive for when JS fails to load. OR
1. Maybe FOUC should be avoided at all costs and would be seen as a regression from current behavior.

IMO I'm okay with FOUC as it makes the content more accessible in edge cases. It can also theoretically be "fixed" by RHDS implementors if they don't want to show the content in no-JS scenarios.

@zeroedin if I have missed anything, please chime in.

Gili-Greenberger:
@adamjohnson: Before I proceed with the shim changes, I’d like to confirm the right direction.

Should I continue with fix (moving the CSS to `rh-accordion-lightdom-shim.css`), even if it might cause some CLS as @Zeroedin mentioned, or would you rather consider using`<details>` as a longer-term solution instead?

elements/rh-accordion/rh-accordion-lightdom-shim.css
```diff
@@ -0,0 +1,68 @@
+/*
+  NOTE: This shim prioritizes reduced CLS over perfect semantic accessibility.
+  As a result, heading hierarchy and some accessibility semantics may be lost when JS is disabled,
+  but content remains visible and stable.
+  This is a conscious trade-off for better perceived performance.
+*/
+
+/* Lightdom shim for rh-accordion when JavaScript is disabled or not yet loaded */
```
adamjohnson: I would move this comment up to line 2, then have the `NOTE:` after it.

elements/rh-accordion/rh-accordion-lightdom-shim.css
```diff
@@ -0,0 +1,68 @@
...
+    light-dark(
+        var(--rh-color-text-primary-on-light, #151515),
+        var(--rh-color-text-primary-on-dark, #ffffff)
+      );
+  background-color: var(--rh-color-surface-lightest, #ffffff);
+  font-weight: var(--rh-font-weight-body-text-medium, 500);
+  font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif);
+  font-size: var(--rh-font-size-body-text-lg, 1.125rem);
+  padding: var(--rh-space-lg, 16px) var(--rh-space-xl, 24px);
+  border-inline-end:
```
adamjohnson: This property can probably be removed. Not sure if it has a purpose of not. This is what I'm seeing on localhost:8000 with JS disabled:

Screenshot of rh-accordion with arrows pointing to a floating right border to the right of the accordion headers

adamjohnson: @Gili-Greenberger looks like you removed the border property from `rh-accordion-panel:not(:defined)` but not from here. Remove the border property from `rh-accordion-header:not(:defined)` too. Then test with JS disabled on localhost:8000 and localhost:8080.

elements/rh-accordion/rh-accordion-lightdom-shim.css
```diff
@@ -0,0 +1,61 @@
...
+        var(--rh-color-text-primary-on-dark, #ffffff)
+      );
+  font-size: var(--rh-font-size-body-text-md, 1rem);
+}
+
+rh-accordion-panel:not(:defined) > *:not(:first-child) {
+  padding-block-start: var(--rh-space-xl, 24px);
+}
+
+/* Style accordion headers when not defined */
```
adamjohnson:
```suggestion
```

No need for a comment here.

elements/rh-accordion/rh-accordion-lightdom-shim.css
```diff
@@ -0,0 +1,61 @@
+/* Lightdom shim for rh-accordion when JavaScript is disabled or not yet loaded */
+
+/*
+  NOTE: This shim prioritizes reduced CLS over perfect semantic accessibility.
+  As a result, heading hierarchy and some accessibility semantics may be lost when JS is disabled,
+  but content remains visible and stable.
+  This is a conscious trade-off for better perceived performance.
+*/
```
adamjohnson:
```suggestion
/**
 *  Lightdom shim for rh-accordion when JavaScript is disabled or not yet loaded
 *
 * NOTE: This shim prioritizes reduced CLS over perfect semantic accessibility.
 * As a result, heading hierarchy and some accessibility semantics may be lost when JS is disabled,
 * but content remains visible and stable.
 * This is a conscious trade-off for better perceived performance.
*/
```

.changeset/yellow-baths-flow.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-accordion>`: Added `rh-accordion-lightdom-shim.css` as an optional file to help ensure content is visible when JavaScript is disabled or fails to load.
```
adamjohnson:
```suggestion
`<rh-accordion>`: added `rh-accordion-lightdom-shim.css` as an optional file to help ensure content is visible when JavaScript is disabled or fails to load.
```

Gili-Greenberger:
@adamjohnson 
When I run the demo page on localhost:8080, the accordion doesn’t display at all when JavaScript is disabled. After checking, I found that the demo is loaded inside an `<iframe>` whose opacity is set to 0 by default, and the `onload `handler that should change it to 1 doesn’t run when JS is disabled - so the entire demo inside the `<iframe>`remains transparent.

Is this behavior expected in the 8080 environment?
And for testing purposes, is it acceptable that I manually change the iframe’s opacity to 1 in order to see the content?

PR #2729, opened 2025-11-10
fix(code-block): fix wrap style interference

zeroedin:
## What I did

1. Add `!important` to pre wrap style when ::slotted
2. Correct docs styles from interfering when `<code>` is deeply nested in a slotted `<pre>`.

## Testing Instructions

1. View [Deploy Preview](https://deploy-preview-2729--red-hat-design-system.netlify.app/elements/code-block/demos/) HTML tab codeblocks should now wrap correctly.


markcaron (approved):
Lines Get Toggle Mode

PR #2732, opened 2025-11-11
fix(code-block) remove intersection observer, improve performance

zeroedin:
## What I did

1. Tap into the `contentvisibilityautostatechange` CSS event fired by content-visibility. This allows us to remove the intersection observers entirely.
2. ~~Employed a 2 stage intersection observer, outer observer (100% rootMargin) observes proximity to add instance (inner observers). The inner observer then calculates the line numbers if it is inside its intersect window (50% rootMargin).~~
3. Isolated the instantiation of ~~intersection and~~ resize observers only when they are necessary to guard the large computation method of lineNumbers.
4. Fixed usage of `content-visibility` with the companion rule `contain-intrinsic-size`, which fixes the scroll jank on thousands demos.
5. Updated docs to not render `uxdot-demo` for thousands demos

## Testing Instructions

1.  View [deploy preview thousands demos](https://deploy-preview-2732--red-hat-design-system.netlify.app/elements/code-block/demos/#demo-thousands), compare to ux.redhat.com demos

## Notes to Reviewers

It likely wouldn't hurt to generate some flame graphs for the before and after.

### DIFF
--- a/.changeset/floppy-bottles-travel.md
+++ b/.changeset/floppy-bottles-travel.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-navigation-vertical>`: corrected caret position in mobile safari
+  
--- a/.changeset/real-singers-chew.md
+++ b/.changeset/real-singers-chew.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-code-block>`: correct wrap functionality for prerendered code blocks
+  
--- a/docs/styles/third-party/prism-rhds.css
+++ b/docs/styles/third-party/prism-rhds.css
@@ -10,11 +10,15 @@
  * Based off prism.css
  **/
 
-:is(code, pre)[class*='language-'] {
+pre[class*='language-'],
+:not(pre) > code[class*='language-'] {
   color: var(--rh-color-text-primary);
+  background:
+    light-dark(var(--rh-color-surface-lighter),
+      var(--rh-color-surface-dark));
   background: none;
   font-size: 1em;
-  text-align: left;
+  text-align: start;
   white-space: pre;
   word-spacing: normal;
   word-break: normal;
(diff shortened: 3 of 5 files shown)


markcaron:
Looking good to me. No jank on the isolated demo in the DP and much much faster loading on the all demos page, especially when compared to what's live.

elements/rh-code-block/rh-code-block.ts
```diff
@@ -194,26 +190,17 @@ export class RhCodeBlock extends LitElement {
...
-      RhCodeBlock.observedElements.delete(this);
-    }
-
-    // Disconnect inner observer if it exists
-    if (this.#innerIO) {
-      this.#innerIO.disconnect();
-      this.#innerIO = undefined;
-    }
+    this._container?.removeEventListener(
+      'contentvisibilityautostatechange',
```
bennypowers: use the lit template, don't need the ref then or the cleanup code

bennypowers: you can also use a class method then since lit autobinds

zeroedin: Fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/2732/commits/e7fef091d8654e22a7c5a16b61faf04d6663444a

.changeset/legal-camels-train.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-code-block>`: improve load performance when using certain property combinations'
```
adamjohnson:
```suggestion
`<rh-code-block>`: improve load performance when using certain property combinations
```

adamjohnson (approved):
Latency Goes Totally Missing

This fixes the scroll jank which I think folks will be pleased with. A marked performance improvement.

bennypowers:
<a href="https://imgflip.com/i/ad2x9g"></a><div><a href="https://imgflip.com/memegenerator">from Imgflip Meme Generator</a></div>

PR #2738, opened 2025-11-12
docs(announcement): enhance inline documentation

### LINKED ISSUE #2488: [docs] `<rh-announcement>` missing documentation
### rh-announcement
- `close()` method missing description

bennypowers:
## What I did

Enhanced inline documentation for `rh-announcement` component to improve developer experience and API understanding, following the documentation standards established in PR #2548 for `rh-progress-stepper`.

- [ ] The component has no ARIA landmark role on the host. A designer should confirm whether `role="banner"` or `role="complementary"` would improve screen reader discovery, given that it sits above the navigation and outside the `<main>` content. cc @adamjohnson 
- [x] The `close` event is cancelable, but there's no visual feedback when 
`preventDefault()` is called — the close button just does nothing. A designer should evaluate whether a shake animation, tooltip, or other affordance is needed to indicate the announcement cannot be dismissed. We should add guidance to users e.g. "If you cancel the event, you MUST ..." to ensure a consistent UX.
- [ ] The `image-position` attribute only takes effect on mobile (<768px). A designer should confirm whether the attribute name is clear enough, or if something like `mobile-image-position` would better communicate its scope.
- [x] Background images are mentioned in the design guidelines (positioned left/right, behind content) but there's no slot or CSS custom property for background images in the component. A designer should clarify whether background images are achieved via external CSS or if a dedicated slot/property should be added.

Closes #2488

### DIFF
--- a/.changeset/fixy-codey-blockey.md
+++ b/.changeset/fixy-codey-blockey.md
@@ -1,5 +1,0 @@
----
-"@rhds/elements": patch
----
-
-`<rh-code-block>`: prevent tooltip from unexpectedly reappearing
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # @rhds/elements
 
+## 4.0.4
+### Patch Changes
+
+- 7cd8446: `<rh-jump-links>`: fix active state race conditions during rapid clicks and smooth scrolling
+- 0ebfaf0: `<rh-code-block>`: prevent tooltip from unexpectedly reappearing
+- bb581f3: `<rh-cta>`: ensure default variant's arrow consistently follows text on firefox
+- 6f23579: `<rh-switch>`: fix switch rendering on some browsers in light/dark schemes
+- 887f0af: `<rh-tooltip>`: corrects the intentionality delay for a user triggering the tooltip
+
 ## 4.0.3
 ### Patch Changes
 
--- a/docs/foundations/interactions/links/index.md
+++ b/docs/foundations/interactions/links/index.md
@@ -216,10 +216,10 @@ Developers can use the following CSS as a starting point for link underlining:
 - This example adds a `max()` function to the `text-underline-offset` property, allowing this value to grow proportionally at large font sizes.
 - As of this writing, Safari has some issues with the `text-decoration` shorthand property, so we separated it out into its component options `(-color, -line, -style, and -thickness)`.
 
-<rh-code-block>
+<rh-code-block actions="wrap copy">
   <script type="text/css">
     :is(p, ul, ol, dl) a {
-      text-decoration-color: var(--rh-color-gray-50, #707070);
+      text-decoration-color: light-dark(var(--rh-color-gray-50), var(--rh-color-gray-40));
       text-decoration-line: underline;
       text-decoration-style: dashed;
       text-decoration-thickness: var(--rh-border-width-sm, 1px);
--- a/docs/foundations/interactions/links/inline-link-demo.html
+++ b/docs/foundations/interactions/links/inline-link-demo.html
@@ -3,7 +3,7 @@
 
 <style>
   :is(p, ul, ol, dl, h1, h2, h3, h4, h5, h6) a {
-        text-decoration-color: var(--rh-color-gray-50);
+        text-decoration-color: light-dark(var(--rh-color-gray-50), var(--rh-color-gray-40));
         text-decoration-line: underline;
         text-decoration-style: dashed;
         text-decoration-thickness: 1px;
--- a/elements/rh-cta/demo/arrow-follows-text.html
+++ b/elements/rh-cta/demo/arrow-follows-text.html
@@ -1,0 +1,26 @@
+<div id="container">
+  <rh-cta href="#">
+    Register for the webinar
+  </rh-cta>
+  <rh-cta>
+    <a href="#">Register for the webinar</a>
+  </rh-cta>
+  <rh-cta href="#">
+    Lebensabschnitts<wbr>gefährtenverlustangst
+  </rh-cta>
+  <rh-cta>
+    <a href="#">Lebensabschnitts<wbr>gefährtenverlustangst</a>
+  </rh-cta>
+</div>
+
+<style>
+  #container {
+    width: 230px;
+    resize: horizontal;
+    overflow: hidden;
+  }
+</style>
+
+<script type="module">
+  import "@rhds/elements/rh-cta/rh-cta.js";
+</script>
--- a/elements/rh-cta/rh-cta.css
+++ b/elements/rh-cta/rh-cta.css
@@ -10,6 +10,7 @@ a,
 ::slotted(:is(a, button, input)) {
   vertical-align: middle !important;
   word-break: break-word !important;
+  white-space: normal !important;
   display: inline !important;
   color: inherit !important;
   font-family: inherit !important;
@@ -21,10 +22,6 @@ a,
   text-decoration: var(--_text-decoration) !important;
   text-underline-offset: var(--_text-underline-offset) !important;
   z-index: 2 !important;
-}
-
-::slotted(:is(a, button, input)) {
-  white-space: break-spaces !important;
 }
 
 a:after,
@@ -57,6 +54,15 @@ a:after,
   border-color: var(--_border-color, transparent);
   border-radius: var(--rh-border-radius-default, 3px);
   border-width: var(--rh-border-width-sm, 1px);
+
+  &:has(rh-icon) {
+    padding-inline-end:
+      /* 2px is a safety margin to not cut off the arrow icon
+*        when the container is just barely too small */
+      calc(
+          var(--rh-icon-size, 1.125rem) + var(--rh-space-md, 8px) + 2px
+        );
+  }
 
   --rh-color-surface: var(--_background-color) !important;
   --_arrow-size: 13px;
--- a/elements/rh-cta/rh-cta.ts
+++ b/elements/rh-cta/rh-cta.ts
@@ -83,7 +83,7 @@ export class RhCta extends LitElement {
     const isDefault = !variant;
     const svg = isDefault;
     const follower =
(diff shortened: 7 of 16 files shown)


elements/rh-announcement/demo/background-image.html
```diff
@@ -0,0 +1,30 @@
...
+  <rh-cta slot="cta"
+          href="#">Register</rh-cta>
+</rh-announcement>
+
+<link rel="stylesheet"
+      href="../rh-announcement-lightdom-shim.css">
+
+<style>
+  rh-announcement {
+    --rh-color-surface: transparent;
```
adamjohnson:
```suggestion
    --rh-color-surface: #e0f0ff
```

Match the background color of the SVG for smaller viewports.

elements/rh-announcement/demo/background-image.html
```diff
@@ -0,0 +1,30 @@
...
+    background-size: cover;
+
+    @media (max-width: 767px) {
```
adamjohnson: Instead of `max-width` here, let's do this mobile first and use a `min-width` media query. Then apply the background image & fallback `background-color` inside the `min-width` media query.

elements/rh-announcement/demo/color-context.html
```diff
@@ -1,3 +1,4 @@
+<meta itemprop="description" content="Announcement wrapped in rh-context-demo to preview light and dark color palettes side by side.">
```
adamjohnson, line 1: I see this meta tag is superceeding markdown files for element descriptions. Let's get these meta tags hidden from demos (that end users copy and paste) before merging this PR.

elements/rh-announcement/rh-announcement.ts
```diff
@@ -12,40 +12,62 @@ import '@rhds/elements/rh-button/rh-button.js';
...
 
 /**
- * Announcements are flexible surfaces used to group information in a full width banner layout, traditionally across the top of a page.
- * They are used to announce new features, promos, or news.
- * @summary Arranges content and interactive elements in a layout
+ * Full-width banner for promoting events or important messages. SHOULD
+ * be placed above primary navigation. USE `dismissable` to add a close
+ * button. Supports `color-palette` for light/dark themes. Tab and
+ * Enter MUST activate interactive elements. Slotted content provides
+ * screen reader context; SHOULD include meaningful text.
```
adamjohnson:
```suggestion
 * Announcements are flexible surfaces used to group information in a full width banner layout, traditionally across the top of a page.
 * They are used to announce new features, promos, or news.
```

Since these descriptions are being output on the overview page, they must be written for humans to understand and consume. Sure, RFC2119 is important, but it's primarily important for technical documentation. Not necessarily as the very first thing someone would read when getting acquanted with a component.

Perhaps there's another `@tag` to add for AI here? `@description` or `@example` perhaps?

elements/rh-announcement/rh-announcement.ts
```diff
@@ -12,40 +12,62 @@ import '@rhds/elements/rh-button/rh-button.js';
...
 @themable
 export class RhAnnouncement extends LitElement {
   static styles = [styles];
 
-  /** Sets color context for child components, overrides parent context */
+  /**
+   * Sets the color context for child components, overriding any inherited
+   * parent context. Valid values include `light`, `dark`, and other
+   * palette names defined by the design system. Determines surface and
+   * text colors. SHOULD contrast with adjacent surfaces (e.g., AVOID
```
adamjohnson:
```suggestion
   * text colors. should contrast with adjacent surfaces (e.g., avoid
```

IMO we should not capitalize `"MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL"`.

I will cite Abelson, Sussman, and Sussman:

> “Programs must be written for people to read, and only incidentally for machines to execute.”

We should prioritize these docs for human readability first and AI second. I would be surprised if there's a HUGE difference (see what I did there? :-) between caps and no caps.

elements/rh-announcement/rh-announcement.ts
```diff
@@ -12,40 +12,62 @@ import '@rhds/elements/rh-button/rh-button.js';
...
+   * using a dark announcement above a dark navigation).
+   */
   @property({ reflect: true, attribute: 'color-palette' }) colorPalette?: ColorPalette;
 
   /**
-   * Make an announcement dismissable
+   * When true, renders a close button that allows the user to dismiss the
+   * announcement. Pressing Enter or Space on the close button fires a
+   * cancelable `close` event. If the event is not canceled, the element is
+   * removed from the DOM. Defaults to `false`.
```
adamjohnson:
```suggestion
   * removed from the DOM.
```

elements/rh-announcement/rh-announcement.ts
```diff
@@ -71,19 +93,29 @@ export class RhAnnouncement extends LitElement {
...
-            <!--
-              If this slot is used, we expect an image tag with a width and height set.
-              An icon, svg, or use of the icon component are also valid in this region.
-            -->
+            <!-- summary: optional visual such as an icon, logo, or image
+                 description: |
+                   Accepts an \`<img>\`, \`<svg>\`, or \`<rh-icon>\` element. Images
+                   SHOULD have explicit \`width\` and \`height\` attributes set.
+                   Images MAY include \`role="presentation"\`. Icons MAY include an \`accessible-label\`.
+                   Max recommended height is 48px for icons or 20-25px for text logos. -->
```
adamjohnson:
```suggestion
                   Max recommended height is 48px for image logos/icons or 20-25px for text-based logos. -->
```

elements/rh-announcement/rh-announcement.ts
```diff
@@ -71,19 +93,29 @@ export class RhAnnouncement extends LitElement {
...
                class="${classMap({ empty: !this.#slots.hasSlotted('image') })}">
-            <!--
-              If this slot is used, we expect an image tag with a width and height set.
-              An icon, svg, or use of the icon component are also valid in this region.
-            -->
+            <!-- summary: optional visual such as an icon, logo, or image
+                 description: |
+                   Accepts an \`<img>\`, \`<svg>\`, or \`<rh-icon>\` element. Images
+                   SHOULD have explicit \`width\` and \`height\` attributes set.
+                   Images MAY include \`role="presentation"\`. Icons MAY include an \`accessible-label\`.
```
adamjohnson:
```suggestion
                   Images using this slot must follow accessibility best practices.
```

elements/rh-announcement/rh-announcement.ts
```diff
@@ -71,19 +93,29 @@ export class RhAnnouncement extends LitElement {
...
                part="image"
                class="${classMap({ empty: !this.#slots.hasSlotted('image') })}">
-            <!--
-              If this slot is used, we expect an image tag with a width and height set.
-              An icon, svg, or use of the icon component are also valid in this region.
-            -->
+            <!-- summary: optional visual such as an icon, logo, or image
+                 description: |
+                   Accepts an \`<img>\`, \`<svg>\`, or \`<rh-icon>\` element. Images
+                   SHOULD have explicit \`width\` and \`height\` attributes set.
```
adamjohnson:
```suggestion
                   Should have explicit \`width\` and \`height\` attributes set.
```

See my [comment above](https://github.com/RedHat-UX/red-hat-design-system/pull/2738/changes#r2897467454).

elements/rh-announcement/rh-announcement.ts
```diff
@@ -71,19 +93,29 @@ export class RhAnnouncement extends LitElement {
...
             <slot name="image"></slot>
           </div>
           <div id="content">
             <div id="body" class="${classMap({ empty: this.#slots.isEmpty(null) })}">
-              <!-- Any content that is not designated for the header or footer slot, will go to this slot. -->
+              <!-- summary: main body text content for the announcement
+                   description: |
+                     USE a \`<p>\` element for body text. Content SHOULD be
+                     concise (65 characters max recommended). Screen readers
+                     will read this content in DOM order. AVOID long or
```
adamjohnson:
```suggestion
                     Use a \`<p>\` element for body text. Content should be
                     concise (65 characters max recommended). Avoid long or
```

elements/rh-announcement/rh-announcement.ts
```diff
@@ -71,19 +93,29 @@ export class RhAnnouncement extends LitElement {
...
+                     multi-paragraph content. -->
               <slot></slot>
             </div>
             <div id="cta" class="${classMap({ empty: !this.#slots.hasSlotted('cta') })}">
-              <!-- If this slot is used, we expect a rh-cta component. -->
+              <!-- summary: call-to-action link for the announcement
+                   description: |
+                     MUST contain an \`<rh-cta>\` element with an \`href\`
+                     attribute. CTA text SHOULD be 25 characters or fewer. -->
```
adamjohnson:
```suggestion
                     Must contain an \`<rh-cta>\` element with an \`href\`
                     attribute. CTA text should follow established guidelines; 
                     ideally keeping character counts to less than 25 characters. -->
```

Massage wording as necessary. The [CTA docs](https://ux.redhat.com/elements/call-to-action/guidelines/#writing-content) do not set a character limit for CTAs but [Announcement docs](https://ux.redhat.com/elements/announcement/guidelines/#character-counts) do mention it.

elements/rh-announcement/demo/color-context.html
bennypowers: this is a valid point, but we can address it in a later docs PR, and in `cem`

elements/rh-announcement/rh-announcement.ts
bennypowers: see above

elements/rh-announcement/rh-announcement.ts
bennypowers: see above

elements/rh-announcement/demo/color-context.html
bennypowers: #2890

PR #2743, opened 2025-11-12
docs(card): add CSS custom properties documentation

### LINKED ISSUE #2494: [docs] `<rh-card>` missing documentation
### rh-card
- `--rh-card-header-background-on-light` missing description
- `--rh-card-header-background-on-dark` missing description
- `--rh-card-heading-font-weight` missing description
- `--rh-card-heading-font-family` missing description
- `--rh-card-heading-font-size` missing description

bennypowers:
## What I did

Enhanced inline documentation for `rh-card` component to improve developer experience and API understanding.

- [ ] The card has no ARIA role or landmark on the host element. A designer should confirm whether `role="region"` with an `aria-label` derived from the header slot content would improve screen reader navigation in card-heavy layouts.
- [ ] The promo variant automatically computes color palette by stripping `-er`/`-est` suffixes and reapplying them. A designer should confirm this automatic palette mapping is intuitive and whether it should be documented more prominently.
- [ ] The card does not fire any custom events (e.g., no slot-change or variant-change events). A designer should evaluate whether events would improve framework integrations.
- [ ] The `full-width` attribute only applies when `variant="promo"` but there is no runtime warning when used without the promo variant. A designer should decide whether a console warning is appropriate.
- [ ] The card has no explicit keyboard interaction patterns since it delegates to slotted interactive elements. A designer should confirm whether any card-level focus management is needed for single-link card patterns where the entire surface should be clickable.

Closes #2494

### DIFF
--- a/elements/rh-card/rh-card.ts
+++ b/elements/rh-card/rh-card.ts
@@ -12,12 +12,12 @@ import styles from './rh-card.css' with { type: 'css' };
 const PALETTE_RE = /(er|est)+/g;
 
 /**
- * USE cards to group small previews of content with optional calls to action.
- * Cards SHOULD contain a header with a heading level tag (h1-h6) and MUST NOT
+ * Use cards to group small previews of content with optional calls to action.
+ * Cards should contain a header with a heading level tag (h1-h6) and must not
  * replace primary page content. Cards do not manage focus; interactive elements
- * inside (links, CTAs) MUST be keyboard-accessible via Tab and activated with
+ * inside (links, CTAs) must be keyboard-accessible via Tab and activated with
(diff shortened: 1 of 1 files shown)


elements/rh-card/rh-card.ts
adamjohnson, line 1: This file has rfc2119 keywords that are capitalized and they should be lowercase. eg: SHOULD -> should.

markcaron: @adamjohnson done. Can you double-check again?

adamjohnson: LGTM

PR #2751, opened 2025-11-12
docs(navigation-secondary): enhance inline documentation

### LINKED ISSUE #2502: rh-navigation-secondary

bennypowers:
## What I did

Enhanced inline documentation for `rh-navigation-secondary` component to improve developer experience and API understanding.

- [ ] The `accessible-label` property defaults to `"secondary"` which is generic. A designer should recommend whether the default should include the product name or if documentation should more strongly encourage overriding this value.
- [ ] The `--rh-navigation-secondary-z-index` defaults to 102, matching the primary navigation z-index. A designer should confirm whether these should be differentiated when both navigations are visible simultaneously.
- [ ] The `layout` attribute on `rh-navigation-secondary-menu` accepts `'fixed-width'` and `'full-width'` but there is no design guidance on when to use `'fixed-width'`. A designer should provide usage recommendations.
- [ ] The `color-palette` on `rh-navigation-secondary-menu` defaults to `'lightest'` and the JSDoc says menus are "always" lightest. A designer should confirm whether the attribute should be locked or if dark-palette menus are a valid future use case.
- [ ] The current page indicator relies on consumers applying `aria-current="page"` manually. A designer should evaluate whether a utility or helper should be provided to simplify this integration for common CMS platforms.
- [ ] The dropdown chevron is implemented via CSS pseudo-elements with hardcoded rotation angles. A designer should confirm whether an `<rh-icon>` approach would be more consistent with the rest of the design system.
- [ ] The mobile menu has no animation when expanding or collapsing. A designer should confirm whether this is intentional or if a transition should be added for a smoother user experience.

Closes #2502

### DIFF
--- a/elements/rh-badge/demo/caution.html
+++ b/elements/rh-badge/demo/caution.html
@@ -1,3 +1,4 @@
+<meta itemprop="description" content="Badge with caution state indicating items requiring immediate attention.">
 <rh-badge number="7" state="caution">7</rh-badge>
 <rh-badge number="24" state="caution">24</rh-badge>
 <rh-badge number="240" state="caution">240</rh-badge>
--- a/elements/rh-badge/demo/color-context.html
+++ b/elements/rh-badge/demo/color-context.html
@@ -1,3 +1,4 @@
+<meta itemprop="description" content="Badge rendered in light and dark color scheme contexts.">
 <rh-context-demo>
   <rh-badge number="1">1</rh-badge>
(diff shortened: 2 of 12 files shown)


coreyvickery:
@zeroedin Unfortunately no, we don't have access to old Adobe XD files anymore.

zeroedin:
I've moved the design and UX review to it's own issue, going to focus on the css and jsdoc documentation I can conver here.

zeroedin (approved):

Issue #1432, opened 2024-01-17
[feat] Create typography and base styles for design system


markcaron:
We need typography and general base styles for use within the design system. End solution should be a base stylesheet that we can add to our CDN or implementors can include in their sites/apps.

- [x] Discovery CSS solution
- [x] Design: define styles
- [x] Dev: write CSS

**Related issues:**
- #2169

markcaron:
Discussed in office hours:

- Have multiple CSS files, like `reset.css`, `typography.css`, `page.css`, `grid.css`, etc.
- Bundle base CSS files in post process on CDN in a separate `bundles` directory, like `rhds-all.css`
- Pattern-based styles for lightDOM would exist in `rhds/patterns/${pattern-name}/${pattern-name}.css`

coreyvickery:
@markcaron If you see this before our 1:1, let's discuss.

coreyvickery:
Design is [here](https://www.figma.com/proto/HXQ0VullESvAhxDLDUUJT0/Type-base-styles?page-id=2001%3A663&node-id=2001-664&p=f&viewport=110%2C132%2C1&t=KtoxAdv5KKCli8BM-1&scaling=min-zoom&content-scaling=fixed), has been done for a while. Ready for development.

markcaron:
Opted for a CSS foundations page under Get Started: https://github.com/RedHat-UX/red-hat-design-system/pull/2755

PR #2755, opened 2025-11-13
docs: adding CSS patterns for typography, grid, etc for developers

markcaron:
## What I did

1. Added a `CSS foundations` section to the Get Started: Developers guide with foundational CSS patterns for typography, grid/layout, etc.
2. Added an InlineCSS code block plugin to handle the code snippets more efficiently.
3. Moved the magic line-length CSS from UX Dot's `typography.css` to its `flow` layer in `styles.css`, since it seems more appropriate.
5. Updated UX Dot's `reset.css` to follow a ["probably slightly better best practice"](https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/)

## Testing Instructions

1. Check out the [CSS foundations section](https://deploy-preview-2755--red-hat-design-system.netlify.app/get-started/developers/css-foundations/)
2. Proof read
3. Double check code quality

### DIFF
--- a/.changeset/green-pandas-peel.md
+++ b/.changeset/green-pandas-peel.md
@@ -1,5 +1,0 @@
----
-"@rhds/elements": patch
----
-
-Corrected import paths for `lit` dependencies
--- a/.changeset/ripe-bobcats-bathe.md
+++ b/.changeset/ripe-bobcats-bathe.md
@@ -1,6 +1,0 @@
----
-"@rhds/elements": patch
----
-
-`<rh-navigation-primary>`: correct rendering issue for safari 26.2
-  
--- a/.changeset/wicked-bobcats-argue.md
+++ b/.changeset/wicked-bobcats-argue.md
@@ -1,5 +1,0 @@
----
-"@rhds/elements": patch
----
-
-Updated `@rhds/icons` to v2.1, adding 82 new icons to the UI and Standard sets!
--- a/.config/cem.yaml
+++ b/.config/cem.yaml
@@ -7,7 +7,7 @@ generate:
   noDefaultExcludes: false
   designTokens:
     spec: npm:@rhds/tokens/json/rhds.tokens.json
-    prefix: --rh
+    prefix: rh
   demoDiscovery:
     fileGlob: elements/*/demo/*.html
     urlPattern: elements/:tag/demo/:demo.html
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @rhds/elements
 
+## 4.0.3
+### Patch Changes
+
+- 5e77e43: `<rh-code-block>`: add `visible` value to typescript types for the `line-numbers` attribute
+- be03ad5: Corrected import paths for `lit` dependencies
+- d703934: `<rh-navigation-primary>`: correct rendering issue for safari 26.2
+- 25990af: Updated `@rhds/icons` to v2.1, adding 82 new icons to the UI and Standard sets!
+
 ## 4.0.2
 ### Patch Changes
 
--- a/docs/_includes/layouts/pages/element.11ty.ts
+++ b/docs/_includes/layouts/pages/element.11ty.ts
@@ -828,7 +828,7 @@ export default class ElementsPage extends Renderer<Context> {
               <thead>
                 <tr>
                   <th>Token</th>
-                  <th>Summary</th>
+                  <th>Description</th>
                   <th>Copy</th>
                 </tr>
               </thead>
@@ -839,7 +839,9 @@ export default class ElementsPage extends Renderer<Context> {
                       <code>${token.name}</code>
                     </a>
                   </td>
-                  <td>${await this.#innerMD(token.summary ?? '')}</td>
+                  <td>${!token.summary ? '' : await this.#innerMD(token.summary)}
+                    ${!token.description ? '' : await this.#innerMD(token.description)}
+                  </td>
                   ${copyCell(token)}
                 </tr>`))).join('')}
               </tbody>
--- a/elements/rh-card/rh-card.css
+++ b/elements/rh-card/rh-card.css
@@ -36,7 +36,12 @@
   flex-direction: column;
   justify-items: flex-start;
   overflow: hidden;
-  border-radius: var(--_border-radius, var(--rh-border-radius-default, 3px));
+  border-radius:
+    var(
+        --_border-radius,
+        /** Card's default border radius */
+        var(--rh-border-radius-default, 3px)
+      );
   font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif) !important;
   line-height: var(--rh-line-height-body-text, 1.5);
   border-style: solid;
--- a/elements/rh-code-block/rh-code-block.ts
+++ b/elements/rh-code-block/rh-code-block.ts
@@ -159,7 +159,7 @@ export class RhCodeBlock extends LitElement {
   @property({ type: Boolean }) wrap = false;
 
   /** When set to `hidden`, the code block's line numbers are hidden */
-  @property({ reflect: true, attribute: 'line-numbers' }) lineNumbers?: 'hidden';
+  @property({ reflect: true, attribute: 'line-numbers' }) lineNumbers?: 'hidden' | 'visible';
 
   @state() private copyButtonState: 'default' | 'active' | 'failed' = 'default';
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rhds/elements",
-  "version": "4.0.2",
+  "version": "4.0.3",
   "description": "Red Hat Design System Elements",
   "overrides": {
     "playwright": "~1.48.0"
--- a/uxdot/ssr-failure-recoverable.ts
+++ b/uxdot/ssr-failure-recoverable.ts
@@ -1,5 +1,5 @@
 import type { RhTabs } from 'elements/rh-tabs/rh-tabs.js';
-import { LitElement, type PropertyValues } from 'lit';
+import { LitElement, render, type PropertyValues } from 'lit';
 
 function isLitElement(e: Element): e is LitElement {
   return 'updateComplete' in e;
@@ -30,6 +30,15 @@ async function forceUpdate(e: LitElement) {
   }
 }
 
+function forceRender(e: LitElement) {
+  render(
+    // @ts-expect-error: this is a workaround for SSR failures
+    e.render(),
+    e.renderRoot,
+    e.renderOptions,
+  );
+}
+
 function forceHydration(node: Element) {
   for (const e of node.shadowRoot?.querySelectorAll('*') ?? []) {
     e.removeAttribute('defer-hydration');
@@ -48,6 +57,7 @@ export class SSRFailureRecoverableElement extends LitElement {
       if (e instanceof Error && e.message.startsWith('Hydration')) {
         // eslint-disable-next-line no-console
         console.warn(e);
+        forceRender(this);
         forceHydration(this);
       } else {
         throw e;
--- a/uxdot/uxdot-pattern-ssr-controller-client.ts
+++ b/uxdot/uxdot-pattern-ssr-controller-client.ts
@@ -1,16 +1,61 @@
 import { RHDSSSRController } from '@rhds/elements/lib/ssr-controller.js';
 import { noChange, type LitElement } from 'lit';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 
 /** Hydrate the results of SSR on the client */
 export class UxdotPatternSSRControllerClient extends RHDSSSRController {
-  allContent = noChange;
-  htmlContent = noChange;
-  jsContent = noChange;
-  cssContent = noChange;
-  hasCss = noChange;
-  hasJs = noChange;
+  allContent: ReturnType<typeof unsafeHTML> | typeof noChange = noChange;
+  htmlContent: ReturnType<typeof unsafeHTML> | typeof noChange = noChange;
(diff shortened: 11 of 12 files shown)


docs/assets/examples/css/typography.css
```diff
@@ -0,0 +1,59 @@
...
+:where(h6) {
+  font-size: var(--rh-font-size-heading-xs);
+}
+
+:where(p) {
+  font-size: var(--rh-font-size-body-text-md);
+}
+
+:where(p, h1, h2, h3, h4, h5, h6, li) {
+  max-width: 56rem; /* warning: magic number */
```
bennypowers: this messed up some layouts that claude generated for me with cem mcp and rhds. hero heading looked off-center.

prompt was:

> Compose a landing page for OpenShift with navigation
  and a three card section

markcaron: Yeah, the intent, here, is to limit line-lengths in general page content on ux.redhat.com. It might be a bit heavy handed for a boilerplate-type CSS.

Thoughts @zeroedin @coreyvickery ?

zeroedin: Yeah, likely the max-width could be moved to a uxdot-specific stylesheet.  Max-width in this use case is more a function of layout and readability.

zeroedin: Maybe belongs in `flow` layer

docs/_includes/styles/reset.css
```diff
@@ -0,0 +1,53 @@
+/* Box sizing rules */
+*,
+*:before,
+*:after {
+  box-sizing: border-box;
+}
```
zeroedin, line 10:
```Suggestion
html {
  box-sizing: border-box; /* Set the root element's box-sizing */
}

*,
*:before,
*:after {
  box-sizing: inherit; /* Make all elements inherit the root's setting */
}
```

Let's update this to use the inheritance model

docs/assets/examples/css/typography.css
markcaron: Seems like flow, IMO.

markcaron: Went ahead and moved this nugget to the `flow` layer in my last commit, and removed it from the example CSS for typography.

bennypowers: Thank you

docs/_includes/styles/reset.css
```diff
@@ -0,0 +1,57 @@
+/* Box sizing rules */
+html {
+  box-sizing: border-box; /* Set the root element's box-sizing */
+}
+
+*,
+*:before,
+*:after {
+  box-sizing: inherit; /* Make all elements inherit the root's setting */
```
adamjohnson, line 9: I see this is the ["probably slightly better best practice"](https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/) you linked to in the PR body; however, when I look at [the DP](https://deploy-preview-2755--red-hat-design-system.netlify.app/get-started/developers/css-foundations/#reset) it shows the following:

```css
*,
*:before,
*:after {
  box-sizing: border-box;
}
```

Something must not be updating somewhere...?

docs/_includes/styles/reset.css
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/* Remove default margin */
+:where(body,
```
adamjohnson, line 13: Any reason to do this over:

```css
* {
  margin: 0;
}
```

Not saying one way is better than the other. Curious if there's intent here.

docs/_includes/styles/reset.css
```diff
@@ -0,0 +1,57 @@
...
+blockquote,
+dl,
+dd) {
+  margin: 0;
+}
+
+/* Default scroll behavior */
+html:focus-within {
+  scroll-behavior: smooth;
+}
```
adamjohnson, line 33: Instead of doing it this way, you might consider:

```css
@media (prefers-reduced-motion: no-preference) {
  html:focus-within {
    scroll-behavior: smooth;
  }
}
```

Then removing lines 43-57. 

This way we explicitly allow it when a user has `no-preference` and don't need to undo it otherwise. Less bytes down the wire.

docs/_includes/styles/reset.css
adamjohnson, line 1: We may consider adding:

1. [Improving media defaults](https://www.joshwcomeau.com/css/custom-css-reset/#six-improve-media-defaults-7)
2. [Enable keyword animations](https://www.joshwcomeau.com/css/custom-css-reset/#three-enable-keyword-animations-4) (add this to the `no-preference` block mentioned below, if applicable)

docs/assets/examples/css/fonts.css
adamjohnson, line 1: Pedantic me would love to see these properties alphabetized. 

```css
@font-face {
  font-family: RedHatDisplay;
  font-style: normal;
  font-weight: 400;
  src: url('path/to/fonts/RedHatDisplay/RedHatDisplay-Regular.woff');
  text-rendering: optimizelegibility;
}

/* ...et all... */
```

YMMV.

docs/assets/examples/css/typography.css
```diff
@@ -0,0 +1,55 @@
+body {
+  font-family: var(--rh-font-family-body-text);
+  font-size: 1rem;
```
adamjohnson:
```suggestion
  font-size: var(--rh-font-size-body-text-md, 1rem);
```

docs/assets/examples/css/typography.css
adamjohnson, line 1: These tokens should probably have fallbacks baked in, eg:

```diff
-  line-height: var(--rh-line-height-heading);
+  line-height: var(--rh-line-height-heading, 1.3);
```

adamjohnson: Also, consider alphabetizing properties, should it suit you.

docs/assets/examples/css/typography.css
```diff
@@ -0,0 +1,55 @@
...
+
+:where(code) {
+  font-size: inherit;
+  font-family: var(--rh-font-family-code);
+  line-height: var(--rh-line-height-code);
+}
+
+:where(kbd) {
+  font-family: var(--rh-font-family-body-text);
+  font-size: 1rem;
```
adamjohnson: Should this be:

```suggestion
  font-size: var(--rh-font-size-body-text-md, 1rem);
```

?

docs/get-started/developers/css-foundations.md
```diff
@@ -0,0 +1,101 @@
...
+
+```css code-block {dedent: true, language: "css", highlighting: "prerendered"}
+{% set cssContent = "./docs/styles/reset.css" | inlineCss %}{{ cssContent | safe }}
+```
+
+## Fonts
+
+### Red Hat CDN
+
+If you are on a `*.redhat.com` domain, you can use our CDN to access the Red Hat fonts:
```
adamjohnson, line 41: Curiously, you _can_ hit this CSS file in your browser if you go to the URL. That said, the actual `.woff2` files are blocked by a CORS error for non-Red Hat domains.

docs/get-started/developers/css-foundations.md
```diff
@@ -0,0 +1,101 @@
...
+
+```html code-block {dedent: true, language: "css", highlighting: "prerendered"}
+https://www.redhatstatic.com/dssf-001/v2/@redhat/redhat-font@4.1.0/font.min.css
```
adamjohnson: Should this be wrapped in a `<link>` tag? Or are people meant to copy-paste this URL and see its contents?

```html
<link rel="stylesheet" href="https://www.redhatstatic.com/dssf-001/v2/@redhat/redhat-font@4.1.0/font.min.css">
```

docs/assets/examples/css/layout-grid.css
adamjohnson, line 1: This is over 2,000 lines and will get out of date if we ever change rhx-grid.

I would recommend removing this source file in favor of referencing the file in the repo. 

Then, add example HTML to show what would be required to use this grid. Something like:

```html
<link rel="stylesheet" href="../path/to/rhx-grid.min.css">
...
<div class="rhx-grid" data-rhx-cols="2xs:2 xs:3 md:4">
  <div>Grid Item 1</div>
  <div>Grid Item 2</div>
  <div>Grid Item 3</div>
  <div>Grid Item 4</div>
</div>
```

adamjohnson (commented):
Is this meant to solve https://github.com/RedHat-UX/red-hat-design-system/issues/1432?

docs/assets/examples/css/fonts.css
markcaron: My mileage varies!

docs/get-started/developers/css-foundations.md
markcaron: Addressed in docs: https://deploy-preview-2755--red-hat-design-system.netlify.app/get-started/developers/css-foundations/#red-hat-cdn

docs/assets/examples/css/layout-grid.css
markcaron: Just gonna remove it completely

docs/get-started/developers/css-foundations.md
markcaron: Done!

docs/styles/reset.css
adamjohnson: With the addition of lines 31-34, this can be removed.

You may also choose to move lines 31-34 right above this `prefers-reduced-motion` media query to group them closer together.

docs/assets/examples/css/fonts.css
```diff
@@ -0,0 +1,119 @@
+@font-face {
+  font-family: RedHatDisplay;
+  src: url('path/to/fonts/RedHatDisplay/RedHatDisplay-Regular.woff');
+  font-style: normal;
+  font-weight: 400;
```
adamjohnson, line 5: I wonder if we should convert these font weights to use token values? eg:

```css
--rh-font-weight-body-text-regular: 400;
--rh-font-weight-body-text-medium: 500;
--rh-font-weight-code-regular: 400;
--rh-font-weight-code-medium: 500;
--rh-font-weight-heading-regular: 400;
--rh-font-weight-heading-medium: 500;
--rh-font-weight-heading-bold: 700;
```

docs/assets/examples/css/typography.css
adamjohnson: Looks like this still needs to be applied?

docs/assets/examples/css/typography.css
adamjohnson: Looks like this still needs to be applied too?

docs/assets/examples/css/typography.css
adamjohnson, line 1: If this file is meant to be shared, we should add fallbacks for our tokens. eg:

```diff
- font-size: var(--rh-font-size-heading-xl);
+ font-size: var(--rh-font-size-heading-xl, 2.5rem);
```

docs/styles/typography.css
```diff
@@ -1,59 +1,60 @@
 body {
-  font-family: var(--rh-font-family-body-text);
-  font-size: 1rem;
-  line-height: var(--rh-line-height-body-text);
+  font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif);
+  font-size: var(--rh-font-size-body-text-md, 1rem);
+  line-height: var(--rh-line-height-heading, 1.3);
```
adamjohnson:
```suggestion
  line-height: var(--rh-line-height-body-text, 1.5);
``` 

Much better at 1.5.

docs/styles/typography.css
```diff
@@ -1,59 +1,60 @@
...
+  -webkit-font-smoothing: antialiased;
```
adamjohnson, line 5: This change causes the font to render differently than it does currently. See [DP](https://deploy-preview-2755--red-hat-design-system.netlify.app/get-started/developers/tokens/) vs [uxdot](https://ux.redhat.com/get-started/developers/tokens/).

cc @marionnegp and @coreyvickery What are your thoughts? Yay or nay?

coreyvickery: @adamjohnson The font smoothing is fine, but why was the line height changed? It's usually `1.5` for body text applications and `1.3` for most heading applications (up to `48px`, then it's `1.1`).

In the screenshot, the line height of `18px` is supposed to be `27`, not `23.4`.

Screenshot 2026-02-23 at 4 54 12 PM

docs/assets/examples/css/typography.css
markcaron: I think I changed the main `/styles/typography.css` file last time, instead of this `examples` one. Probably good I did that too.

docs/styles/typography.css
markcaron: Whoops, yeah

docs/styles/typography.css
markcaron: @coreyvickery Line height was my fault. Copied the wrong token. Fixed now.

docs/assets/examples/css/fonts.css
adamjohnson: Looks like changing these to token values caused font rendering issues. Not sure if this was a race condition, browser rendering issue or what. Changing them back to `400` (et all) fixes the issue.

[8af7177](https://github.com/RedHat-UX/red-hat-design-system/pull/2755/commits/8af7177045d28436f383e6d7336e9903a8019846)

zeroedin: Using CSS variables for the font-weight descriptor within an @font-face rule causes the declaration to break because CSS variables cannot be used in this context. The @font-face descriptors require static values or a range of static values for the browser to correctly match the font file to a specific weight or weight range when the font is requested by CSS rules. 

## Why it Breaks
The @font-face rule is a declaration that happens before the browser renders any element, effectively mapping a font file to a specific set of characteristics (like a weight, style, and range of characters). CSS variables, on the other hand, are resolved at runtime and are dependent on the element's position in the DOM and the cascading styles, which is too late for the font-matching process.

PR #2762, opened 2025-11-19
fix(button): alias `label` attribute to `accessible-label`

adamjohnson:
## What I did

1. `<rh-button>` currently uses the `label` attribute to assign an `aria-label` to the internal `<button>` element.
2. We have largely adopted `accessible-label` for this purpose.
3. This PR adds the `accessible-label` property to `<rh-button>`.
4. This PR keeps the `label` property, but aliases it directly to `accessible-label`.

## Testing Instructions

1. View the rh-button [variants demo](https://deploy-preview-2762--red-hat-design-system.netlify.app/elements/button/demo/variants/) in the DP.
2. Inspect both the "Close" and the "Search" icon buttons.
3. Each of these elements uses `accessible-label` and `label`, respectively.
4. Verify each of these elements correctly outputs the attribute text onto the internal `<button>` tag as the `aria-label` value.
5. Verify this is the approach we want to take inside the rh-button component file.
6. Verify the changesets are appropriate.

## Notes to Reviewers

`staging/eevee` will likely need to be updated with `main`, then we can re-update `<rh-alert>` to use `accessible-label` (see #2761 for details).

### DIFF
--- a/elements/rh-button/rh-button.ts
+++ b/elements/rh-button/rh-button.ts
@@ -44,22 +44,16 @@ export class RhButton extends LitElement {
   @property({ reflect: true }) type?: 'button' | 'submit' | 'reset';
 
   /**
-   * Accessible name for the button, use when the button does not have slotted text
+   * Accessible name for the button, use when the button does not have slotted text.
    */
   @property({ attribute: 'accessible-label' }) accessibleLabel?: string;
 
   /**
-   * @deprecated Use `accessible-label` instead
-   * This property maintains backwards compatibility for the old `label` attribute
+   * Accessible name for the button, use when the button does not have slotted text.
+   * Use `accessible-label` instead.
+   * @deprecated
    */
-  @property({ attribute: 'label' })
-  set label(value: string | undefined) {
-    this.accessibleLabel = value;
-  }
-
-  get label(): string | undefined {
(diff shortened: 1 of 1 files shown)


elements/rh-button/rh-button.ts
```diff
@@ -43,8 +43,23 @@ export class RhButton extends LitElement {
...
+   * Accessible name for the button, use when the button does not have slotted text
+   */
+  @property({ attribute: 'accessible-label' }) accessibleLabel?: string;
+
+  /**
+   * @deprecated Use `accessible-label` instead
+   * This property maintains backwards compatibility for the old `label` attribute
+   */
+  @property({ attribute: 'label' })
+  set label(value: string | undefined) {
```
bennypowers: this can stay as a field, and just use it at the call site `this.accessibleLabel || this.label`

bennypowers (changes requested):

elements/rh-button/rh-button.ts
adamjohnson: Resolved in [6f69dd0](https://github.com/RedHat-UX/red-hat-design-system/pull/2762/commits/6f69dd0bf60675077d936d4f822985a6385fabc1).

bennypowers (approved):

PR #2764, opened 2025-11-20
fix(avatar): inline layout converting to block on small viewports

### LINKED ISSUE #2763: [bug] `<rh-avatar>` layout="inline" property does not force an inline layout
`<rh-avatar layout="inline">` does not force an inline layout

zeroedin:
## What I did

1. Corrected `layout="inline"` from converting back to block on small viewports.
2. Updated CSS to use internal classMap for style switching instead of :host([attr]) based styles.
3. Added layout demo, replaces position demo, layouts naming is more explict

Closes #2763 

## Testing Instructions

1. [View Deploy Preview](https://deploy-preview-2764--red-hat-design-system.netlify.app/elements/avatar/demos/)

## Notes to Reviewers

If we are setting the layout of an avatar, we should set expectations that the layout persists regardless of viewport changes.  If the layout attribute is not set, it will proceed with the responsive layout switch.

Given this change, I wonder whether it could break prior expectations, though I feel that was an incorrect application of the API expectations.  So question the `patch` level on the changeset here. 

Also note given https://github.com/RedHat-UX/red-hat-design-system/pull/2754 One of the two will have merge conflicts when either is merged.  Special care to fix will be needed.

### DIFF
--- a/elements/rh-avatar/demo/position.html
+++ b/elements/rh-avatar/demo/position.html
@@ -1,8 +1,0 @@
-<rh-avatar name="Ada Lovelace"
-           subtitle="Computer Programmer"
-           layout="block"></rh-avatar>
-
-<script type="module">
-  import '@rhds/elements/rh-avatar/rh-avatar.js';
-</script>
-


elements/rh-avatar/rh-avatar.css
```diff
@@ -27,123 +32,119 @@
...
-::slotted(a:hover) {
-  color: var(--rh-color-interactive-primary-hover);
-}
+  /** only apply block layout when it is block or mobile and not inline or plain */
+  &:is(.layout-block, .mobile):not(.layout-inline, .plain) {
+    text-align: center;
+    place-items: center;
+    gap: 0;
+    grid-template-columns: minmax(max-content, 250px);
+    grid-template-areas: 'a' 't' 's';
```
adamjohnson, line 48:
```suggestion
    grid-template-areas: 'avatar' 'text' 'subtitle';
```

Consider writing out the full word instead of abbreviating. These meanings were not immediately obvious to me at first glance.

adamjohnson (commented):
An rh-avatar with a black circle for an image

Not sure what's up here, but this is showing a black circle instead of a [generic avatar](https://ux.redhat.com/elements/avatar/demo/variants/) as seen on uxdot. 

Also, probably a good idea to have @marionnegp or @coreyvickery do a quick design review since there's a fair amount of CSS restructuring. 

Should there be a section about `layout="inline"` vs `layout="block"` on the [Style page](https://deploy-preview-2764--red-hat-design-system.netlify.app/elements/avatar/style/)? 

> Given this change, I wonder whether it could break prior expectations, though I feel that was an incorrect application of the API expectations. So question the patch level on the changeset here.

If this will "break" (fix) folks' implementations, it may be better to base this off of `staging/eevee`. IMO people would likely be less tolerant if they updated from 4.0.0 to 4.0.1 and something breaks vs if they went from 4.0 to 4.1.

elements/rh-avatar/rh-avatar.css
zeroedin: This was previous code that didn't need to be refactored in this PR imho.   Can open a seperate PR to make this change later.

zeroedin:
> Also, probably a good idea to have @marionnegp or @coreyvickery do a quick design review since there's a fair amount of CSS restructuring.

Good idea, that said, the design shouldn't have changed at all. The restructuring removed the `:host([attr])` selectors and implemented CSS nesting, so the diff just looks worse than it really is.

> Should there be a section about layout="inline" vs layout="block" on the [Style page](https://deploy-preview-2764--red-hat-design-system.netlify.app/elements/avatar/style/)?

[There is a section](https://ux.redhat.com/elements/avatar/guidelines/#layout), but it needs further explanation. It talks of alignment as horizontal, aka inline, and vertical, aka block, but nothing further about the responsiveness of either.   API implementation. `layout="inline"` pretty much did nothing and just rendered like it wasn't present, allowing the element to then end up in block defeating its purpose of applying the value.

> If this will "break" (fix) folks' implementations, it may be better to base this on `staging/eevee`. IMO, people would likely be less tolerant if they updated from 4.0.0 to 4.0.1 and something breaks VS if they went from 4.0 to 4.1.

This is the age-old question, which is why the question posed is in the review notes. If we deem it's breaking, it wouldn't be 4.1; it would be 5.0.  

IMHO, this gets grey here; the API doesn't change, but the render output does, which realistically is just a style change, and we make other style changes that aren't breaking, but because we have a slight layout change here, we assume it is. This is the really thin line we walk in this decision.

zeroedin:
> Not sure what's up here, but this is showing a black circle instead of a [generic avatar](https://ux.redhat.com/elements/avatar/demo/variants/) as seen on uxdot.

Good catch nesting close } in the wrong place.  Possibly even a cursor tab button insertion without my noticing.

Fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/2764/commits/2b847d2ae29bad4838e247808bba78f24e7b556c

adamjohnson:
> but the render output does, which realistically is just a style change

IMO if we wanna be conservative, lump it in with 4.1. Otherwise, send it as a patch (4.0.1).

elements/rh-avatar/rh-avatar.css
bennypowers, line 1: Maybe a container query?

zeroedin: It can't be a container query right now as a patch.  For one we need to match the existing expectations first. Then you'd need an external container (pattern) as Avatar is not a full-width element, then an API to switch to the responsiveness to the container, as you can't have both the media and container query existing or they'll conflict.

`<rh-avatar responds-to="container">` or maybe an additional option to layout could work. `inline | block | container` where as container is sorta of like not setting `layout` and getting the media query.

bennypowers: here's one i prepared earlier

https://codepen.io/bennyp/pen/emZPExw

zeroedin: I'll have to think about this more, but right now I'm wary of your example using :host { display: block } from the current :host { display: inline-block } and how that could have effects on layout for people.  That might be breaking?

PR #2797, opened 2025-12-10
docs: register rh-chip-group for SSR

### LINKED ISSUE #2796: [bug] `<rh-chip>` Small size is using wrong font size
### Describe the bug

The small chip appears to be using 14px for the font size, but it should be 12px (`--rh-font-size-body-text-xs`).

### Screenshots

Image

### RHDS version number

4.0.1

### Steps to reproduce

Inspect the small chip's text in the [demo](https://ux.redhat.com/elements/chip/demos/#demo-size).

### Expected behaviour

Text should be at 12px.

### Element HTML

```HTML

```

### Operating System (OS)

macOS

### Browser

_No response_

_No response_

bennypowers:
Fixes #2796

## What I did

1. register chip group as an ssr'able element in the uxdot config

## Testing Instructions

1. visit [deploy preview](https://deploy-preview-2797--red-hat-design-system.netlify.app/elements/chip/demos/#demo-size)
2. verify that chips in a chip group with size="sm" in fact render with small text.

## Notes to Reviewers
Since chip group provides the `size` context for the child chips, it must be loaded at ssr time (i.e. when the site builds). Therefore we need to explicitly include it in the config.

### DIFF
--- a/docs/_includes/layouts/pages/element.11ty.ts
+++ b/docs/_includes/layouts/pages/element.11ty.ts
@@ -279,7 +279,6 @@ export default class ElementsPage extends Renderer<Context> {
     const { doc } = ctx;
     const { tagName } = doc.docsPage;
     return [
-      content,
       html`
       <section class="band" id="installation">
         ${this.#header('Importing')}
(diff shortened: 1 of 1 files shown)


zeroedin (approved):
Line Generated The Markup

PR #2802, opened 2025-12-17
feat(select): add `<rh-select>`

### LINKED ISSUE #2557: [feat] Create `<rh-select>` element
### Description

We need to build the `<rh-combobox>` element (which could have an aka of `<rh-select>`) that will follow the [WAI Combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/)

**For the initial release**, we're going to focus on the [`select-only` pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/).

### Acceptance Criteria

- [x] Design done
- [ ] Development done

### Images

Image

The select-only combobox can use the default browser listbox, but here are designs for use cases where they want to customize the listbox (like the scheme switcher on docs.redhat.com).
Image

### Design file

[Figma link](https://www.figma.com/design/BOGsHyL2K07gtPKIXppbHQ/rh-combobox?node-id=4001-769&m=dev&t=Eyv849ftOAMRUU1M-1)

### Other resources

After discussing in office hours, we've clarified that a [Combobox with select-only pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/) combines a toggle element with a listbox of options to select as a value.

This is inherently different from a Menu Dropdown which has a button that toggles a menu list of actions (typically links or buttons). [See WAI menu pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/), like our [`<rh-menu-dropdown>` element](https://github.com/RedHat-UX/red-hat-design-system/issues/2556).

adamjohnson:
## What I did

1. Adds the `<rh-select>`, `<rh-option>`, and `<rh-option-group>` elements.
2. Closes #2557.

## Testing Instructions

1. View the following demos:
    * [Base Demo](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/)
    * [Color context](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/color-context/)
    * [Events](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/events/)
    * [Methods](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/methods/)
    * [Option descriptions](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/option-descriptions/)
    * [Option group](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/option-group/)
    * [Option icons](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/option-icons/)
    * [States](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/demo/states/)
2. Ensure this element meets all the accessibility criteria set forth via the [APG Select-only Combobox](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/)
3. Ensure this design implementation matches the [Figma specs](https://www.figma.com/design/BOGsHyL2K07gtPKIXppbHQ/rh-combobox?node-id=4012-2450&m=dev)
4. Ensure the correct design tokens have been used in rh-select's CSS
5. Ensure rh-select implements the custom `change`, `open`, and `close` events correctly
6. Should we add any `part="xyz"` attributes?
7. Make sure colors in Color Context demo are correct
8. Check Code docs for accuracy
9. Ensure [Select docs](https://deploy-preview-2802--red-hat-design-system.netlify.app/elements/select/) match the [Figma docs](https://www.figma.com/design/NjzH9zRmFYpkq17uoO9GMP/Element---Select?node-id=26-6116&m=dev).

## Notes to Reviewers

  * Should the dropdown (listbox) have a max-height / scroll when there are lots of options?

## Outstanding To Do's:

- [x] Docs
    * Make docs specifying the difference between a select, disclosure, and menu dropdown. Could be Do/Do Not or a11y docs.
    * Ensure a11y docs denote a select must have some form of accessible label (label + placeholder, accessible-label, placeholder only)
- [x] screenshot.png
- [x] Changeset
- [x] Tests
- [x] PR's for ATFocusController, ComboboxController, RovingTabIndexController in PFE to account for patches found in this PR
    * https://github.com/patternfly/patternfly-elements/pull/2963

### DIFF
--- a/.changeset/seven-bugs-wash.md
+++ b/.changeset/seven-bugs-wash.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-tag>`: improved element theming capabilities
+  
--- a/.changeset/wet-singers-try.md
+++ b/.changeset/wet-singers-try.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tile>`: fixed `color-palette` rendering when global color tokens are not already defined
--- a/elements/rh-tag/rh-tag.ts
+++ b/elements/rh-tag/rh-tag.ts
@@ -38,16 +38,16 @@ export class RhTag extends LitElement {
   /**
    * Icon set to display in the tag
    */
-  @property({ attribute: 'icon-set' }) iconSet: IconSetName = 'ui';
+  @property({ attribute: 'icon-set', reflect: true }) iconSet: IconSetName = 'ui';
 
   /** The variant of the tag. */
-  @property() variant?: 'filled' | 'outline' | 'desaturated' = 'filled';
+  @property({ reflect: true }) variant?: 'filled' | 'outline' | 'desaturated' = 'filled';
 
   /** The size of the tag. */
-  @property() size?: 'compact';
+  @property({ reflect: true }) size?: 'compact';
 
   /** optional href for linked tag. */
-  @property() href?: string;
+  @property({ reflect: true }) href?: string;
 
   /**
    * Whether an interactive tag is disabled.
--- a/elements/rh-tile/rh-tile.css
+++ b/elements/rh-tile/rh-tile.css
@@ -342,29 +342,29 @@ svg {
  * specific color themes from the design system.
  */
 :host([color-palette^='dark']) {
-  --_interactive-color: var(--rh-color-border-interactive-on-dark);
-  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-dark);
-  --_text-color-secondary: var(--rh-color-text-secondary-on-dark);
-  --_disabled-background-color: var(--rh-color-surface-dark);
-  --_border-color: var(--rh-color-border-subtle-on-dark);
+  --_interactive-color: var(--rh-color-border-interactive-on-dark, #92c5f9);
+  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-dark, #b9dafc);
+  --_text-color-secondary: var(--rh-color-text-secondary-on-dark, #c7c7c7);
+  --_disabled-background-color: var(--rh-color-surface-dark, #383838);
+  --_border-color: var(--rh-color-border-subtle-on-dark, #707070);
 
-  color: var(--rh-color-text-primary-on-dark);
+  color: var(--rh-color-text-primary-on-dark, #ffffff);
 }
 
-:host([color-palette='darkest']) { background-color: var(--rh-color-surface-darkest); }
-:host([color-palette='darker']) { background-color: var(--rh-color-surface-darker); }
-:host([color-palette='dark']) { background-color: var(--rh-color-surface-dark); }
+:host([color-palette='darkest']) { background-color: var(--rh-color-surface-darkest, #151515); }
+:host([color-palette='darker']) { background-color: var(--rh-color-surface-darker, #1f1f1f); }
+:host([color-palette='dark']) { background-color: var(--rh-color-surface-dark, #383838); }
 
 :host([color-palette^='light']) {
-  --_interactive-color: var(--rh-color-border-interactive-on-light);
-  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-light);
-  --_text-color-secondary: var(--rh-color-text-secondary-on-light);
-  --_disabled-background-color: var(--rh-color-surface-light);
-  --_border-color: var(--rh-color-border-subtle-on-light);
+  --_interactive-color: var(--rh-color-border-interactive-on-light, #0066cc);
+  --_focus-interactive-color: var(--rh-color-interactive-primary-hover-on-light, #003366);
+  --_text-color-secondary: var(--rh-color-text-secondary-on-light, #4d4d4d);
+  --_disabled-background-color: var(--rh-color-surface-light, #e0e0e0);
+  --_border-color: var(--rh-color-border-subtle-on-light, #c7c7c7);
 
-  color: var(--rh-color-text-primary-on-light);
+  color: var(--rh-color-text-primary-on-light, #151515);
 }
 
-:host([color-palette='lightest']) { background-color: var(--rh-color-surface-lightest); }
-:host([color-palette='lighter']) { background-color: var(--rh-color-surface-lighter); }
-:host([color-palette='light']) { background-color: var(--rh-color-surface-light); }
+:host([color-palette='lightest']) { background-color: var(--rh-color-surface-lightest, #ffffff); }
+:host([color-palette='lighter']) { background-color: var(--rh-color-surface-lighter, #f2f2f2); }
+:host([color-palette='light']) { background-color: var(--rh-color-surface-light, #e0e0e0); }


elements/rh-select/demo/option-descriptions.html
```diff
@@ -0,0 +1,55 @@
+<div class="container limit-width">
+  <label for="select-3">
+    With option descriptions
+  </label>
+  <rh-select id="select-3" placeholder="Select an item">
+    <rh-option>
+      Item 1
+      <span slot="description">Description goes here</span>
```
bennypowers: I noticed these are all plain text, which indicates that these should be slot/attribute pairs; the slot being preferred for rich content and the attribute being preferred for plain text.

elements/rh-select/demo/option-icons.html
```diff
@@ -0,0 +1,55 @@
+<div class="container limit-width">
+  <label for="select-2">
+    With option icons
+  </label>
+  <rh-select id="select-2" placeholder="Select an item">
+    <rh-option>
+      <rh-icon slot="icon" set="ui" icon="document"></rh-icon>
```
bennypowers: Likewise, perhaps we add icon and icon set attributes to rh-option.

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,102 @@
...
+  #internals = InternalsController.of(this, { role: 'group' });
+
+  #optionEls: RhOption[] = [];
+
+  override firstUpdated() {
+    this.#optionEls = this.#getChildOptions();
+    this.#disableChildren();
+  }
+
+  override updated(changedProperties: PropertyValues<this>) {
```
bennypowers: Consider replacing this with an observes decorator.

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,102 @@
...
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  @query('slot:not([name])') private defaultSlot!: HTMLSlotElement;
+
+  // eslint-disable-next-line no-unused-private-class-members
+  #internals = InternalsController.of(this, { role: 'group' });
+
+  #optionEls: RhOption[] = [];
+
+  override firstUpdated() {
```
bennypowers, line 49: I suspect this will have to be adjusted for SSR scenarios.

elements/rh-select/rh-option.ts
```diff
@@ -0,0 +1,120 @@
...
+  /** Whether option is selected */
+  @property({ type: Boolean, reflect: true }) selected = false;
+
+  /** Optional option description; overridden by description slot. */
+  @property() description = '';
+
+  /**
+   * This option's position relative to the other options
+   */
+  set posInSet(posInSet: number | null) {
```
bennypowers: We might consider making this private. How would that work? The select element would provide a context. That context would be an object containing the list of options, and each option would privately set its internal pos-in-set by consulting the options list.

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,330 @@
...
+  /** Whether the select is disabled */
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  /** Whether the select listbox is expanded */
+  @property({ type: Boolean, reflect: true }) expanded = false;
+
+  /**
+   * Enable to flip listbox when it reaches boundary
+   */
+  @property({ attribute: 'enable-flip', type: Boolean }) enableFlip = false;
```
bennypowers: Is this something that we want to expose? Maybe it should just always be true.

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,330 @@
...
+    const { selected } = this.#combobox;
+    return (selected ? this.value : '')
+      || this.#computePlaceholderText()
+      || 'Select a value';
+  }
+
+  override render() {
+    const { disabled, expanded, placeholder } = this;
+    const { anchor = 'bottom', alignment = 'start', styles = {} } = this.#float;
+    const { height, width } = this.getBoundingClientRect?.() || {};
```
bennypowers: This is not going to work in SSR. It's going to return an empty object, which means height and width will be undefined.

elements/rh-select/rh-select.ts
adamjohnson: Didn't seem to be working here or in PFE so I removed it in [12fb174](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/12fb174ab6ac0acc1086135b6861938fc0726c6d).

elements/rh-select/demo/option-icons.html
adamjohnson: Handled in [5a61b70](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/5a61b703982ccedcb7b68f24ad97da2c66ae2e84).

elements/rh-select/demo/option-descriptions.html
adamjohnson: Handled in [b3e5f95](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/b3e5f9516113bf443c75644eabb69d2361cf2c04).

elements/rh-select/rh-option-group.ts
adamjohnson: Handled in [2b45f3d](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/2b45f3d0f5124b44744c439a708de171689454e2).

elements/rh-select/rh-option-group.ts
adamjohnson: Handled in [f936e8c](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/f936e8cd03121366f795dbd6076d95675d0923e1).

elements/rh-select/rh-select.ts
adamjohnson: Fixed in [dbd99ad](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/dbd99ad979afe5b7abd8dfaca6bc0723d52d00c7). Good call!

elements/rh-select/rh-option.ts
adamjohnson: I looked into this. The listbox controller sets `aria-setsize` and `aria-posinset` on [lines 202-203](https://github.com/patternfly/patternfly-elements/blob/main/core/pfe-core/controllers/listbox-controller.ts#L202-L203). At this time, there doesn't appear to be an option to disable this functionality. 

Alternatively, we could use context and elementInternals while manually removing the attributes on `updated()`, but that seems heavy handed.

Thoughts? Thanks for the feedback.

marionnegp (changes requested):
- [ ] Is there a way to hide the slot for the help text when it's not being used? It's adding extra space at the bottom. 
Image

- [ ] I'm not sure if this is intentional, but I'm seeing rh-select break out of the container with a max width. 
Image

This scenario _should_ be unlikely, but I've added this to the Figma mockup just in case: 
<img width="365" height="140" alt="Image" src="https://github.com/user-attachments/assets/a3fb7aa3-7ea5-47ed-9e54-19751ffecdc4" /

- [ ] Can the `font-family` be set to `--rh-font-family-body-text` by default? It's showing Arial now.

- [ ] For the status/state icons in the select list and help text, can they be 14px? It'll be the size that the unified version would use because PF likes to tie their icon sizes to their font sizes, which makes sense to me. Not sure what your thoughts are though, since we don't have a token for that yet.

- Heads up that the help text size might change to `--rh-font-size-body-text-xs` based on unified design system specs. If it changes too late, maybe we can add that to a theming issue instead?

Nice job!!

elements/rh-select/demo/color-context.html
```diff
@@ -0,0 +1,44 @@
+<rh-context-demo>
+  <div class="limit-width">
+    <label for="select-1">
+      Label
+    </label>
+    <rh-select id="select-1" placeholder="Select an item">
+      <rh-option>Ansible Lightspeed</rh-option>
+      <rh-option>
+        <rh-icon slot="icon" set="ui" icon="document"></rh-icon>
```
bennypowers: Let's use `<rh-option icon icon-set>` in most demos, and add a separate demo for slot overrides

The general rule for attribute/slot pairs: prefer to use the attribute in most cases (plain text) and use the slot when you have rich or custom content.

elements/rh-select/demo/index.html
```diff
@@ -0,0 +1,84 @@
+<div class="container limit-width">
```
bennypowers: Try to avoid wrapper divs in demos, they add noise

elements/rh-select/demo/index.html
```diff
@@ -0,0 +1,84 @@
...
+  <label for="select-1">
+    Red Hat Products <span class="required" aria-label="Required field">*</span>
+    <rh-tooltip>
```
bennypowers: move this pattern to a separate demo. Index should be a simple example of basic usage

bennypowers: The snippet from the changeset is a good example

elements/rh-select/demo/option-descriptions.html
```diff
@@ -0,0 +1,39 @@
...
+  <label for="select-3">
+    With option descriptions
+  </label>
+  <rh-select id="select-3" placeholder="Select a country">
+    <rh-option label="Canada" value="ca" description="Longest coastline of any country"></rh-option>
+    <rh-option label="Germany" value="de" description="Worlds third largest importer and exporter"></rh-option>
+    <hr>
+    <rh-option label="India" value="in" description="Seventh largest country by area"></rh-option>
+    <rh-option>
+      Israel
```
bennypowers: כל עוד בלבב, פנימה...

elements/rh-select/demo/states.html
```diff
@@ -0,0 +1,67 @@
+<div class="container limit-width">
+  <label for="select-5">
+    Success
+  </label>
+  <rh-select aria-describedby="help-text-5" id="select-5" placeholder="Select an item" state="success">
+    <rh-option>3scale API Management</rh-option>
+    <rh-option>Ansible Automation Platform</rh-option>
+    <rh-option>Ansible Lightspeed</rh-option>
+    <p id="help-text-5" slot="help-text">Sucess message</p>
```
bennypowers: If the design for this permits, we should consider making this an attr/slot pair, and that should allow us to remove the requirement for users to add `aria-describedby`

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,104 @@
...
+
+import { InternalsController } from '@patternfly/pfe-core/controllers/internals-controller.js';
+import { observes } from '@patternfly/pfe-core/decorators/observes.js';
+
+import { RhOption } from './rh-option.js';
+
+import styles from './rh-option-group.css';
+
+/**
+ * Group of options within a select / listbox
```
bennypowers: Add some rfc2119 guidance here 

Use when

Avoid if

Etc

elements/rh-select/rh-option.ts
bennypowers: what would we need to change in listbox in order to enable our desired behaviour?

Here's what we have now:
`ariaSetSize and `ariaPosInSet are set on each item by the items setter in ListboxController:

```ts
  /**
   * register's the host's Item elements as listbox controller items
   * sets aria-setsize and aria-posinset on items
   * @param items items
   */
  set items(items: Item[]) {
    this.#items = items;
    this.#items.forEach((item, index, _items) => {
      item.ariaSetSize = _items.length.toString();
      item.ariaPosInSet = (index + 1).toString();
    });
  }
```
This is not the only place in the controller where aria attrs are set as a side effect of some state change:

1. the default `setItemSelected` callback applies `aria-selected=true/false` to each item in the `selected` setter
2. `aria-disabled` and `aria-multiselectable` are set in the `hostUpdated` callback

I think we could either:
1. add a third and fourth parameter to `setItemSelected`, making the signature `setItemSelected(item: Item, selected: boolean, index: number; items: Item[])`
2. add a new option `setItems(items: Items[])`, which would default to the `forEach` in the snippet above, with documented caveats for callers that override, like we already have with `setItemSelected`

elements/rh-select/docs/00-overview.md
bennypowers, line 1: needs docs, primarily in jsdoc

elements/rh-select/docs/data.yaml
```diff
@@ -0,0 +1,16 @@
+$schema: ../../../docs/schemas/element-data.schema.json
+tagName: rh-select
+name: Select
+type: element
+overallStatus: ready
+description: >-
+  The select element represents a control that provides a menu of options.
```
bennypowers:
```suggestion
```
belongs inline

bennypowers: but this would be a good
```js
/**
 * @summary a control that provides a menu of options.
 */
```

elements/rh-select/docs/screenshot.png
bennypowers, line 1: is this up to date?

elements/rh-select/README.md
bennypowers, line 1: update prose in this file

elements/rh-select/rh-option-group.ts
bennypowers: same for public class fields

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,104 @@
...
+   */
+  #getChildOptions(): RhOption[] {
+    if (!this.defaultSlot) {
+      return [];
+    }
+
+    const assignedElements = this.defaultSlot.assignedElements();
+    return assignedElements.filter((element): element is RhOption =>
+      element instanceof RhOption
+    );
```
bennypowers:
```suggestion
    const assignedElements = this.defaultSlot?.assignedElements?.() ?? [];
    return assignedElements.filter((element): element is RhOption =>
      element instanceof RhOption
    );
```
save a few bytes

elements/rh-select/rh-option.css
```diff
@@ -0,0 +1,77 @@
+:host {
+  /* Remove this re-declaration when the `status-disabled` variables are defined in tokens */
+  --rh-color-text-status-disabled:
```
bennypowers: can you make this a fallback instead?

elements/rh-select/rh-option.ts
```diff
@@ -0,0 +1,153 @@
...
+import { ifDefined } from 'lit-html/directives/if-defined.js';
+
+import { InternalsController } from '@patternfly/pfe-core/controllers/internals-controller.js';
+
+import { observes } from '@patternfly/pfe-core/decorators/observes.js';
+
+import styles from './rh-option.css';
+
+/**
+ * Option within a select / listbox
```
bennypowers: more jsdoc please. see https://bennypowers.dev/cem/docs/usage/effective-mcp-descriptions/

elements/rh-select/rh-option.ts
```diff
@@ -0,0 +1,153 @@
...
+  @property({ type: Boolean, reflect: true }) disabled = false;
+
+  /** Form value for this option */
+  @property()
+  get value() {
+    if (this.#value) {
+      return this.#value;
+    }
+    // Fall back to display label
+    return this.displayLabel;
```
bennypowers: reverse this: don't ship the displayLabel as the value, instead have the displayLabel fall back to the value

elements/rh-select/rh-option.ts
```diff
@@ -0,0 +1,153 @@
...
+  set value(v: string) {
+    this.#value = v;
+  }
+
+  /**
+   * Gets the display text for the rh-option.
+   * Priority: slotted text content -> label attr -> value attr -> ''
+   */
+  get displayLabel(): string {
+    if (!isServer) {
```
bennypowers: this is a lot to do in a getter. imagine this property being accessed in a hot loop. setup up a slotchange listener or MO instead

elements/rh-select/rh-select.css
```diff
@@ -0,0 +1,211 @@
+:host {
+  /* Remove these re-declarations when the `status-disabled` variables are defined in tokens */
+  --rh-color-text-status-disabled:
```
bennypowers: can you make these fallbacks instead?

elements/rh-select/docs/00-overview.md
bennypowers: likewise next 3

patches/@patternfly+pfe-core+5.0.3.patch
bennypowers, line 1: Don't merge until this patch is resolved upstream (please keep this comment unresolved until then)

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,590 @@
...
+import { SlotController } from '@patternfly/pfe-core/controllers/slot-controller.js';
+import { InternalsController } from '@patternfly/pfe-core/controllers/internals-controller.js';
+import { FloatingDOMController } from '@patternfly/pfe-core/controllers/floating-dom-controller.js';
+
+import { arraysAreEquivalent } from '@patternfly/pfe-core/functions/arraysAreEquivalent.js';
+import { observes } from '@patternfly/pfe-core/decorators/observes.js';
+
+import { RhOption } from './rh-option.js';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { RhOptionGroup } from './rh-option-group.js';
```
bennypowers:
```suggestion
import './rh-option-group.js';
```

maybe we should leave this out, and have users who need it explicitly import it

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,590 @@
...
+
+import '@rhds/elements/rh-icon/rh-icon.js';
+
+import styles from './rh-select.css';
+
+export class RhSelectChangeEvent extends Event {
```
bennypowers, line 34: jsdoc

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,590 @@
...
+    super('change', { bubbles: true });
+  }
+}
+
+/**
+ * A select list enables users to select one or more items from a list.
+ *
+ * @summary enables users to select one or more items from a list
+ *
+ * A select component consists of a toggle control to open and close a menu of actions or links.
```
bennypowers:
```suggestion
 * @summary enables users to select one or more items from a list
 *
 * A select component consists of a toggle control to open and close a menu of actions or links.
```

bennypowers: add rfc2119 words, same for all public apis see guides above.

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,590 @@
...
+  get options(): RhOption[] {
+    if (isServer) {
+      return []; // TODO: expose a DOM property to allow setting options in SSR scenarios
+    } else {
+      return [
+        this._placeholder,
+        ...Array.from(this.querySelectorAll('rh-option')),
+      ].filter((x): x is RhOption => !!x && !x.hidden);
+    }
+  }
```
bennypowers, line 146:
```suggestion
  #options: RhOption[] = [];
  
  /** List of options */
  get options(): RhOption[] {
    return this.#options;
  }
  
  set options(v: RhOption[]) {
    this.#options = [];
    this.requestUpdate('options', null);
  }
```

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,590 @@
...
+        .join(', ');
+
+    return selectedLabels
+      || this.#computePlaceholderText()
+      || 'Select a value';
+  }
+
+  override connectedCallback() {
+    super.connectedCallback();
+    this.addEventListener('keydown', this.#captureKeydown, { capture: true });
```
bennypowers, line 249:
```suggestion
    this.addEventListener('keydown', this.#captureKeydown, { capture: true });
    if (!isServer) {
      this.#options = [
        this._placeholder,
        ...Array.from(this.querySelectorAll('rh-option')),
      ].filter((x): x is RhOption => !!x && !x.hidden);
    }
```

do this also in a mutation observer or slotchange

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,590 @@
...
+            <rh-option id="placeholder"
+                       disabled
+                       ?inert="${placeholderIsInert || hasSelection}"
+                       ?hidden="${!placeholder && this.#slots.isEmpty('placeholder')}">
+              <!-- placeholder text for the select. Overrides the \`placeholder\` attribute. -->
+              <slot name="placeholder">${placeholder ?? ''}</slot>
+            </rh-option>
+            ${this.#combobox.renderItemsToShadowRoot()}
+            <!-- insert \`rh-option\` and/or \`rh-option-groups\` here -->
+            <slot></slot>
```
bennypowers:
```suggestion
            <slot @slotchange="${this.#getOptions}"></slot>
```

elements/rh-select/demo/states.html
bennypowers: actually we already have a slot, which means we can set internals.ariaDescribedbyElements (if supported) or internals.ariaDescription from the slot's flattened text node content

bennypowers (changes requested):

elements/rh-select/docs/screenshot.png
adamjohnson: Nope. We need a proper `screenshot.png` cc @marionnegp.

elements/rh-select/README.md
adamjohnson: Handled in [383fa58](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/383fa584c8bfdd34dca0683e29882b0d9b45d466).

elements/rh-select/docs/data.yaml
adamjohnson: Handled in [64db901](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/64db9012a6a5f23b06b38ea392cf1193c8df4284). Updated the summary in [9fc5657](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/9fc565796889580dff57a5685d49d5945d6a0fdc).

elements/rh-select/rh-option.ts
adamjohnson: Added in [76566b8](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/76566b849f6f8586098b874722d16413590d0963).

elements/rh-select/rh-option-group.ts
adamjohnson: Optimized in [4e99ff5](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/4e99ff59fe4bf03d37be747b957d16464ea7da24).

elements/rh-select/demo/index.html
adamjohnson: Fixed in [647eddc](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/647eddc6c8b7758fcb108a30ab0c9a29f3a72589).

elements/rh-select/demo/index.html
adamjohnson: Fixed in [fdd7762](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/fdd77620190143c855176b6cb2132022240c2360).

elements/rh-select/rh-option-group.ts
adamjohnson: Handled in [76566b8](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/76566b849f6f8586098b874722d16413590d0963).

elements/rh-select/rh-option.css
adamjohnson: I took a stab at what you mean by this in [b53c1db](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/b53c1dbced84700e0c0dc822a5fd066f4a94b296).

Is this what you meant? If so, I will apply it to rh-select.css. If not, kindly provide an example.

elements/rh-select/rh-option.ts
adamjohnson: I implemented this in [d2b8a92](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/d2b8a922d94bbcba17ccf57a21c3566321125a1d); however...

If a user has an option:

```html
<option value="openshift">Red Hat OpenShift</option>
```

...wouldn't we want `option.value` to return `openshift` instead of `Red Hat OpenShift`? Or should we recommend to implementors to only use `value` OR slotted text (never both)?

Reversing them (like how it's currently implemented) returns the displayLabel as the value.

NB (mostly for me): tests would need updated if we keep the implementation this way

elements/rh-select/rh-option.ts
adamjohnson: Handled in [5e72a1b](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/5e72a1ba6e30376de89206acad7f6e8910942640).

elements/rh-select/rh-select.css
adamjohnson: See my question at https://github.com/RedHat-UX/red-hat-design-system/pull/2802#discussion_r2722752698. Once that's clarified, we can tackle these in this file.

elements/rh-select/rh-select.ts
adamjohnson: `@fires` is listed below. Should those docs move up to this line? See [`@fires` docs](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/files#diff-e2671ccd7c80793550fcb507d98b2021a4d4d2a6abc319c1c27d416d96f97fc8R39).

elements/rh-select/rh-select.ts
adamjohnson: Handled in [a06d592](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/a06d592c0c447383b1b33f0a2ddc0d5abdae3e83).

elements/rh-select/rh-select.ts
adamjohnson: Removed in [0b16691](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/0b16691d0af25909cc74783ffbe6309b0fb8d677).

elements/rh-select/rh-option.css
bennypowers: yes that's good. you could also make it:

```css
  /* Nix this fallback `color` declaration when the `status-disabled` vars are defined in tokens */
  color: var(--rh-color-text-status-disabled,
    light-dark(var(--rh-color-gray-50, #707070),
      var(--rh-color-gray-60, #4d4d4d))
  );
```

but i think yours is better, since the token names aren't shipped yet, so we shouldn't squat on the --rh- prefix

elements/rh-select/rh-option.ts
bennypowers: that's correct

```html
<rh-option id="a" value="openshift"></rh-option>
<rh-option id="b">Red Hat OpenShift</rh-option>
<rh-option id="c" value="openshift">Red Hat OpenShift</rh-option>
<rh-option id="d" value="openshift" display-label="Display Label">Red Hat OpenShift</rh-option>
```
| id  | computed `value` | computed `displayValue` |
| --- | ---------------- | ----------------------- |
| a | 'openshift' | 'openshift' |
| b | 'Red Hat OpenShift' | 'Red Hat OpenShift' |
| c | 'openshift' | 'Red Hat OpenShift' |
| d | 'openshift' | 'Display Label' |

elements/rh-select/rh-select.ts
bennypowers: the `RhSelectChangeEvent` [class declaration](https://github.com/webcomponents/custom-elements-manifest/blob/d2f79f0d22c4d48a68628cf867c2319576ffc5d2/schema.d.ts#L483-L555) itself needs jsdoc, independent of the fact that the `change` event is documented on the rh-select CustomElementDeclaration

elements/rh-select/rh-select.ts
adamjohnson: Handled in [97573f1](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/97573f175ff78a3111dcd73d4f15088a6d383339).

elements/rh-select/rh-select.ts
adamjohnson: Handled in [97573f1](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/97573f175ff78a3111dcd73d4f15088a6d383339).

elements/rh-select/rh-select.ts
adamjohnson: Handled in [97573f1](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/97573f175ff78a3111dcd73d4f15088a6d383339).

elements/rh-select/demo/color-context.html
adamjohnson: Handled in [388ac5a](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/388ac5a06e00831d5a413a176372fbd9cdb6a6a8).

elements/rh-select/rh-select.ts
adamjohnson: Handled in [eb9171e](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/eb9171e1bb933161f45842db35f74c53508e406e).

elements/rh-select/rh-option.css
adamjohnson: Had to use your formatting actually. 

tl;dr Using `color` 2x caused the 2nd color to override the first. Since `rh-color-text-status-disabled` is currently undefined, it would be set to no color/transparent and override the previous `color` property.

Fixed in [a1f4145](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/a1f4145f4141c2eec0707e72a5a0f16831a4a1c4).

elements/rh-select/rh-select.css
adamjohnson: Fixed in [a1f4145](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/a1f4145f4141c2eec0707e72a5a0f16831a4a1c4). See comment in other thread linked above for further details.

elements/rh-select/rh-option.ts
adamjohnson: Fixed in [6e1234e](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/6e1234e485a29ad7b5fe58ababb426feefde49c6).

elements/rh-select/demo/states.html
adamjohnson: Made this a attr/slot pair in [2ec42a1](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/2ec42a191e82242eca622e73906aa3c39ee9bfaf). Now, users don't need to muck around with aria-describedby. It's handled automatically. Great suggestion!

  * If a user adds `<rh-select help-text="Hello">`, the internals controller sets `ariaDescription` to "Hello".
  * If a user slots in help text (eg: `<p slot="help-text">Hi</p>`), the element is auto-assigned an ID and `ariaDescribedByElements` is set via the internals controller.
  * If for some strange reason they set both the attr _and_ slot in content, the slotted content "wins".
  
  PTAL. Thanks again for the suggestion.

elements/rh-select/rh-option.ts
adamjohnson: Going to explore option two next week.

elements/rh-select/demo/states.html
bennypowers: > If a user slots in help text (eg: `<p slot="help-text">Hi</p>`), the element is auto-assigned an ID and ariaDescribedByElements is set via the internals controller.

we don't need to add IDs unless ariaDescribedByElements is unsupported - you can just set it to the list of elements

as it turns out, [ariaDescribedByElements is baseline 2025](https://github.com/patternfly/patternfly-elements/blob/f43c1bce9596fab50ed2cdf00835f75ee96741b9/core/pfe-core/controllers/slot-controller.ts#L216) so i think we can fire and forget

```ts
this.#internals.ariaDescribedByElements = this.#slots.getSlotted('help-text');
```

This should be enough, really. with the caveat that slot controller only goes [one level deep](https://github.com/patternfly/patternfly-elements/blob/f43c1bce9596fab50ed2cdf00835f75ee96741b9/core/pfe-core/controllers/slot-controller.ts#L216), so if we're threading content through multiple layers of slots we might not get the elements. this might not be a problem for AT though

adamjohnson: Implemented in [d209c0d](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/d209c0dcbda7876dc19b6946a77003afbcec44aa). 

This does make my spidey sense tingle a little bit (getting rid of the compat guards), but hey, it's Baseline. Trust the process!

elements/rh-select/rh-option.ts
adamjohnson: Implemented option number two in [611d885](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/611d88505f156cd5529c3fd45b35a1b90bb55a4e). No more visible `aria-*` attributes.

I will apply the patches made here to https://github.com/patternfly/patternfly-elements/pull/2963.

bennypowers: Let's consider the setup *prior to* 611d88505f156cd5529c3fd45b35a1b90bb55a4e

1. pf-select has a ComboboxController
2. ComboboxController has a ListboxController
3. ComboboxController's `items` setter forwards to LBC's items setter
4. LBC's `items` setter is where `aria-posinset` and `aria-setsize` are set

Step 4 there is where the problem lives: it's where control inverts from the element to the controller.

In this commit, we have a couple of items which raise interest:
1. rh-option exposes `posInSet` and `setSize`, which raises the question "Why would a downstream user ever have a need to manually set these?"
2. rh-select needs to pass a no-op `setItems` option, which hints towards an inversion of control

let's consider an alternative. pf-select, pf-option, rh-select, and rh-option all already use InternalsController. What if InternalsController provided the means to manage this state?

```ts
// pseudocode
export class InternalsController {
  static setAriaPosInSet(listboxItem: ReactiveControllerHost, value: number | null): void {
    const instance = this.instances.get(listboxItem);
    if (!instance) {
      if (value == null)
        listboxItem.removeAttribute('aria-posinset');
      else
        listboxItem.setAttribute('aria-posinset', value);
    } else {
      instance.ariaPosInSet = value;
    } 
  }

  static setAriaSetSize(listboxItem: ReactiveControllerHost, value: number | null): void {
    // etc...
  }
}
```

ListboxController can now do updates in hostUpdate

```ts
// pseudocode
export class ListboxController {
  set items(i) {
    this.#items = i; // no more side effects in setter
  }

  hostUpdate() {
    this.#items.forEach((item, index, items) => {
      InternalsController.setAriaPosInSet(item, index + 1);
      InternalsController.setAriaSetSize(item, items.length);
    });
  }
}
```

PROs:
- option elements would not need to do anything
- works regardless of item type - falls back to aria attrs

CONs:
- adds InternalsController to the module graph of ListboxController (but why wouldn't it be there otherwise?)

What do you think about that?

hellogreg (changes requested):
Keyboard nav of the options works well until an option has been selected and the select is re-opened. At that point, two things happen:

1. When trying to arrow up, the "Select an item" option can no longer be selected, so arrowing up repeatedly will not roll back over to the bottom.
2. When arrowing down at the bottom of the list, you have to arrow down twice to roll back over to the top option.

You can see these behaviors in the attached video.

https://github.com/user-attachments/assets/d6e855b7-fc8f-4de2-b988-dce9f9c5a8cb

elements/rh-select/rh-option.ts
adamjohnson: This makes great sense. See: [af1eca5](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/af1eca51d5c8fecd7fc1ac56d7c5c27bf2e958a7) [b81df78](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/b81df78748fd2d9eb6c05f42dadf3103755ca287) and [17e43f1](https://github.com/patternfly/patternfly-elements/pull/2963/commits/17e43f1518ba7bfa07ad239d3818efeb0748a35b) [cdf6ea1](https://github.com/patternfly/patternfly-elements/pull/2963/commits/cdf6ea113289c37573330a8f4abf77acc2a364ed) on the PFE side. PTAL.

.changeset/optionally-better-select.md
zeroedin, line 1: the changeset file name I hope isn't foreshadowing 😆

adamjohnson: Oh god 😨  

Can we manually change that? If so, I'ma do that. I am superstitious.

zeroedin: yes you can rename the file

zeroedin: but let it be known it was previously known as `rare-bugs-follow.md` 😁

adamjohnson: [076c986](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/076c986387d2afafbb24aa21a3bd470a8ac3efc3) Rare bugs be gone!

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,865 @@
...
+ * include a `<label>` or `accessible-label` so screen readers can announce
+ * the ARIA combobox role and current selection.
+ * @summary A control that provides a menu of options
+ * @alias select
+ * @fires {Event} open - Fires when the dropdown listbox opens. Does not bubble. The event has no `detail` payload.
+ * @fires {Event} close - Fires when the dropdown listbox closes. Does not bubble. The event has no `detail` payload.
+ * @fires {RhSelectChangeEvent} change - Fires when the selected value changes. Bubbles. The event has no `detail` payload; read the new value from the `value` property.
+ */
+@customElement('rh-select')
+export class RhSelect extends LitElement {
```
zeroedin, line 50: Should likely have the `@themeable`decorator

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,865 @@
...
+
+  override disconnectedCallback() {
+    super.disconnectedCallback();
+    this.removeEventListener('keydown', this.#captureKeydown, { capture: true });
+  }
+
+  protected formResetCallback() {
+    this.value = '';
+    this.selected = [];
+  }
```
zeroedin, line 369: Should this instead of clearing the selection return to the initial selection if one was set like a native `<select>`?

elements/rh-select/rh-select.css
```diff
@@ -0,0 +1,260 @@
...
+  background-color: var(--rh-color-border-subtle);
+  block-size: 1px;
+  display: block;
+  inline-size: 100%;
+}
+
+::slotted(rh-option-group + hr),
+::slotted(hr:has(+ rh-option-group)) {
+  display: none;
+}
```
zeroedin: These are invalid selectors.  

::slotted can not take combinators (+, >, ~, space), pseudo-classes that depend on relationships (:has(), :nth-child() relative to siblings outside slot) or complex selectors

elements/rh-select/rh-select.css
```diff
@@ -0,0 +1,260 @@
...
+
+  /** Medium spacing design token for vertical padding */
+  padding-block: var(--rh-space-md, 8px);
+  position: absolute;
+  opacity: 0;
+}
+
+#outer.expanded #listbox-container {
+  opacity: 1;
+  z-index: 9999;
```
zeroedin: can we use a lower z-index?  given the stacking context?  If we are afraid of stuck behind bugs, let's make this z-index configurable via a css var.

elements/rh-select/rh-select.css
```diff
@@ -0,0 +1,260 @@
...
+
+  /** Body text font stack for the toggle button */
+  font-family: var(--rh-font-family-body-text, RedHatText, 'Red Hat Text', Helvetica, Arial, sans-serif);
+
+  /** Small body text size for the toggle button */
+  font-size: var(--rh-font-size-body-text-sm, 0.875rem);
+
+  /** Large spacing between toggle text and icons */
+  gap: var(--rh-space-lg, 16px);
+  min-block-size: 37px;
```
zeroedin, line 130: This seems like a magic number.  Is there no combination of padding + text that works in place of a hard coded value?

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,114 @@
...
+  @query('slot:not([name])') private defaultSlot!: HTMLSlotElement;
+
+  // eslint-disable-next-line no-unused-private-class-members
+  #internals = InternalsController.of(this, { role: 'group' });
+
+  #optionEls: RhOption[] = [];
+
+  override firstUpdated() {
+    if (!isServer) {
+      this.#optionEls = this.#getChildOptions();
```
zeroedin: What happens if child options are added dynamically?  #getChildOptions only ever runs in firstUpdated() which wouldn't run again if a option was injected in via JS.  I would suggest a slotchange event on the slot to also run #getChildOptions() to recalculate.

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,114 @@
...
+   */
+  #updateDisabledChildren(): void {
+    for (const childOption of this.#optionEls) {
+      if (this.disabled) {
+        childOption.disabled = true;
+      } else {
+        childOption.disabled = false;
+      }
+    }
+  }
```
zeroedin, line 104: You could likely reduce the overhead by combining these methods, removing the need for #disableChildren();

```
  override firstUpdated() {
    if (!isServer) {
      this.#optionEls = this.#getChildOptions();
      if (this.disabled) {
        this.#updateDisabledChildren();
      }
    }
  }
```

then just setting the childOption.disabled to the value of this.disabled instead of having the addition if/else

```
  /**
   * Updates disabled state of child options when rh-option-group disabled state changes
   */
  #updateDisabledChildren(): void {
    for (const childOption of this.#optionEls) {
      childOption.disabled = this.disabled;
    }
  }
```

elements/rh-select/test/rh-select.e2e.ts
```diff
@@ -0,0 +1,25 @@
...
+    await componentPage.snapshot();
+  });
+
+  test('ssr', async ({ browser }) => {
+    const fixture = new SSRPage({
+      tagName,
+      browser,
+      demoDir: new URL('../demo/', import.meta.url),
+      importSpecifiers: [
+        `@patternfly/elements/${tagName}/${tagName}.js`,
```
zeroedin:
```Suggestion
        `@rhds/elements/${tagName}/${tagName}.js`,
```

elements/rh-select/rh-select.css
zeroedin: Ironically came across this article posted 2 days ago mear hours later.

https://css-tricks.com/the-value-of-z-index/

bennypowers: we have a few other z indexes in the repo - how will this interact with nav1/2, etc?

elements/rh-select/rh-select.css
adamjohnson: `37px` is right out of the [design spec](https://www.figma.com/design/BOGsHyL2K07gtPKIXppbHQ/rh-select?node-id=4002-413&t=3CDiYZoH4dsogtjZ-1). Since it's `min-block-size` and not just `block-size`, I think this is fine. There is no one or two token values that easily add up to `37px`.

marionnegp: This might have to do with how line heights and font sizes are calculated in Figma, but this does give us a good standard height for form fields. It's also only 1px larger than a button's height, which is helpful when a select and a menu dropdown have to be used next to each other.

elements/rh-select/rh-select.ts
adamjohnson: 👉 [fa3a9bf](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/fa3a9bfed72f9d38d8c1abfa82abd4984b278023)

elements/rh-select/rh-select.css
adamjohnson: Made it `1000` to match `<rh-menu-dropdown>`. LMK if we wanna add a public variable here for customization.

[d08ef7d](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/d08ef7d3e3f4db292ce738bfc191922acce68e23)

elements/rh-select/rh-select.css
adamjohnson: 👉 [b3c6b79](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/b3c6b79c7d691349e5babf3886a4ba820d34c996)

elements/rh-select/rh-select.ts
adamjohnson: Fixed in [6dc0ee1](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/6dc0ee12dd901c98ac4da7e424fabcaa84ab5099).

elements/rh-select/rh-option-group.ts
adamjohnson: Spent some time on this a while back to ensure it works (it does without changes). Added a demo for it: [27f17b5](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/27f17b5b6553acbe696523a8539e5ba2983381ec). If you see things that need improvement, I'm all ears. 

(I'm going to "Resolve conversation". If you suggest further changes, hit unresolve).

elements/rh-select/rh-option-group.ts
adamjohnson: Great suggestion! Implemented in [21d9cd6](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/21d9cd65f81fceb4fa48e69dfdca10c25b8ea78b). TY.

patches/@patternfly+pfe-core+5.0.3.patch
bennypowers: has this been handled?

elements/rh-select/rh-option.ts
```diff
@@ -0,0 +1,166 @@
...
+   * Priority: value attr -> displayLabel -> ''
+   */
+  @property()
+  get value() {
+    return this.#value ?? this.displayLabel ?? '';
+  }
+
+  set value(v: string) {
+    this.#value = v;
+  }
```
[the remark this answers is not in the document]
bennypowers: verify this finding - doesn't the `@property` decorator handle that for us?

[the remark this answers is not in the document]
bennypowers: file references are to the lit/lit repository

# Lit `@property` Decorator on Accessor Pairs

**Yes, Lit automatically calls `requestUpdate()` for you**, even when you define your own accessor pair. The user does **not** need to call it manually.

## Legacy decorators (current TypeScript default)

1. `createProperty()` at `reactive-element.ts:709-711` detects the existing accessor via `this.prototype.hasOwnProperty(name)` and sets `options.wrapped = true`.

2. `getPropertyDescriptor()` at `reactive-element.ts:759-766` reads the user's getter/setter from the prototype.

3. It returns a **new wrapping descriptor** at `reactive-element.ts:784-793` that:
   - Captures the old value via the user's getter (line 787)
   - Calls the user's setter (line 788)
   - Calls `this.requestUpdate(name, oldValue, options)` (line 789)

## Standard decorators (TC39)

In `decorators/property.ts:162-168`, when `kind === 'setter'`, it returns a wrapper function that:
- Captures the old value (line 165)
- Calls the original setter (line 166)
- Calls `this.requestUpdate(name, oldValue, options, true, value)` (line 167)

## Proof via test

`packages/reactive-element/src/test/decorators/property_test.ts:125-193` -- the test "can decorate user accessor with @property" demonstrates this concretely:

- A class defines `get foo()` / `@property() set foo(v)` with a manual backing field `_foo` (lines 131-138).
- The setter body is just `this._foo = v` -- no `requestUpdate()` call.
- After `el.foo = 5`, the test asserts `el.updatedProperties!.has('foo')` is `true` (line 174), confirming `requestUpdate` was called automatically.
- Old values are tracked correctly too: after setting `foo` to 7, `updatedProperties.get('foo')` equals the previous value 5 (line 188).

## Bottom line

Lit wraps user-defined setters. You get `requestUpdate()`, change tracking, and attribute reflection for free.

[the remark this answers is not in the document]
bennypowers: so this is a WONTFIX: working as intended.
@adamjohnson you might want to add a quick unit test for this, to keep the bots happy

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,102 @@
...
+  #optionEls: RhOption[] = [];
+
+  override firstUpdated() {
+    if (!isServer) {
+      this.#optionEls = this.#getChildOptions();
+      if (this.disabled) {
+        this.#updateDisabledChildren();
+      }
+    }
+  }
```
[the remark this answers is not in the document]
bennypowers: verify that we don't have a slotchange listener or a mutationobserver - even one implemented in a controller (like combobox controller)

elements/rh-select/rh-option-group.ts
```diff
@@ -0,0 +1,102 @@
...
+  render() {
+    return html`
+      <div id="label-container"
+           role="presentation">
+        <!-- Group label as inline text. Overrides the \`label\` attribute. Screen readers announce this text when the group receives focus. -->
+        <slot name="label">${this.label}</slot>
+      </div>
+      <!-- Insert \`<rh-option>\` elements. Each option must have accessible text content for screen readers. -->
+      <slot></slot>
+    `;
```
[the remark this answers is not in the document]
adamjohnson: `aria-label`/`aria-labelledby` is not a required attribute for `role="group"`. We can omit this extra code. If it turns into an issue later, we can revisit.

elements/rh-select/rh-option-group.ts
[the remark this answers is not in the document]
adamjohnson: Handed in [0804439](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/0804439c9a3b9435acd63f25b857c2122967249f).

elements/rh-select/rh-select.ts
```diff
@@ -0,0 +1,889 @@
...
+      && !event.ctrlKey
+      && !event.altKey
+      && !event.metaKey;
+
+    if (isPrintable) {
+      // Don't scroll after pressing space and don't toggle the listbox
+      event.preventDefault();
+      event.stopPropagation();
+      this.#handleTypeAhead(event.key);
+    }
```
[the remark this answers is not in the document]
adamjohnson: Handed in [1932a7a](https://github.com/RedHat-UX/red-hat-design-system/pull/2802/commits/1932a7a0aef1d45d334806791eb066a1606cf9c7).

hellogreg:
I'm getting one issue in one browser and screen reader combo--in the Methods demo, of course. For some reason in Windows Firefox with NVDA, it's not recognizing keyboard clicks on the Toggle, Show, and (I assume) Hide buttons.

Firefox works fine with a screen reader off, and it's fine with JAWS. Chrome works fine with NVDA. Mouse works as expected.

So, it's just the Firefox/NVDA/keyboard combo. Gonna check on another computer.

PR #2813, opened 2026-01-14
fix: import dependencies from 'lit' package

bennypowers:
This commit replaces imports from `lit-html` with imports from `lit`.

When importing lit utilities like `classMap`, it's possible to import them from the `lit-html` package or from the `lit` package. Both will work when using node module resolution e.g. during bundling, at development time. However, there's a problem: our package.json lists `lit` as a dependency, but not `lit-html`, which is a _transitive_ dependency of `lit`. When generating import maps from package.json, transitive dependencies are scoped to the real dependencies, thus direct imports from `lit-html` fail, when not imported from `lit` modules.

## What I did

1. search/replaced `from 'lit-html` with `from 'lit`
2. cleaned up imports, removed some unused imports

## Testing
- Try the dev server
- Check the DP, particularly demos for affected elements

## Notes to Reviewers

This fix is required to adopt `cem`/`mappa`


zeroedin (approved):
Loose Gunk Thoroughly Mopped

PR #2818, opened 2026-01-29
feat: update RH Icons to version 2.1

markcaron:
## What I did

Updated packages to get latest RH Icons (v2.1)


markcaron:
Redoing.

PR #2829, opened 2026-02-11
fix(navigation-primary): Safari 26.2 render bug

zeroedin:
## What I did

1. Corrected a rendering issue for Safari 26.2 where it isn't correctly able to use :host and container definition.  Added duplicated container definition in lightdom.

## Testing Instructions

1.  View [Deploy Preview](https://deploy-preview-2829--red-hat-design-system.netlify.app/elements/navigation-primary/demo/)
2. Test in Safari 26.2 does the navigation render correctly.
3. Ensure we didn't break anything in any other browser.

## Notes to Reviewers

Possibly related to: https://bugs.webkit.org/show_bug.cgi?id=267793? I can't completely confirm this thought because that would mean we'd see this issue elsewhere.

This same fix was applied by @zhawkins on RHDC but for a separate reason, and is why we saw RHDC working but not on ux.redhat.com


docs/styles/demo/dev-server.css
bennypowers, line 1: can we do this part in a separate pr, or is it necessary here?

elements/rh-navigation-primary/rh-navigation-primary-lightdom.css
```diff
@@ -1,4 +1,6 @@
 rh-navigation-primary {
+  container: navigation-primary / inline-size;
```
bennypowers, line 3:
```suggestion
  /* workaround for apparent regression in Safari 26.2 for containers defined on :host */
  container: navigation-primary / inline-size;
```

docs/styles/demo/dev-server.css
zeroedin: wasn't necessary, but was annoying me so i fixed it while working on the safari bug.  So it was fixed in situ.  I can move it if you really want me to.

bennypowers: it's fine

PR #2831, opened 2026-02-16
fix(switch): remove legacy fallback for `ariaDescribedbyElements` and correctly test for support of `ariaDescription`

adamjohnson:
## What I did

1. `ariaDescribedbyElements` is [now baseline](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/ariaDescribedByElements), so I removed the fallback check for it
2. Fixed the browser support test for `ariaDescription`

## Testing Instructions

1. Open up [the DP](https://deploy-preview-2831--red-hat-design-system.netlify.app/elements/switch/demo/).
2. In a Blink based browser, inspect `<rh-switch>`.
3. Once you're in "Elements", change the DevTools tab from "Styles" to "Accessibility"
4. With `<rh-switch>` highlighted, verify the "Description" field changes messages appropriately when toggling the switch.
5. Now follow steps 1-4 for the [Rich Messages demo](https://deploy-preview-2831--red-hat-design-system.netlify.app/elements/switch/demo/rich-messages/) and the [Without Messages demo](https://deploy-preview-2831--red-hat-design-system.netlify.app/elements/switch/demo/without-messages/).
    * Note: the without messages will not show anything for the "Description" field (hence the name)
6. Test these in your favorite screen reader.

### DIFF
--- a/elements/rh-switch/test/rh-switch.spec.ts
+++ b/elements/rh-switch/test/rh-switch.spec.ts
@@ -100,9 +100,15 @@ describe('<rh-switch>', function() {
       expect(snapshot.checked).to.be.false;
     });
 
+    // The a11y snapshot does not expose `description` when it is set via
+    // ElementInternals.ariaDescribedByElements. We verify the implementation instead:
+    // the slotted element for the current state has an `id` (assigned by the component)
+    // and the expected text content.
     it('should show the message for the unchecked state', async function() {
-      snapshot = await a11ySnapshot({ selector: 'rh-switch' });
-      expect(snapshot.description).to.equal('Message when off');
+      const messageOff = element.querySelector<HTMLElement>('[slot="message-off"]');
+      expect(messageOff).to.be.ok;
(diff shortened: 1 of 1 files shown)


adamjohnson:
@bennypowers 

> do you have notes on this from webkit and gecko lands, respectively? we have to support firefox esr, which by the time this goes out will be baseline 2025 iirc

`ariaDescribedByElements` is [Baseline](https://developer.mozilla.org/en-US/docs/Web/API/ElementInternals/ariaDescribedByElements) as of April 2025, so, as [also stated](https://github.com/RedHat-UX/red-hat-design-system/pull/2802#discussion_r2717069537) while working on `<rh-select>`, we should be good to just use it, yes?

We're still doing fallback checks for `ariaDescription`, so no worries there.

Release dates for [Webkit](https://browsercalendar.com/browsers/safari) and [Gecko](https://browsercalendar.com/browsers/firefox) if you wanna take a look. IMO I think we're good to go here.

PR #2770, opened 2025-12-01
feat(avatar): add public CSS variable to customize avatar grid

adamjohnson:
## What I did

1. Adds a public CSS variable for `<rh-avatar>`'s `grid-template-columns` property on the `#container` element.
2. Closes #2765.
3. Improves existing public CSS API docs

## Testing Instructions

1. Open up an avatar demo [in the DP](https://deploy-preview-2770--red-hat-design-system.netlify.app/elements/avatar/demo/).
2. In DevTools, add something like `rh-avatar { --rh-avatar-grid-columns: 100px 400px; }`
    * Alternatively, you could visit the [Custom Grid demo](https://deploy-preview-2770--red-hat-design-system.netlify.app/elements/avatar/demo/custom-grid/) and view the styles from the `<style>` block to see `--rh-avatar-grid-columns` being customized.
4. Verify the changes to the grid reflect in the browser.
5. Verify the new property is properly documented on the [Code page](https://deploy-preview-2770--red-hat-design-system.netlify.app/elements/avatar/code/#rh-avatar-css-parts-img).
6. Verify the other changes to the docs on the Code page are an improvement to what's on [uxdot](https://ux.redhat.com/elements/avatar/code/#rh-avatar-apis). 

## Notes to Reviewers

This change allows implementors much greater flexibility as to how avatar looks. As always, "With great power comes great responsibility." Designers may choose to police avatar more stringently in the future.

PR #2832, opened 2026-02-17
fix(avatar): force block layout in narrow containers

### LINKED ISSUE #2763: [bug] `<rh-avatar>` layout="inline" property does not force an inline layout

### LINKED ISSUE #2765: [feat] `<rh-avatar>`: add a public variable for `#container` `grid-template-columns` property
## Description

End users want to be able to customize the value of `grid-template-columns` on the `#container` inside rh-avatar.

## Suggested solution

1. Add a new public CSS property for the value of `grid-template-columns` eg `rh-avatar-grid-template-columns`
2. In the host's CSS, set this property to its current value: `min-content minmax(max-content, 250px)`
3. Remove the `width` property on the container
    * Note that end users should use the `min-max` function and specify a maximum width if we remove the `width` property.

## Additional context

This change is a result of a conversation with Marionne and Chris Doherty in the RHDS channel on 11-21-2025.

bennypowers:
## What I did

1. add container query to avatar
2. force block layout below a certain container width
3. add docs illustrating the issue
4. change rh-avatar's display property from `inline-block` to `block`

## Testing Instructions

1. see https://deploy-preview-2832--red-hat-design-system.netlify.app/elements/avatar/demo/layouts-in-narrow-containers/

## Notes to Reviewers
Closes #2770 
Closes #2764
Closes #2765 
Fixes #2763 

We don't expect this to break any existing page layouts, but in the event it does, something like this on that page should fix it:

```css
rh-avatar {
  display: inline-block;
  min-width: 268px;
}
```

### DIFF
--- a/elements/rh-avatar/demo/layout-inline.html
+++ b/elements/rh-avatar/demo/layout-inline.html
@@ -1,6 +1,6 @@
 <rh-avatar name="Grace Hopper"
            subtitle="Rear Admiral"
-           src="https://ux.redhat.com/elements/avatar/demo/hopper.jpg"
+           src="hopper.jpg"
            layout="inline"></rh-avatar>
 
 
--- a/web-dev-server.config.js
+++ b/web-dev-server.config.js
@@ -1,5 +1,6 @@
 // @ts-check
 import { pfeDevServerConfig } from '@patternfly/pfe-tools/dev-server/config.js';
+import { deslugify } from '@patternfly/pfe-tools/config.js';
 import { glob, readFile } from 'node:fs/promises';
 import { join } from 'node:path';
 import { makeDemoEnv } from './scripts/environment.js';
@@ -227,6 +228,34 @@ export default pfeDevServerConfig({
         console.error(e);
       }
       return;
+    },
+    /**
+     * Serve demo assets for element demo pages.
+     * Demo index pages are served at /elements/<slug>/ so relative asset
+     * references (e.g. src="khayyam.jpg") resolve to /elements/<slug>/asset
+     * instead of /elements/<slug>/demo/asset. This middleware maps those
+     * requests to the actual file in elements/<tagName>/demo/.
+     * @param ctx koa context
+     * @param next next koa middleware
+     */
+    async function(ctx, next) {
+      if (/\.(js|ts|css|html|map)$/i.test(ctx.path)) {
+        return next();
+      }
+      const match = ctx.path.match(/^\/elements\/([\w-]+)\/(?:.*\/)?([\w.-]+\.\w+)$/);
+      if (match) {
+        const [, slug, filename] = match;
+        const tagName = deslugify(slug);
+        const filePath = join(process.cwd(), 'elements', tagName, 'demo', filename);
+        try {
+          ctx.body = await readFile(filePath);
+          ctx.type = filename.split('.').pop() ?? 'bin';
+          return;
+        } catch {
+          // File not found in demo directory, fall through
+        }
+      }
+      return next();
     },
     /**
      * @param ctx koa context
--- a/elements/rh-avatar/demo/color-context.html
+++ b/elements/rh-avatar/demo/color-context.html
@@ -9,7 +9,7 @@
              plain></rh-avatar>
   <rh-avatar name="Grace Hopper"
              subtitle="Rear Admiral"
-             src="https://ux.redhat.com/elements/avatar/demo/hopper.jpg"
+             src="hopper.jpg"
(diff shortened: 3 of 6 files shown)


elements/rh-avatar/demo/layouts-in-narrow-containers.html
```diff
@@ -0,0 +1,69 @@
...
+
+<style>
+  #test {
+    display: grid;
+    grid-template-columns: repeat(1, minmax(0, 1fr));
+    gap: var(--rh-space-xl);
+    padding: var(--rh-space-xl);
+    justify-content: space-between;
+    flex-wrap: wrap;
+    width: 450px;
```
adamjohnson:
```suggestion
    max-inline-size: 450px;
```

ala [WCAG SC 1.4.10](https://www.w3.org/WAI/WCAG21/Understanding/reflow.html)

elements/rh-avatar/demo/layouts-in-narrow-containers.html
```diff
@@ -0,0 +1,69 @@
...
+
+    &.repro {
+      width: 250px;
```
adamjohnson:
```suggestion
      max-inline-size: 250px;
```

elements/rh-avatar/demo/layouts-in-narrow-containers.html
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+
+  rh-avatar {
+    --rh-avatar-size: var(--rh-length-3xl);
```
adamjohnson: `var(--rh-length-3xl)` 

Should we include fallback values here in case these demos are copy and pasted into other environments?

elements/rh-avatar/rh-avatar.css
```diff
@@ -1,38 +1,58 @@
...
+*:before {
+  box-sizing: border-box;
+}
+
 :host {
-  display: inline-block;
-  width: min-content;
+  display: block;
+  container: avatar / inline-size;
+  width: 100%;
```
adamjohnson:
```suggestion
  inline-size: 100%;
```

elements/rh-avatar/demo/layout-inline.html
adamjohnson, line 1: When an avatar does not have the attribute `layout="block/inline"`, the avatar is not responsive at small screen sizes. This is a behavior change from the current implementation on uxdot.

What's the best way forward here? If an avatar does not have a `layout` attribute, should it default to `layout="inline"`?

elements/rh-avatar/rh-avatar.ts
```diff
@@ -77,12 +77,18 @@ export class RhAvatar extends LitElement {
   }
 
   render() {
+    const { layout, name, pattern, plain, src, subtitle, variant } = this;
```
adamjohnson, line 71: Nice destructuring here, very clean 👏

elements/rh-avatar/demo/layout-inline.html
bennypowers: 2d6d20cc

elements/rh-avatar/rh-avatar.css
```diff
@@ -1,153 +1,209 @@
...
-  @media (max-width: 575px) {
-    margin-block-end: var(--rh-space-lg, 16px);
-  }
-}
+  /** Avatar subtitle font weight */
+  font-weight: var(--rh-font-weight-heading-regular, 400);
 
-slot {
-  display: block;
-  max-width: 250px;
```
adamjohnson, line 77: Avatar width uxdot vs this DP, this DP doesn't have a max width

One key difference this PR makes is that the avatar Title and Subtitle no longer have this max width. This will probably cause a layout change for current users of rh-avatar. Is there a way to avoid that?

bennypowers: 86e70668

web-dev-server.config.js
```diff
@@ -228,6 +229,34 @@ export default pfeDevServerConfig({
...
+        try {
+          ctx.body = await readFile(filePath);
+          ctx.type = filename.split('.').pop() ?? 'bin';
+          return;
+        } catch {
+          // File not found in demo directory, fall through
+        }
+      }
+      return next();
+    },
```
zeroedin, line 259: I added this change to allow the relative paths of the images to work in both dev-server and prod

bennypowers: Oof thanks for that. I fixed it in cem but forgot to support wds config

PR #2841, opened 2026-02-23
fix(tooltip): activation intentionality

zeroedin:
## What I did

1. Corrected the css delay for a users intentionality to display a tooltip

## Testing Instructions

1. browse to DP demo https://deploy-preview-2841--red-hat-design-system.netlify.app/elements/tooltip/demo/dense-ui/
2. run your mouse along the column of buttons
3. expected: no noisy flash of multiple tooltips
4. place your mouse on a single button
5. expected: tooltip appears after a barely perceptible delay

## Notes to Reviewers

Designers please ensure that this change meets your expectations.

Tooltip now takes 300ms before it appears, this allows a users intention to be clear.  Then reverses the fade out of a tooltip instead removing it quickly as the user has moved on.


bennypowers (approved):
lazily generate tooltip messages

PR #2842, opened 2026-02-24
docs(button-group): add docs around `role="group"` and `role="toolbar"`

### LINKED ISSUE #2769: [docs] `<rh-button-group>`: add accessibility docs
### Description

Docs need to be added for `<rh-button-group>`'s Accessibility page about:
 
- Wrapping button groups with `role="toolbar"` (where appropriate)
    - If we decide to implement `<rh-button-group-toolbar>`, factor that into these docs
- When users should use the button group element with relation to accessibility.
- Potentially docs around `aria-orientation` in combination with `role="toolbar"`.
- Docs around possibly adding `aria-label` to button groups.

### Links

- [MDN Group role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role)
- [MDN Toolbar role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/toolbar_role)
- [MDN ARIA Orientation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation)

### Screenshots

_No response_

### Additional context

See the `<rh-button-group>` PR: https://github.com/RedHat-UX/red-hat-design-system/pull/2605#pullrequestreview-3450333253.

adamjohnson:
## What I did

1. Added docs surrounding `role="group"` and `role="toolbar"` to the `<rh-button-group>` Accessibility page.
1. Closes #2769.

## Testing Instructions

1. Go to the Accessibility page for `<rh-button-group>` [in the DP](https://deploy-preview-2842--red-hat-design-system.netlify.app/elements/button-group/accessibility/).
2. Verify spelling, grammar, clarity, order.
3. Verify content is accurate in regards to accessibility.
4. Like an subscribe.


elements/rh-button-group/rh-button-group.ts
bennypowers, line 1: this has to go in the appropriate places on public apis like slots and attr, or in the element description, otherwise it won't make it into the final manifest

see #2738 for examples

adamjohnson: Solved in [44febc0](https://github.com/RedHat-UX/red-hat-design-system/pull/2842/commits/44febc09057a3afcdbb326def8c2234b7ff17722). PTAL.

PR #2843, opened 2026-02-25
fix(switch): make css stack more resilient

bennypowers:
## What I did

1. make css stack explicit with fallbacks for checked state

## Testing Instructions

1. see DP/elements/switch/demo/ in a chromium browser, then open dev tools
2. see that colours obtain

## Notes to Reviewers
Thanks Michael Wells for reporting this issue

### DIFF
--- a/.changeset/tasty-games-chew.md
+++ b/.changeset/tasty-games-chew.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-switch>`: fix switch rendering on some browsers in light/dark schemes


elements/rh-switch/rh-switch.css
```diff
@@ -75,7 +75,10 @@
 
   &.checked {
     --_switch-track-background-color: /** The background color of the switch when it is checked. */
-      var(--rh-switch-checked, var(--rh-color-accent-base));
+      var(--rh-switch-checked,
+        light-dark(
+          var(--rh-color-accent-base-on-light, #0066cc),
+          var(--rh-color-accent-base-on-dark, #92c5f9)));
```
zeroedin, line 81: Is this an attempt to ensure accent colors are loaded before theming css injection?

bennypowers: In my experience light-dark() has to be called as far as possible down the chain in order to pick up computed properties. i think blink's css renderer somehow doesn't pick up this edge case. in any event i tested with "ungoogled chromium" flatpak on gentoo and it worked with dev tools open

PR #2845, opened 2026-02-26
fix(pagination): prevent input from stealing focus on page load

zeroedin:
## What I did

1.  Prevent input from stealing focus on page load due to incorrect reporting of validity on lifecycle events, should only report validity on user interaction.

## Testing Instructions

1.  View deploy preview [borderless variant demo](https://deploy-preview-2845--red-hat-design-system.netlify.app/elements/pagination/demo/borderless-compact/) and reduce viewport to a mobile size, refresh the page, the input field should not gain focus on reload. 
2. Compare to ux dot [open demo](https://ux.redhat.com/elements/pagination/demo/open/) under same constraints, you should notice that the production demo does gain focus.  This is incorrect.

### DIFF
--- a/.changeset/proud-socks-stare.md
+++ b/.changeset/proud-socks-stare.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-pagination>`: prevented page number input from receiving focus on page load
+  


elements/rh-pagination/test/rh-pagination.spec.ts
```diff
@@ -23,6 +24,30 @@ describe('<rh-pagination>', function() {
...
+            <li><a href="#3">3</a></li>
+            <li><a href="#4">4</a></li>
+            <li><a href="#5">5</a></li>
+          </ol>
+        </rh-pagination>
+      `);
+      await element.updateComplete;
+      await nextFrame();
+    });
+
```
adamjohnson, line 44:
```suggestion
    it('should be accessible', async function() {
      await expect(element).to.be.accessible();
    });
```

I can never get the whitespace right when adding these suggestions. Would recommend adding this manually instead of via the Github UI to make sure white space is formatted correctly.

zeroedin: as discussed this will go into the next PR opening to fix the onchange -> onsubmit.

PR #2846, opened 2026-02-26
fix(pagination): replace onchange with onsubmit

zeroedin:
## What I did

1. Wrapped numeric input in a `<form>` for mobile submit support
2. Added `enterkeyhint="go"` attribute replacing `inputmode="numeric"` giving the mobile keyboard a enter key the represents a navigation action.

## TODO

- [x] Write tests

## Testing Instructions

1.  View [demo](https://deploy-preview-2846--red-hat-design-system.netlify.app/elements/pagination/demo/) Changing the value of the input will no longer update the URL onChange, now requires a enter keypress.  On mobile the `enterkeyhint="go"` should display a navigation style enter key.

## Notes to Reviewers

Please look at mobile browser if you have android phone I do not so I can't test the enterkeyhint there.

This change does change the user expectation that the input field change auto submits via onChange -> onSubmit which needs the user to press enter or hit the "go" button?  Does we think this constitute enough for a major change?  Feel it maybe in a grey area.

### DIFF
--- a/elements/rh-pagination/rh-pagination.ts
+++ b/elements/rh-pagination/rh-pagination.ts
@@ -12,7 +12,6 @@ import { Logger } from '@patternfly/pfe-core/controllers/logger.js';
 import { themable } from '@rhds/elements/lib/themable.js';
 
 import styles from './rh-pagination.css' with { type: 'css' };
-import { classMap } from 'lit/directives/class-map.js';
 
 const L1 = html`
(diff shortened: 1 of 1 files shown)


bennypowers (approved):
Lucid "go" text manifested

adamjohnson (approved):
Tested on Android 16 with Chrome and Opera Mobile. Works as advertised. Nice job.

PR #2850, opened 2026-03-04
feat(drawer): add `<rh-drawer>`

### LINKED ISSUE #2465: [feat] `<rh-drawer>`

zeroedin:
## What I did

A collapsible side-panel component (`<rh-drawer>`) with four layout variants, responsive breakpoint switching, and full accessibility support.

### Variants

- **`auto`** (default): Overlay at narrow container widths, inline at wider widths. The breakpoint is configurable via `overlay-threshold`. When collapsible, the collapse toggle is visible in inline mode; in overlay mode a `mode-change` event fires so consumers can show/hide an external trigger.
- **`overlay`**: Panel overlays content without shifting layout. Supports `collapsible`, `resizable`, and `none` panel modes. When collapsible, fires `mode-change` to indicate whether the built-in toggle is visible (`collapsible`) or the consumer needs an external trigger (`overlay`).
- **`fixed`**: Panel is `position: fixed`, overlaying the page at all viewport sizes. Toggled exclusively by an external trigger via `trigger-id`. Ignores `collapsible` panel mode.
- **`flow`**: Panel participates in the parent layout above the breakpoint (sidebar) and becomes a fixed overlay below it. Open state tracks the media query and restores the user's last explicit choice when crossing back below the breakpoint.

### Overlay threshold

Configurable breakpoint via `overlay-threshold` using t-shirt sizes (`md`, `lg`, `xl`, `2xl`) mapped to `@rhds/tokens/media.js` values. Defaults to 768px when not set.

- For `auto`: JS `ResizeObserver` checks container width against the threshold and toggles between overlay and inline layout.
- For `flow`: viewport-width `@media` query controls the layout switch.

### Panel modes (`panel` attribute)

- **`collapsible`** (default for auto/overlay): Integrated toggle button that partially collapses the panel while keeping the toggle visible.
- **`resizable`**: Drag handle for manual width resizing via pointer or keyboard (fixed and overlay variants only; ignored for auto/flow).
- **`none`**: No integrated panel controls. Requires `trigger-id` for reopening.

### Events

- **`open`** (cancelable, composed): Fires when the panel opens. `event.trigger` holds the HTMLElement that initiated the action, or `null` when opened via `show()`. Call `preventDefault()` to keep the drawer closed.
- **`close`** (cancelable, composed): Fires when the panel closes. Call `preventDefault()` to keep the drawer open.
- **`mode-change`** (composed): Fires when the layout mode changes. `event.mode` is `'inline'`, `'overlay'`, or `'collapsible'`. Consumers should listen for this to show/hide external trigger buttons.

### Expand mode

`expand="full-viewport"` adds a toggle button that expands the panel to fill the entire viewport. Collapse toggle and resize handle are hidden in full-viewport mode.

### Accessibility

- Panel uses `role="dialog"` for overlay-style layouts (fixed, overlay) and `role="complementary"` for inline layouts. Auto and flow variants switch roles dynamically at their breakpoints.
- `accessible-label` provides an `aria-label` fallback (defaults to "Panel"). When the header slot has content, `aria-labelledby` references the header instead.
- Escape closes dialog-role panels from a single document-level listener. Complementary-role panels do not close on Escape.
- `aria-controls`, `aria-expanded`, and `aria-labelledby` on toggle buttons.
- Resize handle uses `role="separator"` with `aria-valuenow`, `aria-valuemin`, `aria-valuemax`.
- Closed panels receive the `inert` attribute so they are removed from the tab order and hidden from assistive technology.

### Other features

- **`position`**: Panel on `inline-start` or `inline-end` side, with full RTL support via logical properties and `:dir(rtl)`.
- **`trigger-id`**: Links an external button to toggle the panel. Required when `panel` is `none` or `resizable`.
- **`storage-key`**: Per-property `sessionStorage` persistence (open state, panel mode, panel width).
- **`--rh-drawer-panel-padding`** and **`--rh-drawer-content-padding`** custom properties for consumer overrides.
- Transitions are suppressed during variant/position changes and container resize breakpoint crossings to prevent visual jank.
- Respects `prefers-reduced-motion: reduce`.

Closes #2465

## Testing instructions

1. [View DP demos](https://deploy-preview-2850--red-hat-design-system.netlify.app/elements/drawer/demos/)

## Notes to reviewers


.changeset/slimy-mugs-repeat.md
```diff
@@ -0,0 +1,15 @@
+---
+"@rhds/elements": minor
+---
+
+✨ Added `<rh-drawer>`
+
+A collapsible side panel for supplementary content or navigation. Supports four layout variants (`auto`, `overlay`, `fixed`, `flow`), configurable breakpoints via `overlay-threshold`, and panel styles with `collapsible`, `resizable` and `none`. Fires `mode-change` events so consumers can respond to layout switches. Set `storage-key` to persist open state, panel mode, and panel width across page navigation via `sessionStorage`.
+
```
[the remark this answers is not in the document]
zeroedin: This is true and untrue, the pattern is just slotting the `<rh-navigation-vertical>` into the body of the panel.  So it is true that we need to likely call out a dedicated pattern, there is a demo that does slot the `<rh-navigation-vertical>`.  Maybe I should remove that it closes the sidenav pattern PR. https://github.com/RedHat-UX/red-hat-design-system/issues/1329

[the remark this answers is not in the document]
bennypowers: agreed, that issue requires both to land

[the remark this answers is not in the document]
zeroedin: I've done quite a bit of around the pattern , but ran into and rabbit holed a bit on the uxdot-pattern problems and the dual layer SSR  / hydration issues which I have a solution for.  I am going to pull out that code from my working branch and open a  PR to demonstrate /resolve that issue.  

But for the sake of drawer specifically, the uxdot-pattern itself is insufficient due to its limited horizontal viewport scale for some patterns which take full width media queries such as a drawer use of the flow variant to reproduce ux dot style navigation.  We can fix some of these with improvments to uxdot-pattern itself, fullscreen style api etc.

For now I'm going to remove the closing of #1329 on this PR, we can deal with the pattern issue post merge.

.changeset/empty-birds-invent.md
adamjohnson, line 1: With the exception of `slimy-mugs-repeat.md`, these other changesets should be removed.

elements/rh-drawer/demo/fixed.html
adamjohnson, line 1: Heads up: this demo seems to violate [WCAG 2.4.11: Focus not obscured AA](https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum.html). 

Consider adding `inert` to the page outside of the drawer so that the drawer doesn't cover interactive elements and violate 2.4.11.

elements/rh-drawer/demo/overlay.html
adamjohnson, line 1: Same as the fixed demo (see [comment](https://github.com/RedHat-UX/red-hat-design-system/pull/2850/changes#r3102913178)), make sure to deactivate the page content while the drawer is open so we don't violate WCAG 2.4.11.

elements/rh-drawer/rh-drawer.css
```diff
@@ -0,0 +1,844 @@
...
+  :host {
+    --_panel-width: 320px;
+    --_panel-min-width: 320px;
+
+    /** Collapse toggle button size */
+    --_collapse-toggle-size: var(--rh-length-3xl, 48px);
+
+    /** Visible width reserved for the collapse toggle when the panel is closed */
+    --_collapse-visible-width: var(--rh-space-2xl, 32px);
+    --_resize-handle-width: 12px;
```
adamjohnson, line 19: Keep in mind touch targets must have a minimum clickable area of 24px by 24px. So, we'd need to increase the clickable width here to 24px one way or another.

See [WCAG SC 2.5.8 Target Size Minimum](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html).

elements/rh-drawer/docs/40-accessibility.md
adamjohnson, line 1: We want to make sure to have good docs here surrounding the use of drawer's features. Things like how different slots interact with aria-labelledby and attributes that are required for drawer triggers.

elements/rh-drawer/rh-drawer.css
```diff
@@ -0,0 +1,844 @@
+@layer base, states, variants, overrides;
```
adamjohnson, line 1: Nice use of layers!

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,770 @@
...
+             aria-label="${!hasHeader ? this.accessibleLabel : nothing}"
+             aria-labelledby="${hasHeader ? 'header' : nothing}"
+             ?inert=${!this.open && !showCollapsible}
+             style=${styleMap(panelStyles)}>
+          <div id="panel-body"
+               ?inert=${!this.open && showCollapsible}>
+            <div id="actions" ?hidden="${!this.expand && showCollapsible}">
+              ${this.expand ? html`
+                <!-- csspart expand-button: The full-viewport expand toggle button. -->
+                <button id="expand-button"
```
adamjohnson, line 268: We may consider adding `aria-pressed` to this button to help communicate state.

adamjohnson (changes requested):
Here's a first pass accessibility review. I'd like to address these then take another look.

There's certainly a question about whether drawer is is a modal or non-modal (and that can change depending on the viewport and demo). In many cases, it's probably non-modal. We should be sure to document this in 40-accessibility.md when the time comes.

elements/rh-drawer/rh-drawer.css
adamjohnson: cc @coreyvickery

elements/rh-drawer/docs/40-accessibility.md
zeroedin: Lots of implementation examples and the docs surrounding them are 100% needed.

.changeset/empty-birds-invent.md
zeroedin: Good catch, must have have been a bad merge.  Missed that.

elements/rh-drawer/demo/overlay.html
zeroedin: @adamjohnson implemented inert for when drawer owns the content then updated fixed and flow demos to show how this would be done in a implementation pattern.

hellogreg:
As Adam said above, we want to shoot for 24x24 targets. As far as WCAG conformance, the resize bar _is_ under 24px square, but it can be considered conformant under the ["undersized targets" exception](https://www.w3.org/TR/WCAG22/#target-size-minimum), if no other targets are within a 24px diameter bubble from the bar's center:

> The size of the target for pointer inputs is at least 24 by 24 CSS pixels, except when:
> 
> **Spacing**
> Undersized targets (those less than 24 by 24 CSS pixels) are positioned so that if a 24 CSS pixel diameter circle is centered on the bounding box of each, the circles do not intersect another target or the circle for another undersized target;

The biggest risk for violating this would likely be any content that is underneath the drawer, but if it's all inert when the drawer is open, then no interactive element is currently a possible target.

CleanShot 2026-05-06 at 15 26 17


That said, 12 pixels is pretty narrow. I was testing it on my phone and didn't always hit the target. (Could one take that exception to its illogical extreme and have a 1px square target, so long as it's spaced properly? I guess so.) Also, I'd think we want some visual cues that the inert content is not currently active.

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+            </button>
+            <span id="collapse-label" class="visually-hidden">
+              <span ?hidden=${!this.open}><!-- Accessible label for the collapse toggle when panel is open --><slot name="collapse-label-open">Collapse panel</slot></span>
+              <span ?hidden=${this.open}><!-- Accessible label for the collapse toggle when panel is closed --><slot name="collapse-label-closed">Expand panel</slot></span>
+            </span>
+          ` : nothing}
+        </div>
+        ${hasContentSlot ? html`
+          <div id="content-container"
+               ?inert=${this.open && this.#panelRole === 'dialog'}>
```
bennypowers:
```suggestion
               ?inert="${this.open && this.#panelRole === 'dialog'}">
```

et al.

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+                    @click=${this.toggle}>
+              <rh-icon set="ui" icon="caret-left"></rh-icon>
+            </button>
+            <span id="collapse-label" class="visually-hidden">
+              <span ?hidden=${!this.open}><!-- Accessible label for the collapse toggle when panel is open --><slot name="collapse-label-open">Collapse panel</slot></span>
+              <span ?hidden=${this.open}><!-- Accessible label for the collapse toggle when panel is closed --><slot name="collapse-label-closed">Expand panel</slot></span>
+            </span>
+          ` : nothing}
+        </div>
+        ${hasContentSlot ? html`
```
bennypowers:
```suggestion
        ${!hasContentSlot ? nothing : html`
```

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+    const hasFooter = this.#slots.hasSlotted('footer');
+    const panelStyles = this._panelWidth != null ?
+        { '--_panel-width': `${this._panelWidth}px` }
+      : {};
+    const hasContentSlot = this.variant !== 'fixed' && this.variant !== 'flow';
+    const isAuto = this.variant === 'auto';
+    const isFlow = this.variant === 'flow';
+    const classes = {
+      'open': this.open,
+      'closed': !this.open,
```
bennypowers: if we can make these to single-word classnames, and try to reduce their number, better.

```ts
    return html`
      <div id="container"
           class="${classMap({ open, auto, fixed, flow, wide, overlay,
                              inline, start, end, mode,
                              collapsible, resizable, full })}">
```

maybe linter won't like? maybe we should adjust our lint rules.

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+             aria-label="${!hasHeader ? this.accessibleLabel : nothing}"
+             aria-labelledby="${hasHeader ? 'header' : nothing}"
+             ?inert=${!this.open && !showCollapsible}
+             style=${styleMap(panelStyles)}>
+          <div id="panel-body"
+               ?inert=${!this.open && showCollapsible}>
+            <div id="actions" ?hidden="${!this.expand && showCollapsible}">
+              ${this.expand ? html`
+                <!-- csspart expand-button: The full-viewport expand toggle button. -->
+                <button id="expand-button"
```
bennypowers, line 285: This should probably _not_ be a part of the drawer element, but a command target in page elements like nav

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+                <button id="close-button"
+                        part="close-button"
+                        type="button"
+                        aria-controls="panel"
+                        aria-labelledby="close-label"
+                        @click=${this.close}>
+                  <rh-icon set="microns" icon="close"></rh-icon>
+                </button>
+                <span id="close-label" class="visually-hidden">
+                  <slot name="close-label">Close drawer</slot>
```
bennypowers: make this a slot/attr pair

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+              ` : nothing}
+            </div>
+            <div id="panel-content">
+              <!-- csspart header: The panel header area for slotted heading content. -->
+              <div id="header" part="header" ?hidden=${!hasHeader}>
+                <slot name="header"></slot>
+              </div>
+              <!-- csspart body: The panel body content area. -->
+              <div id="body" part="body" ?hidden=${!hasBody}>
+                <slot name="body"></slot>
```
bennypowers, line 322: why is there a panel container with layout slots, including a body slot, but also an anonymous slot for content?

this seems overloaded. drawer should open and close, and have some content.

what content goes in the drawer should be a pattern level thing

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,785 @@
...
+    this.open = !this.open;
+    this.#setFlowUserState(this.open);
+  }
+}
+
+declare global {
+  interface HTMLElementTagNameMap {
+    'rh-drawer': RhDrawer;
+  }
+}
```
bennypowers, line 802: Let's see how this stacks up:
 
- Shoelace: [332 LoC](https://github.com/shoelace-style/webawesome/blob/next/packages/webawesome/src/components/drawer/drawer.ts)
- MWC: [72 LoC](https://github.com/material-components/material-web/blob/main/labs/navigationdrawer/internal/navigation-drawer.ts)
- Fluent: [210 LoC](https://github.com/microsoft/fluentui/blob/master/packages/web-components/src/drawer/drawer.ts)
- Lion: [228 LoC](https://github.com/ing-bank/lion/blob/master/packages/ui/components/drawer/src/LionDrawer.js)
- Spectrum: [215 LoC](https://github.com/adobe/spectrum-web-components/blob/main/1st-gen/packages/sidenav/src/Sidenav.ts)
- SAP: [765 LoC](https://github.com/UI5/webcomponents/blob/main/packages/fiori/src/SideNavigation.ts) (unsurprisingly)
- Vaadin: [1 LoC](https://github.com/vaadin/web-components/blob/e4c30d4335d7b62ecf660f3cedca02b282596ea2/packages/app-layout/src/vaadin-app-layout.js#L153)

I realize  some of those are unfair comparisons, like SWC which splits things into smaller subcomponents

But the general trend is clear: this element tries to do *too much*

bennypowers (changes requested):
We need to drastically simplify this element.

elements/rh-drawer/rh-drawer.ts
zeroedin: good catch

elements/rh-drawer/rh-drawer.ts
zeroedin: fixed in [0b62bab](https://github.com/RedHat-UX/red-hat-design-system/pull/2850/commits/0b62bab1afaf4ce00c7e1cbebb926590808089e2)

elements/rh-drawer/rh-drawer.ts
zeroedin: fixed in [0b62bab](https://github.com/RedHat-UX/red-hat-design-system/pull/2850/commits/0b62bab1afaf4ce00c7e1cbebb926590808089e2)

elements/rh-drawer/rh-drawer.ts
zeroedin: I wanted to be careful with `inline-start` and `inline-end` as we could essentially have `block-start` and `block-end`.   Patternfly drawer also has those.  Does that change your opinion there?  inline-mode is different from the positions didn't want to conflate those either.  While these are internal they made it easier for me to reason by being more explict.

open and closed likely can merged.  `full-viewport` can be dropped to `full` although if we remove that feature completely as you suggest in a later comment maybe we don't need it at all.

elements/rh-drawer/rh-drawer.ts
zeroedin: The anonymous slot is what is inside the drawer, the panel is like the front fascia of the door with the handle so to speak to make it analogy.  

We need to wrap the contents otherwise half this feature set won't work.   A fixed position drawer does not need to own its content, but others do, so it depends on the use case.  This speaks to the complexity problem you've already mentioned and that should be looked at.

elements/rh-drawer/rh-drawer.ts
zeroedin: I dont disagree, and we should look at how these different features work and where to break them up.  They all have a use case.

zeroedin: I would like that to be design spec driven.

elements/rh-drawer/rh-drawer.ts
zeroedin: Are you thinking slottable actions?

elements/rh-drawer/rh-drawer.ts
zeroedin: open and closed is now a single class.

elements/rh-drawer/rh-drawer.ts
bennypowers: we spoke about this on call today and the path forward is for @coreyvickery and @zeroedin to work together in order to clarify the design spec
- visually, and in terms of supported features, everything in the design spec stays the same as it is now
- BUT we clarify which parts of the design apply to `<rh-drawer>` - a minimal container with lowest-common-denominator design applied in css - and which go in patterns documentation

PR #2861, opened 2026-03-06
fix(pagination): simplify to single input with wrapped form and mobile submit

zeroedin:
## What I did

1.  Wrapped numeric input in a form for mobile submit capabilities
2. Limited render to single instance of the input that is always in the end position on all viewports as suggested in chat

Replaces the closed branch #2846 to tidy up the changes.

## Testing Instructions

1. View Deploy preview

## Notes to Reviewers

@coreyvickery  If additional design spec changes are necessary let me know.  I know there was an outstanding question about how this moves forward with the unified theme by @markcaron.  

@hellogreg when you have a moment give this a once over again in screen readers, not much if anything should have changed from the previous review in #2846 but just to be certain I didn't break anything in pulling those changes here.

### DIFF
--- a/elements/rh-pagination/rh-pagination.ts
+++ b/elements/rh-pagination/rh-pagination.ts
@@ -217,6 +217,7 @@ export class RhPagination extends LitElement {
                    name="page"
                    min=1
                    max="${this.total}"
+                   @click="${this.#onClick}"
                    .value="${currentPage}">
           </form>
         <!-- summary: preposition text between page input and total (default: "of")
@@ -345,6 +346,13 @@ export class RhPagination extends LitElement {
     }
     this.requestUpdate();
     await this.updateComplete;
+  }
+
+  /** Firefox does not focus the input when clicking spinner arrows */
+  #onClick(event: Event) {
+    if (event.target instanceof HTMLInputElement) {
+      event.target.focus();
+    }
   }
 
   #onSubmit(event: Event) {


elements/rh-pagination/rh-pagination.ts
```diff
@@ -319,50 +302,35 @@ export class RhPagination extends LitElement {
...
+   * String ids click shadow DOM steppers after rendering ensures their href is set.
+   * After the click, request an update in case a SPA prevented full navigation.
+   * @param id stepper name or 1-based page number
    */
   async #go(id: 'first' | 'prev' | 'next' | 'last' | number) {
-    await this.updateComplete;
     if (typeof id === 'number') {
-      const link = this.#links?.[id - 1];
-      link?.click?.();
+      this.#links?.[id - 1]?.click();
```
bennypowers, line 342: this might have unintended side effects in spa environments, etc.
perhaps we should use the navigation API instead?

bennypowers: We'll revisit this when navigation api is more mature

zeroedin: 100% on the navigation api.  Exactly my thoughts the other day and even posted about it in the slack channel that it is on my todo list to learn asap.

coreyvickery (changes requested):
@zeroedin Small things.

- [x] Greg posted a video of something I see too, but it works as normal after a few tries
- [x] Number boxes seem to be a tad bigger than the control boxes
- [ ] Input field with page number has no border (on landing it doesn't, but it does when I navigate to a different page)
- [x] Controls (except for page input field tiny up/down arrows) don't work (not sure if that's a bug or not for this particular PR
- [ ] When the page number stuff is below the controls, reduce the padding to `--rh-space-xl`
- [x] Mobile behavior and location of page number stuff looks good

zeroedin:
> Not sure if this is a Firefox thing or an us thing.

Was definitely a firefox thing, added code that now forces focus back on the input after the spinner is clicked.

https://bugzilla.mozilla.org/show_bug.cgi?id=1012818

hellogreg (approved):
Tested with mouse, keyboard, and screen reader:
- [x] Mac Safari (VoiceOver)
- [x] Mac Chrome (VoiceOver)
- [x] Mac Firefox (VoiceOver)
- [x] Windows Edge (Narrator)
- [x] Windows Chrome (JAWS)
- [x] Windows Firefox (NVDA)

Tested with touchscreen, onscreen keyboard, and screen reader:
- [x] iOS Safari (VoiceOver)
- [x] Android Chrome (TalkBack)

LGTM

PR #2890, opened 2026-03-25
docs: strip meta itemprop tags from demo code blocks

bennypowers:
## Summary

- Strips `<meta itemprop>` elements (e.g. `<meta itemprop="description" content="...">`) from demo HTML before rendering code blocks on element pages
- These tags are used by CEM for health scoring but shouldn't appear in the displayed demo source on the docs site

@adamjohnson this should address some of your concerns about meta tags showing up in demos. We can work on a more comprehensive solution after the eevee release.

## Test plan

- [x] Verify element demo pages no longer show `<meta itemprop="description">` in the HTML code blocks
- [x] Verify CEM health scores are unaffected (meta tags are still in the source files, just stripped at render time)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bennypowers:
Thanks Adam

PR #2953, opened 2026-04-16
fix(audio-player): document media slot via class-level JSDoc

### LINKED ISSUE #2923: [bug] `<rh-audio-player>` media slot missing documentation
## Description

The `media` slot on `rh-audio-player` has an inline HTML doc comment, but its `summary` and `description` are not included in the generated `custom-elements.json` manifest. This means the slot appears without documentation on the [Code page](https://ux.redhat.com/elements/audio-player/code/).

## Root cause

An `<input type="hidden">` void element immediately before the doc comment prevents the CEM analyzer (`@pwrs/cem`) from associating the comment with the `<slot>`. This is a tree-sitter HTML parser issue in the CEM tool — tracked upstream.

## Fix

Wrap the `<input>` in a `<span hidden>` so the HTML parser produces a proper element node as the preceding sibling:

```diff
-        <input type="hidden" value="${this.#readyState}">
+        <span hidden><input type="hidden" value="${this.#readyState}"></span>
         <!-- summary: html `audio` element
              description: |
                Must contain an `<audio>` block element with source children. -->
         <slot id="media" name="media" @slotchange="${this.#initMediaElement}"></slot>
```

This has been verified locally — the slot documentation appears correctly in the manifest after the change.

## Related

- Upstream CEM bug: https://github.com/bennypowers/cem/issues/285

bennypowers:
## Summary
- The `media` slot's inline HTML doc comment was not included in the CEM manifest because a preceding void `<input>` element prevented the tree-sitter parser from associating the comment with the `<slot>`.
- Added a `@slot media` tag to the class-level JSDoc as a workaround, so the slot description now appears in `custom-elements.json` and on the Code page.

Closes #2923

## Test plan
- [x] `cem generate -q` succeeds
- [x] `cem health --component rh-audio-player` shows slots at 15/15
- [x] `media` slot appears in `custom-elements.json` with its description

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/elements/rh-audio-player/rh-audio-player.ts
+++ b/elements/rh-audio-player/rh-audio-player.ts
@@ -42,6 +42,10 @@ import '@rhds/elements/rh-icon/rh-icon.js';
  *
  * @alias audio-player
  *
+ * @slot media - Must contain an `<audio>` block element with source children.
+ *               The audio element is visually hidden but remains accessible
+ *               to screen readers for native media controls.
+ *
  * @csspart toolbar - The main controls toolbar container.
  * @csspart panel - The expandable content panel below the toolbar.
  * @csspart about - The about panel slot container.


zeroedin (approved):
Layered Granular Text: Media

PR #2131, opened 2025-01-16
fix: use `@lit/context`

bennypowers:
## What I did

1. replace our custom context protocol implementation with lit's

## Testing Instructions

1. tbd

## Notes to Reviewers
pay special attention to the heading level elements, this may require a wholesale refactor to get working under the new context regime.

PR #2954, opened 2026-04-16
fix(accordion): guard SSR-unsafe DOM operations

bennypowers:
## Summary
- Guard `updateAccessibility()` and `MutationObserver.observe()` behind `isServer` in `connectedCallback`, matching the pattern already used in other elements (accordion-header, code-block, tabs, navigation-secondary).

Supersedes the accordion portion of #2131.

## Test plan
- [ ] SSR build completes without errors
- [ ] Accordion expands/collapses normally in the browser
- [ ] Accordion headers have correct `aria-controls`/`aria-labelledby` after hydration

### DIFF
--- a/.changeset/accordion-ssr-guard.md
+++ b/.changeset/accordion-ssr-guard.md
@@ -1,0 +1,4 @@
+---
+"@rhds/elements": patch
+---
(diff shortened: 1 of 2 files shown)


zeroedin (approved):
Lit Guarded; Timely Manipulation

PR #2955, opened 2026-04-16
Fix/uxdot pattern/ssr flow

zeroedin:
## What I did

Replaces uxdot-pattern's Lit SSR controller architecture with a build-time Eleventy transform that inlines pattern content as slotted light DOM children.

Previously, uxdot-pattern used an SSR controller (UxdotPatternSSRController) that read pattern source files at render time and injected them via unsafeHTML. This created two nested layers of Lit SSR: the outer uxdot-pattern element and the inner pattern content -- which couldn't hydrate together reliably.

This PR moves all of that work out of the component and into two new Eleventy plugins:

- `uxdot-pattern-inline` -- An Eleventy transform that runs before SSR. For each <uxdot-pattern src="..."> in the rendered HTML, it reads the pattern file, extracts styles/scripts/links, builds a live demo wrapper (slot="content"), syntax-highlights the separated HTML/CSS/JS sources, and appends them as slotted <rh-code-block> children (slot="html-source|css-source|js-source").
- `markdown-it-block-elements` -- Registers custom elements (from elements/ and uxdot/) as block-level HTML tags for markdown-it by scanning their CSS for :host { display: block|flex|grid|contents }. This prevents markdown-it from wrapping block-level custom elements in `<p>` tags, which would break DOM structure.

The uxdot-pattern component itself is simplified significantly:

- Now extends LitElement directly instead of the custom SSRFailureRecoverableElement
- Removes src, css-src, and js-src properties (no longer fetches its own content)
- Uses SlotController to detect slotted content and conditionally disable CSS/JS tabs
- Renders named slots instead of unsafeHTML directives
- The #content container now allows resize: both instead of only resize: vertical

#### Deleted files:

- uxdot/ssr-failure-recoverable.ts
- uxdot/uxdot-pattern-ssr-controller.ts
- uxdot/uxdot-pattern-ssr-controller-server.ts
- uxdot/uxdot-pattern-ssr-controller-client.ts

### Motivation
Eliminates the dual-SSR hydration mismatch by ensuring all pattern content is already present in the HTML as light DOM slots before Lit SSR processes the page. This makes uxdot-pattern a straightforward slotting component with no server-side data fetching of its own.

This is a convenience otherwise we can write out the uxdot-patterns manually, although having the sources done automatically is a nice win.

```html
<uxdot-pattern>
  <div slot="content">
    <!-- your live demo markup here -->
    <rh-alert>Hello world</rh-alert>
  </div>
  <rh-code-block slot="html-source" highlighting="prerendered">
    <!-- pre-highlighted HTML source -->
  </rh-code-block>
  <rh-code-block slot="css-source">
    /* your CSS */
  </rh-code-block>
  <rh-code-block slot="js-source">
    // your JS
  </rh-code-block>
</uxdot-pattern>
```

## Testing Instructions

1. Review all patterns pages


zeroedin:
Replacing with a clean PR rebased onto main (the original branch carried staging/growlithe history).

PR #2956, opened 2026-04-16
docs(uxdot-pattern): fix SSR hydration bugs

zeroedin:
## What I did

Replaces uxdot-pattern's Lit SSR controller architecture with a build-time Eleventy transform that inlines pattern content as slotted light DOM children.

Previously, uxdot-pattern used an SSR controller (UxdotPatternSSRController) that read pattern source files at render time and injected them via unsafeHTML. This created two nested layers of Lit SSR: the outer uxdot-pattern element and the inner pattern content -- which couldn't hydrate together reliably.

This PR moves all of that work out of the component and into two new Eleventy plugins:

- `uxdot-pattern-inline` -- An Eleventy transform that runs before SSR. For each <uxdot-pattern src="..."> in the rendered HTML, it reads the pattern file, extracts styles/scripts/links, builds a live demo wrapper (slot="content"), syntax-highlights the separated HTML/CSS/JS sources, and appends them as slotted <rh-code-block> children (slot="html-source|css-source|js-source").
- `markdown-it-block-elements` -- Registers custom elements (from elements/ and uxdot/) as block-level HTML tags for markdown-it by scanning their CSS for :host { display: block|flex|grid|contents }. This prevents markdown-it from wrapping block-level custom elements in `<p>` tags, which would break DOM structure.

The uxdot-pattern component itself is simplified significantly:

- Now extends LitElement directly instead of the custom SSRFailureRecoverableElement
- Removes src, css-src, and js-src properties (no longer fetches its own content)
- Uses SlotController to detect slotted content and conditionally disable CSS/JS tabs
- Renders named slots instead of unsafeHTML directives
- The #content container now allows resize: both instead of only resize: vertical

#### Deleted files:

- uxdot/ssr-failure-recoverable.ts
- uxdot/uxdot-pattern-ssr-controller.ts
- uxdot/uxdot-pattern-ssr-controller-server.ts
- uxdot/uxdot-pattern-ssr-controller-client.ts

### Motivation
Eliminates the dual-SSR hydration mismatch by ensuring all pattern content is already present in the HTML as light DOM slots before Lit SSR processes the page. This makes uxdot-pattern a straightforward slotting component with no server-side data fetching of its own.

This is a convenience otherwise we can write out the uxdot-patterns manually, although having the sources done automatically is a nice win.

```html
<uxdot-pattern>
  <div slot="content">
    <!-- your live demo markup here -->
    <rh-alert>Hello world</rh-alert>
  </div>
  <rh-code-block slot="html-source" highlighting="prerendered">
    <!-- pre-highlighted HTML source -->
  </rh-code-block>
  <rh-code-block slot="css-source">
    /* your CSS */
  </rh-code-block>
  <rh-code-block slot="js-source">
    // your JS
  </rh-code-block>
</uxdot-pattern>
```

## Testing Instructions

1. [Review all patterns examples pages](https://deploy-preview-2956--red-hat-design-system.netlify.app/patterns/)

## Notes to Reviewers


zeroedin:
Looks like there is some issues just identified on tabs/examples in the js, properities of null (reading 'closest') and prism.  

https://deploy-preview-2956--red-hat-design-system.netlify.app/patterns/tabs/examples/#south

Update: some of which preexisted.
- prism language
- preexisting issue with code tabs css showing up in the code tab example.

new: 
- properties of null (reading 'closest')

zeroedin:
image
vs 
image


Looks like the markdown-it-elements is being a bit too aggressive, capturing `<rh-alert>` when used with backticks.  Might have to look at another way of doing that parsing.

bennypowers:
We use  ssrcontroller for RHDS elements as well

Rather than introduce a new rendering paradigm for a subset of elements, can we fix the problems with ssrcontroller?

zeroedin:
> We use ssrcontroller for RHDS elements as well
> 
> Rather than introduce a new rendering paradigm for a subset of elements, can we fix the problems with ssrcontroller?

@bennypowers you'll have to explain, I actually deleted `uxdot-pattern-ssr-controller` so how is this a new rendering paradigm, to me it's now closer to the paradigm then it was before.  All this does is handle chopping up the element for you.

IMHO why did uxdot-pattern need special treatment in the first place?

bennypowers:
🤔 you drive a hard bargain 😃

I'll take a closer look next week

docs/_plugins/uxdot-pattern-inline.ts
```diff
@@ -0,0 +1,314 @@
...
+  element: Element,
+  inputPath: string,
+): Promise<void> {
+  const src = getAttribute(element, 'src');
+  if (!src) {
+    return;
+  }
+
+  const cssSrc = getAttribute(element, 'css-src');
+  const jsSrc = getAttribute(element, 'js-src');
```
adamjohnson, line 167: We should probably check for code block's `full-height` attr here too:

```js
const fullHeightAttr = getAttribute(element, 'full-height') !== null ? ' full-height' : '';
```

Then use it below in various places:

```js
<rh-code-block ${fullHeightAttr}>
```

Otherwise this attribute doesn't get output when it should.

docs/_plugins/uxdot-pattern-inline.ts
```diff
@@ -0,0 +1,314 @@
...
+  if (cssContent) {
+    contentHtml += `<style>${cssContent}</style>`;
+  }
+  contentHtml += htmlContent;
+
+  const contentFragment = parseFragment(contentHtml);
+  for (const child of [...contentFragment.childNodes]) {
+    if (isElementNode(child)) {
+      setAttribute(child, 'slot', 'content');
+    }
```
adamjohnson: What about text nodes? They may be appended unslotted and leak into the default slot. Consider wrapping text nodes in `<div slot="content">`.

docs/_plugins/rhds.ts
```diff
@@ -12,6 +12,7 @@ import slugify from 'slugify';
 import RHDSAlphabetizeTagsPlugin from '#11ty-plugins/alphabetize-tags.js';
 import RHDSElementDocsPlugin from '#11ty-plugins/element-docs.js';
 import RHDSElementDemosPlugin from '#11ty-plugins/element-demos.js';
+import UxdotPatternInlinePlugin from '#11ty-plugins/uxdot-pattern-inline.js';
```
adamjohnson: Is this being used? It's being flagged as a dead import.

adamjohnson (changes requested):
1. Alert on [Guidelines page of Logo Wall](https://deploy-preview-2956--red-hat-design-system.netlify.app/patterns/logo-wall/guidelines/#art-direction) busted.
    * This will likely be fixed alongside the `<rh-tile>` alert you already know about.
1. [Examples page](https://deploy-preview-2956--red-hat-design-system.netlify.app/patterns/logo-wall/examples/) of Logo Wall shows two images simultaneously for each logo (it should only show one)
    * Looks like this is an issue with logo-wall-lightdom.css, the surface no longer has a `color-palette` attribute:
        * `[color-palette^="light"] > .logo-wall-container .logo-wall--dark { display: none; }`
1. Line numbers on all pattern pages that use `<rh-code-block>` are missing ([example](https://deploy-preview-2956--red-hat-design-system.netlify.app/patterns/call-to-action/examples/))
    * This may be due to the missing `isIntersecting` class / IntersectionObserver not firing.

docs/_plugins/rhds.ts
zeroedin: fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/2956/commits/08d619887fa2a4ebcebcdad2c06c0a459feb0a98

docs/_plugins/uxdot-pattern-inline.ts
zeroedin: fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/2956/commits/08d619887fa2a4ebcebcdad2c06c0a459feb0a98

docs/_plugins/uxdot-pattern-inline.ts
zeroedin: fixed in https://github.com/RedHat-UX/red-hat-design-system/pull/2956/commits/08d619887fa2a4ebcebcdad2c06c0a459feb0a98

zeroedin:
After office hours tried to unpack the comment about attempting async from the server to do this task.  Here is an excerpt from my claude conversation.

Yes, ThunkedRenderResult fully supports async work on the server — and in fact, the code you just deleted on this branch was already doing exactly that. The infrastructure is still in place in `worker.ts`.

**How it works today**

Your `RHDSSSRableRenderer.renderShadow()` already returns a `ThunkedRenderResult` with an async thunk:

```js
  override renderShadow(renderInfo: RenderInfo): ThunkedRenderResult {
    return [
      // Render styles.
      this.#renderStyles(),
      // Thunk that sets up controllers first, then renders template
      async () => {
        for (const controller of this.getControllers()) {
          if (controller.ssrSetup) {
            await controller.ssrSetup(renderInfo);
          }
        }
        return renderValue(
          // @ts-expect-error: if upstream can do it, so can we
          this.element.render(),
          renderInfo,
        );
      },
    ];
  }
  ```

The Thunk type explicitly allows returning a Promise:

```js
export type Thunk = () => void | string | ThunkedRenderResult | Promise<string | ThunkedRenderResult>;
```

And collectResult (which renderPage calls) handles it by awaiting Promises in its inner loop. So you could absolutely bring back an RHDSSSRController with an `ssrSetup()` method that does async file I/O, syntax highlighting, etc. — the rendering pipeline will await it before calling `this.element.render()`.

**Why you moved away from it**

The problem was never the server-side async — that part worked fine. The problem was the client-side hydration that had to match. The old approach created a gnarly cycle:

Server: `ssrSetup()` populated `this.allContent` / `this.htmlContent` etc. with `unsafeHTML()` directives, then `render()` emitted them into the shadow DOM.
Client: `UxdotPatternSSRControllerClient` had to reverse-engineer those values from the already-rendered shadow DOM before hydration ran, because `noChange` sentinel values caused "Primitive found where TemplateResult expected" hydration errors.

Recovery: `SSRFailureRecoverableElement` caught Hydration errors in update() and tried to forceRender + forceHydration — removing duplicate #container elements, stripping defer-hydration, and force-cycling every reactive property.

That's ~230 lines of workaround code just to bridge the server/client gap for `unsafeHTML`.

**Could you do it better with async thunks?**

You could, but the fundamental tension remains: any content injected via `unsafeHTML` in an SSR controller will be opaque to the Lit hydration system. The hydration protocol expects the client-side template to structurally match what the server emitted. `unsafeHTML` is inherently non-deterministic from the hydrator's perspective — the client can't reconstruct the same directive result without repeating the same async work (reading files, highlighting code), which you obviously can't do in the browser.

The cleanest way to use async thunks without hydration issues would be to have the async thunk produce content that lands in a place the client doesn't need to hydrate — but that's essentially what your new approach already does. By inlining as slotted light DOM, the content sits outside the shadow root that Lit hydrates. The component just has `<slot name="content">` and `<slot name="html-source">`, which are structurally identical on server and client. No mismatch, no recovery needed.


### TLDR; Takeaway

The problem we are trying to solve for here is the dynamic shadowroot content.  

The way the code works outside this branch is we are trying to shove dynamic content into the shadowroot and not use `<slot>`.  However that can't be hydrated without errors on the client because of the mismatch of server vs client known state, then we were trying to "recover" from those errors. However that doesn't always seem to work as @adamjohnson found out with `rh-alert` opting instead to link to the demo instead of creating a real pattern using `<uxdot-pattern>` on the patterns page.  I also ran into that same issue with `<rh-drawer>` patterns.  This was the result of trying to figure out why.

The other ask in the office hours conversation was to think about encapsulation of the pattern content.  But I question what we are trying to actually solve for/ fix with that?  Is that a real problem to be concerned with.

We own the site these patterns are being shown on.  These patterns wouldn't be constructed via a uxdot-pattern component on another site, there is a reason they are called `uxdot` components.  We can and should control any style leakage into the patterns using other methods available to us on the site.  I personally feel like the added code caused more problems than it solved and the changes in this branch are more appropriate and give us the end result we want.

bennypowers:
there isn't any state in the interpolated content, there's nothing to hydrate there other than the HTML which is already there

zeroedin:
> there isn't any state in the interpolated content, there's nothing to hydrate there other than the HTML which is already there

again an AI assisted response:

> The content is stateless, but the hydration protocol is all-or-nothing per shadow root. You can't tell Lit "hydrate the tabs and color picker, but skip the unsafeHTML parts." A mismatch anywhere poisons the whole tree. So while the observation is technically correct — the content itself doesn't need hydration — it's irrelevant because the mechanism doesn't allow you to be selective about it.

It just feels like we are fighting against how the Lit SSR works, and I'm not sure of the real reason / problem we are trying to solve here.  I need more of a reason on the why not just do what this branch is doing.

bennypowers:
Thanks for identifying this issue and putting together a fix, Steven. The hydration errors on pattern pages have been a real annoyance, and the previous client-side extraction workarounds were genuinely brittle -- good call investigating a better approach.

I pushed a different approach to this branch (reverted your commits and added a smaller fix).

**TL;DR**: The hydration mismatch is a digest problem in the SSR controller, not a deep-shadow-root problem. The fix is a small `ssrAdopt` directive that patches the marker digest so Lit's hydration walker adopts the existing DOM. ~120 lines added, ~150 deleted, no new build plugins. Server-side concerns (filesystem reads, highlighting) stay server-side, SSR concerns stay in SSR files, no duplication of SSR stuff in eleventy; lit problems are handled with lit APIs.

<details>
<summary>On the root cause</summary>

The problem isn't that Lit SSR can't handle deep/nested shadow roots -- that's its primary design goal, and it works. The actual issue is a server/client data asymmetry in the SSR controller. The server uses `unsafeHTML(content)` to inject pattern HTML, which internally impersonates a `TemplateResult` with a digest derived from the content string. On the client, that exact string is unrecoverable: browser HTML serialization differs from parse5's `serialize()`, and SSR adds `defer-hydration` attributes. Any mismatch breaks the digest check and throws "Unexpected TemplateResult rendered to part." The `SSRFailureRecoverableElement` catch-and-re-render was papering over this.
</details>

<details>
<summary>On the scope of the original changes</summary>

The 11ty transform approach (~470 new lines, two new plugins) moved the work out of the SSR controller into the build pipeline. This duplicated content-reading and highlighting logic that the server SSR controller already handles, and introduced a markdown-it block-elements plugin that's tangential to the hydration bug. The component also lost its `src`/`css-src`/`js-src` attributes, becoming dependent on the transform.
</details>

<details>
<summary>What the new fix does</summary>

A custom `ssrAdopt` directive (~120 lines, one new file) tells the hydration walker to adopt existing SSR'd DOM without re-rendering. It creates a no-op `TemplateResult` and patches the SSR comment marker's digest to match, so the walker enters a `template-instance` with no parts and skips all nested markers. On re-renders, it returns `noChange` to preserve the content. This lets us delete `SSRFailureRecoverableElement` and all the client controller's extraction/cleanup workarounds.
</details>

<details>
<summary>Tradeoffs</summary>

The directive patches `Comment.data` on the SSR marker, which relies on Lit's hydration walker reading `marker.data` after `resolveDirective` returns (current behavior, but not contractually guaranteed). The fake `TemplateResult` mirrors what `unsafeHTML` itself does internally. Both coupling points are documented in `ssr-adopt-directive.ts`. `digestForTemplateResult` is a public export from `@lit-labs/ssr-client`; `ChildPart.startNode` is a public getter on `lit-html`.
</details>

bennypowers (approved):

PR #2971, opened 2026-04-22
fix(tabs): corrected focus selector

### LINKED ISSUE #2970: [bug] `<rh-tabs>`: panel `:focus` and `:focus-within` causing multiple concurrent outlines
Our tab panels currently receive focus outlines when:

1. The panel itself has keyboard focus (good)
2. When clicked within via the mouse ([not necessary](https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html), but maybe we're doing it deliberately?)
3. When any other element within has keyboard focus (bad).

 
I think issue 3 above should be fixed, so only one item on the page has a visible focus state at a time. And we can discuss 2.

Here's a screencap of dual focus. Note the outline on both the panel and the link within.

Tab panel with two focus outlines

[Example page here](https://www.redhat.com/en/about/brand/standards/color#how-we-use-color)

It can get even more complicated if someone is using tabs within tabs, which I guess has been known to happen!

Looks like the relevant code is in [rh-tab-panel.css](https://github.com/RedHat-UX/red-hat-design-system/blob/9d6ee86c59aa1052e3d767d7015d030b2d095987/elements/rh-tabs/rh-tab-panel.css#L12).

Potential solution: replace `:is(:focus, :focus-within)` with `:focus-visible`, unless there's a reason we prefer `:focus` to `:focus-visible`.

zeroedin:
## What I did

1. Improved accessibility so only one focus outline is visible at a time when navigating within a tab panel

Closes #2970 

## Testing Instructions

1. View [Deploy Preview Tabs Demo](https://deploy-preview-2971--red-hat-design-system.netlify.app/elements/tabs/demo/) tab to containers tab activate then tab the "Focusable element" link only 1 focus ring should be shown

### DIFF
--- a/.changeset/loose-years-sip.md
+++ b/.changeset/loose-years-sip.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tabs>`: improved accessibility so only one focus outline is visible at a time when navigating within a tab panel
+  
--- a/elements/rh-tabs/rh-tab-panel.css
+++ b/elements/rh-tabs/rh-tab-panel.css
@@ -11,7 +11,7 @@
 
 :host(:focus-visible) {
   /** Focus ring, design token rh-color-interactive-primary */
-  outline: 1px auto var(--rh-color-interactive-primary-default);
+  outline: var(--rh-border-width-lg, 3px) auto var(--rh-color-interactive-primary-default);
 }
 
 #container {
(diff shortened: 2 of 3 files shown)


bennypowers (approved):

hellogreg (changes requested):

hellogreg (approved):
line glitch triggered multiples

adamjohnson (approved):
Less Glare, Tabs Managed

PR #2951, opened 2026-04-15
docs(demos): use YAML frontmatter for demo metadata

bennypowers:
## Summary

- Replace `<meta itemprop="description">` tags with YAML frontmatter `description:` fields in all 382 demo HTML files (including 15 previously undocumented demos)
- Port 4 companion `.md` files to a `doc:` frontmatter field and delete the originals
- Update the 11ty build to strip frontmatter from demo pages: add a `renderDemoFile` shortcode, render `demo.description` from the CEM manifest, and parse the `doc` field for supplementary rich content
- Bump `@pwrs/cem` to ^0.9.19 for frontmatter support

Closes #2949

## Test plan

- [ ] Run `npx cem generate` and verify `custom-elements.json` demos have `description` fields populated from frontmatter
- [ ] Run `npx tsc --noEmit -p docs/tsconfig.json` and verify no type errors
- [ ] Build the docs site and verify fullscreen demo pages (`/elements/*/demo/`) render without visible frontmatter text
- [ ] Verify demos page (`/elements/*/demos/`) shows descriptions and supplementary `doc` content (e.g. toast, nested nav, anchor links)
- [ ] Spot-check code blocks on demos pages to confirm frontmatter is not included in the HTML tab

Assisted-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

PR #2983, opened 2026-04-29
docs: update fonts to use variable fonts and modern css syntax

zeroedin:
## What I did

1. Update the fonts style sheet to use updated variable fonts and modern css syntax
2. Updated the guidance example with this same modern css syntax linking to the new fonts.

## Testing Instructions

1.  Check for correct font rendering in the Deploy Preview.


.stylelintrc.yml
```diff
@@ -38,6 +38,9 @@ rules:
   # Avoid conflicts with valid descriptors in @font-face (e.g., text-rendering)
   at-rule-descriptor-no-unknown: null
 
+  # The tech() function in @font-face src is valid modern CSS but not yet recognized
```
bennypowers: can we bump stylelint? are there lots of breaking changes?

marionnegp (changes requested):
I think the 600 font weight is still rounding up to 700.

Here's what they should look like, based on Google Fonts:
Image

And here's what it looks like if I try to use each weight on ux dot:
Image

.stylelintrc.yml
zeroedin: that's a good question, something I can look into tomorrow.

zeroedin:
> I think the 600 font weight is still rounding up to 700.

@marionnegp hey there was some syntax issues that was actually causing the fonts not to load will you try again and let me know if you still see the issue?

docs/patterns/side-navigation/patterns/layout-flow.html
```diff
@@ -0,0 +1,150 @@
+<meta itemprop="description" content="The panel participates in the parent layout at wide viewports, sitting in a CSS grid as a navigation sidebar. Uses overlay-threshold=&quot;md&quot; to set the breakpoint at 992px. Below the threshold it becomes a fixed overlay toggled by an external trigger. The default threshold for flow is 768px when not set. This demo uses rh-navigation-vertical to show a realistic navigation use case.">
```
bennypowers: #2951

zeroedin: ugh, was trying to teach myself worktrees yesterday obviously got somethign messed up 👎

.stylelintrc.yml
zeroedin: @bennypowers So after quick research. we'd likely want to do some updating in @rhds/tokens to ensure we remove any *.cjs plugins as 17+ drops cjs support.  Other then that looks fairly trivial.  I'll open an issue on @rhds/tokens to update there first to get the ball rolling.

zeroedin: https://github.com/RedHat-UX/red-hat-design-tokens/issues/187

Might not need to drop the .cjs just work around it, maybe give it an official deprecation?

marionnegp (approved):
Thanks, @zeroedin!

bennypowers (approved):

.stylelintrc.yml
bennypowers: better to use stylelint-ignore comments for now then

PR #2827, opened 2026-02-10
epic: adding Unified theme page and CSS

markcaron:
## What I did

1. Fixed `npm run start` errors in Lit workers that exists in `main` (not sure why)
2. Added Themes subsection to Theming
3. Added Unified theme page
4. Added `unified-theme-preview.css`

## Testing Instructions

Check each sub PR below.

## Sub-PRs (per-element)

- [x] #2986 — `rh-accordion`
- [x] #2987 — `rh-alert`
- [x] #2988 — `rh-badge`
- [x] #3001 — `rh-switch`
- [x] #2989 — `rh-breadcrumb`
- [x] #2990 — `rh-button`
- [x] #2991 — `rh-card`
- [x] #2992 — `rh-chip`
- [x] #2993 — `rh-cta`
- [x] #2998 — `rh-progress-stepper`
- [x] #2999 — `rh-spinner`
- [x] #3003 — `rh-tag`
- [x] #2994 — `rh-dialog`
- [x] #2996 — `rh-menu`
- [x] #3002 — `rh-tabs`
- [x] #2997 — `rh-navigation-vertical`
- [x] #2995 — `rh-jump-links`
- [x] #3000 — `rh-subnav`

PR #2986, opened 2026-04-30
docs: Unified Theme rh-accordion

markcaron:
## Summary
Unified Theme work for `rh-accordion`.

Sub-item of #2827.


marionnegp (changes requested):
@markcaron, if you need to double check specs, here's are the [accordion specs](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.0?m=dev&node-id=4107-8303&t=WNDgxCQXA1KfrQ3h-1) for our Figma updates.

- [x] Focus states should have light gray background color and bg opacity used in hover state.
- [x] Bordered version shouldn't have a border at the top of the first item.
- [x] For an expanded `rh-accordion-panel`, inline padding for both sizes and variants (bordered/borderless) should be `var(--rh-space-lg, 16px)`. I'm seeing 24px on left and right. The bottom padding should be `var(--rh-space-xl, 24px)`. I think the Figma file you used had funky padding values. These changes should help align the content with the `rh-accordion-header`'s text and caret.
- [x] Panel shouldn’t show right border when expanded.
￼Image
- [x] Should the default variant be renamed to `borderless`?
- [ ] Should the caret icon use `rh-icon`? I see it's using an SVG in the base element though.
- [x] Here's the padding for `rh-accordion-header`: Default/small size should use `padding: 8px 16px`. Large size should use `padding: 16px`.

- Checking with @coreyvickery  about the spacing between the header text and the caret. PF has it at 4px for all sizes, which seems small to me.

marionnegp:
@coreyvickery, what are your thoughts on making the large variant of most (if not all) elements be the default? This will be different from what PF does because they need more content density.

marionnegp (changes requested):
Forgot to check the dark scheme accordion in my previous review.

**Dark scheme bugs:**
- [x] The border color for the bordered variant should be gray-60.
- [x] The header gets some funky styling on hover when it's expanded or collapsed. The corners are sharp and the background color gets darker when it shouldn't.
Image
Image
- [ ] In active state, I'm seeing blue borders appear. This is also what the focus state looks like, with a thin focus ring being overlapped by sharp corners.
Image
- [ ] Caret icon should be using `gray-10`.

**Both color schemes:**
- [ ] Not sure if some site styles are bleeding into the demo, but the font of the header text looks too bold. Should be medium weight.

markcaron:
> * [ ]  Should the caret icon use `rh-icon`? I see it's using an SVG in the base element though.

@marionnegp Probably at some point, but that would require a separate code change. I don't think that's a blocker, but a separate issue.

marionnegp (changes requested):
- [x] `block-padding-end` for `rh-accordion-panel` should use `--rh-space-xl`. (I had this wrong in my Figma file.)
- [x] `block-padding-start` for `rh-accordion-panel` should use `--rh-space-md`.
- [x] Light gray border for the bordered variants in light scheme should be `gray-30`. I'm seeing `--rh-color-gray-20`.
- [x] I was mistaken about the dark scheme's caret color. It should be white, so `--rh-color-icon-secondary` is ok to use for the caret.
- [x] Background color for `rh-accordion-panel` should be `--rh-color-surface-lightest` (light scheme) or `gray-80` (dark scheme).

bennypowers (approved):

PR #2988, opened 2026-04-30
docs: Unified Theme rh-badge

markcaron:
## Summary
Unified Theme work for `rh-badge`.

Sub-item of #2827.


markcaron:
**Blocked:**
Seems that Badge may have been overlooked in the design. Hoping this won't be conflated with PF6's Label component, which is already conflated with the Chip UI pattern. Three completely different semantic uses.

marionnegp:
@markcaron, yeah, I'm keeping this the same until we can sync with the PF team. We have similar styles but different color options.

bennypowers:
Closed pending DPO level design review

PR #2989, opened 2026-04-30
docs: Unified Theme rh-breadcrumb

markcaron:
## Summary
Unified Theme work for `rh-breadcrumb`.

Sub-item of #2827.


adamjohnson (commented):
Regarding Qodo's reivew:

1. This is correct, but the design specs don't specify focus or visited states. Personally, I would rather have the visited state of rh-breadcrumbs persist (absent of other direction).
2. This suggestion is valid, but it seems like reordering the stylesheets + removing important forces us to not use `@layer` for the unified breadcrumbs theme, which contradicts our existing organizational pattern using layers. So, probably moot.

Overall, this PR looks pretty good. We should decide on `#1` and then LGTM.

markcaron:
@adamjohnson I'm ok with the visited = hover color. But perhaps @marionnegp has a different understanding/plan?

marionnegp:
> @adamjohnson I'm ok with the visited = hover color. But perhaps @marionnegp has a different understanding/plan?

Yeah, I agree that our visited state can persist. Our `interactive-primary-visited` token is aligned with PF's, so I'd use the `interactive-primary-visited-*` tokens we already have.

docs/theming/themes/unified-theme/unified-theme-preview.css
```diff
@@ -1218,8 +1218,41 @@
...
     --rh-border-radius-default: var(--unified-border-radius-action-plain);
     --rh-color-border-interactive: var(--unified-color-interactive-primary-default);
+    --rh-font-size-body-text-sm: var(--unified-font-size-body-text-sm);
+    /* stylelint-disable @stylistic/declaration-colon-newline-after, @stylistic/max-line-length */
+    --rh-breadcrumb-caret-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 5 9' fill='none'%3E%3Cpath d='M.5 9a.5.5 0 0 1-.354-.146.5.5 0 0 1 0-.708L3.793 4.5.146.854a.5.5 0 0 1 0-.708.5.5 0 0 1 .708 0l3.823 3.823a.75.75 0 0 1 0 1.061L.854 8.854A.5.5 0 0 1 .5 9Z' fill='%23151515'/%3E%3C/svg%3E");
+    /* stylelint-enable @stylistic/declaration-colon-newline-after, @stylistic/max-line-length */
+
+    & > ol > li > a[href]:not([aria-current]) {
+      text-decoration: underline var(--unified-border-width-sm) dashed !important;
+      text-decoration-color: #8c8c8c !important;
```
marionnegp: `#8c8c8c` is now `gray-45` in our [tokens](https://ux.redhat.com/tokens/color/#color-gray).

marionnegp (changes requested):
- [x] I think the `text-decoration-color` should be the same color as the the link text, which is how the hover state for an unvisited link works anyway.
- [x] caret icons should use `--unified-color-icon-secondary`
- [x]  For links that have not been visited, the text color should be using `--unified-color-text-interactive-*`. I'm seeing our regular interactive token colors being used.

docs/theming/themes/unified-theme/unified-theme-preview.css
markcaron: Benny replaced that with `text-decoration-color: currentcolor !important;` — which will always match the `text-decoration-color` with the text `color` per:

> I think the text-decoration-color should be the same color as the the link text, which is how the hover state for an unvisited link works anyway.

Do we need `gray-45` here then, if it's blue or purple?

marionnegp (approved):

bennypowers (approved):

PR #2990, opened 2026-04-30
docs: Unified Theme rh-button

markcaron:
## Summary
Unified Theme work for `rh-button`.

Sub-item of #2827.

Figma links: [Light scheme](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.3?m=dev&node-id=4114-14776) | [Dark scheme](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.3?m=dev&node-id=10572-7139)

Note: The Project Felt-themed buttons from those files are highlighted in the screenshot's purple boxes.
Screenshot 2026-06-12 at 4 14 36 PM


marionnegp (changes requested):
Apologies in advance for not using semantic token names. :face_with_peeking_eye:

**General**

- [ ] The focus ring doesn't go away after you've clicked. I don't think we need a focus ring for the active state anyway.
Image

- [x] The flash of blue that appears for the active state doesn't abide by the border radius. I'm seeing sharp corners in the blue overlay when I click. (Edit (6/22): I'm not seeing this anymore!) 

- [x] The active state animation doesn't quite match up with what PF's doing. They have some info about their [button ripple micro animation](https://www.patternfly.org/design-foundations/motion/#button-ripple) on their Motion page.

**Light Scheme**

- [x] Active state primary button in light scheme is using dark blue text. Should use white.
Image

**Dark Scheme**

- [x] Text color for primary buttons and primary danger buttons in dark scheme should be `gray-90`.
- [x] Tertiary button in dark scheme: Text color in default state should be `blue-20`. Hover, focus, and active states should have a `blue-10` text color.
- [x] Link button in dark scheme: Text color in default state should be `blue-20`. Icon and text colors should be `blue-10` in hover, active, and focus states.
- [x] Secondary/tertiary button in dark scheme: Text color in hover, active, and focus states should be `blue-10`.
- [x] Link danger button in dark scheme: Text in hover, active, and focus states should be `red-orange-20`. Icon in default state should be `--rh-color-red-orange-50`. In other states, icon should be `--rh-color-red-orange-40`.
- [x] Secondary danger and tertiary danger buttons in dark scheme: Icon color in default state should be `red-orange-50`.

**Compact Variant**
- [x] Block padding for compact button should be 4px.
- [x] Need danger primary, danger secondary, and link danger buttons in compact too

**Link Variant**

- [x] Seeing uneven padding for the link-style button when an icon is added. I think it's compensating for padding that was around the icon in the Figma mockups, but I don't think we need it here because there's no additional inline padding.
- [x] Link buttons have less inline padding than the other variants. Large sizes use 8px all around. Small sizes use 4px all around.

**Close Variant**

- [x] Just to keep some options in each variant similar, can there be a large close button? I have one in [this Figma mockup](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.0?m=dev&node-id=4114-14776).

markcaron:
@marionnegp

> The focus ring doesn't go away after you've clicked. I don't think we need a focus ring for the active state anyway.

AFAIK, this exists on `<rh-button>` today. Will likely need a element CSS change from `:focus` to `:focus-visible`

> The flash of blue that appears for the active state doesn't abide by the border radius. I'm seeing sharp corners in the blue overlay when I click.

I'll need you to show me. I can't reproduce.

markcaron:
@marionnegp per the icon hover/active/focus color updates:

> The request that link danger icon in hover/active/focus be `red-orange-40` (while text is `red-orange-20`) is not achievable through external CSS alone. `rh-icon` inherits `color: currentcolor` from the button, and there's no way to intercept that per-state from outside the shadow DOM using `::part()` — `::part(icon):hover` only fires when the icon itself is hovered, not when the button is. Addressing this would require adding a dedicated CSS custom property (e.g. `--rh-button-icon-color`) to the component's element-level CSS.

marionnegp:
> @marionnegp per the icon hover/active/focus color updates:
> 
> > The request that link danger icon in hover/active/focus be `red-orange-40` (while text is `red-orange-20`) is not achievable through external CSS alone. `rh-icon` inherits `color: currentcolor` from the button, and there's no way to intercept that per-state from outside the shadow DOM using `::part()` — `::part(icon):hover` only fires when the icon itself is hovered, not when the button is. Addressing this would require adding a dedicated CSS custom property (e.g. `--rh-button-icon-color`) to the component's element-level CSS.

@markcaron, is this something that we can tackle in a patch or future release?

docs/theming/themes/unified-theme/unified-theme-preview.css
```diff
@@ -632,10 +693,24 @@
...
+          /* Dark scheme: icon default is red-orange-50;
+             light scheme unchanged (red-orange-60 = text color) */
+          color: var(--unified-color-status-danger);
+        }
+
+        /* Set --rh-button-icon-color on the host so it reliably inherits
+           into the shadow DOM; ::part() custom props may not cascade to
+           shadow-DOM descendants of the targeted part. */
+        &:is(:hover, :focus-within, :active) {
+          --rh-button-icon-color: var(--unified-color-icon-status-danger-hover);
```
adamjohnson, line 759: > The request that link danger icon in hover/active/focus be red-orange-40 (while text is red-orange-20) is not achievable through external CSS alone. 

This is the definition for `--unified-color-icon-status-danger-hover` in the `unified-theme-preview.css`:

```css
    --unified-color-icon-status-danger-hover:
      light-dark(
          var(--unified-color-red-orange-70),
          var(--unified-color-red-orange-40)
        );
```

Is it supposed to be `--unified-color-red-orange-70` in light and `--unified-color-red-orange-40` in dark? This is how it works currently. 

Or are those values supposed to be swapped?

marionnegp: Yeah, the current color definition is right.

marionnegp:
@adamjohnson @markcaron, colors for the secondary and tertiary variants' interactive states weren't switching to the correct colors. Let me know if any of the CSS looks weird to you. For example, in lines 625-629 for the secondary danger variant, do I need `--rh-color-text-primary: var(--unified-color-text-status-danger-hover);` if `--rh-color-status-danger: var(--unified-color-text-status-danger-hover);` seems to work?

adamjohnson:
> do I need... if ...seems to work?

If it works, it works. 

I suppose you could make an argument for explicitly setting it for clarity. But then you'd probably have to repeat it for other variants, causing the code to be less [dry](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

IMO if it works, leave it be.

marionnegp (approved):

docs/theming/themes/unified.md
adamjohnson, line 1: Now that #3048 has merged, do these changes need to be removed?

adamjohnson (approved):
LGTM, but check out my comment below before merging.

docs/theming/themes/unified.md
markcaron: Removed button patch and accordion patch (since it was hanging out unnecessarily)

PR #2991, opened 2026-04-30
docs: Unified Theme rh-card

markcaron:
## Summary
Unified Theme work for `rh-card`.

Sub-item of #2827.


adamjohnson:
This PR has `Files changed: 0`. 

If there are changes, kindly push them up. I'm going to mark this as Draft until then.

marionnegp (changes requested):
@markcaron 

- [x] Can we add `plain` and `secondary` variants through the theme? Not sure how engineers would implement that. Here are those styles in [basic card](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.3?m=auto&node-id=10174-759&t=owhJd2MIlYKG0Fyd-1)

- [x] Here are the font size updates for each variant. I think right now the body text is set to inherit the font size, which _might_ introduce frustration if someone's trying to use a specific size variant.
**Compact**
Heading: 16px (no `font-size-heading` token for this though)
Body: `--rh-font-size-body-text-sm`
**Default**
Heading: `--rh-font-size-heading-xs` (Note: This differs from the 16px heading PF uses for their default-sized card's heading, but 16px seems too small for marketing purposes. cc: @coreyvickery)
Body: `--rh-font-size-body-text-md`
**Large**
Heading: Font size is already correct.
Body: `--rh-font-size-body-text-md`

- [x] Border colors for light scheme should be `--rh-color-gray-20`. Should be `--rh-color-gray-60` in dark. I'm calling this `--felt-color-border-subtle` in Figma (which has different values from `--felt-color-border-interactive-subtle-*`).

marionnegp (changes requested):
Made the following edits:

- Changed font sizes for body copy to 18px for large size, 16px for default, and 14px for compact.
- Changed heading size for large size to 24px
- Changed some cards in demo to secondary and plain variants

Other changes that I didn't make:
- [x] The slotted `<p>` seems to be adding space. I don't see this happening with our regular card. What's the best fix?

markcaron:
@marionnegp `<p>` tags were likely inheriting a default margin from the user agent stylesheet, since there we no typographic styles added to the unified demos. Since our element demos pull in UX dot's `styles.css`, I add those to the unified card demo. Should be good now.

marionnegp (approved):

PR #2992, opened 2026-04-30
docs: Unified Theme rh-chip

markcaron:
## Summary
Unified Theme work for `rh-chip`.

Sub-item of #2827.

[Figma link](https://www.figma.com/design/bdwYbLWL19VAvJ1tid4ngr/add-unified-components?node-id=3229-2068)


markcaron:
**Blocked:**
<rh-chip> looks and functions a bit differently than what's expected in the Unified mockups, which is based on PF6's Label component (which conflates the Chip UI pattern with simple "labels" or really Tags).

marionnegp:
@markcaron, I updated the icon sizes, padding, and space between the text and icon to align with unified theme styles for `rh-tag`. Here's the [Figma link](https://www.figma.com/design/bdwYbLWL19VAvJ1tid4ngr/add-unified-components?node-id=3229-2068). Otherwise, we can keep this the same for now.

bennypowers:
barring an element API change, we'll have to keep the gap between icon and text as-is, for now.

marionnegp (changes requested):
- [x] Can we get updates made to the small chip variant too?

markcaron:
There appears to be some animation issues when chips with 2+ words are selected

coreyvickery (changes requested):
@markcaron Everything looks good visually, but the `Clear all` chip action doesn't appear to do anything in the demo.

markcaron:
@coreyvickery the "Clear all" button failure seems to be an SSR issue unrelated to the Unified theme work. I filed a separate issue to fix

https://github.com/RedHat-UX/red-hat-design-system/issues/3072

PR #2994, opened 2026-04-30
docs: Unified Theme rh-dialog

markcaron:
## Summary
Unified Theme work for `rh-dialog`.

Sub-item of #2827.


docs/theming/themes/unified-theme/unified-theme-preview.css
```diff
@@ -1082,6 +1082,7 @@
   rh-dialog {
     --rh-border-radius-default: var(--unified-border-radius-large);
     --rh-color-icon-secondary: var(--unified-color-icon-secondary);
+    --rh-font-weight-body-text-regular: var(--rh-font-weight-body-text-medium, 500);
```
adamjohnson: Qodo's remarks about changing the font weight here are valid.

It achieves the correct effect for the header and primary button, but also causes the "Cancel" rh-button to be bold too. 

The only way to solve this may be to add classes to these elements to target them directly, but that might defeat the whole purpose of this stylesheet.

So, we have to decide how to approach this. Either go `body-text-medium` everywhere for rh-dialog or omit it and live with slightly less bolded text.

marionnegp:
@coreyvickery, I don't remember looking closely at dialog during the unified theme's design phase. Do you think the [specs from the shared Unified Theme file](https://www.figma.com/design/qRn3S225WwGHIdgGZX0LTC/Unified-Theme--Design-Tokens--Styles---Specs?node-id=10708-5866&m=dev) are good to go?

coreyvickery:
@marionnegp When you get back, seems aligned to me.

adamjohnson (dismissed):
Few things:

- [x] The [Figma](https://www.figma.com/design/qRn3S225WwGHIdgGZX0LTC/Unified-Theme--Design-Tokens--Styles---Specs?m=dev&node-id=10866-5200) (look for "Modal" in Pages) shows multiple different sizes. rh-dialog does have a [variant attribute](https://ux.redhat.com/elements/dialog/code/#rh-dialog-attributes-variant) that has different sizes. We should add these sizes to our unified theme demo.
- [x] After completing ☝️, the Figma shows different Dialog "Modal title" heading sizes. 
    * Small = 20px
    * Medium (default) = 24px
    * Large = 28px
    * Currently shows 18px
- [x] Consider adding a "No header" version to make sure it matches design specs
- [x] Looks like the "Confirm" and "Cancel" buttons now have a regular font weight (currently shows as bold)

docs/theming/themes/unified-theme/unified-theme-preview.css
adamjohnson: tl;dr over a month later, this line can be removed. Once removed, it should revert back to 400.

bennypowers (approved):
LGTM

coreyvickery (changes requested):
@markcaron At tablet and phone breakpoints, the `no-header` dialog window has padding on each side which looks nice. Can we add padding to the other sizes?

#### With padding ✅

Screenshot 2026-06-30 at 10 41 29 AM

#### No padding ❌

Screenshot 2026-06-30 at 10 41 42 AM

markcaron:
@coreyvickery the `variant` sizes are independent of `no-header`. These behave the same on our standard RHDS Dialog, and are not part of the Unified theme updates (none of the CSS added affects this).

`medium` and `large` force the width without padding.

image

coreyvickery (approved):
@markcaron Looks good!

I will create a separate issue for the padding.

PR #2996, opened 2026-04-30
docs: Unified Theme rh-menu-dropdown

markcaron:
## Summary

Unified Theme work for `rh-menu-dropdown`.

Sub-item of #2827. Tied to #3045.

See https://github.com/RedHat-UX/red-hat-design-system/pull/3045 for testing instructions. This requires cloning this branch/PR to your computer and taking the CSS from 3045.


adamjohnson:
@marionnegp (For when you are back from vacation)

Unified menu dropdown (aka menu) is ready for your review. What you see in this PR's DP is not accurate. You will have to follow the testing instructions in #3045 in order to see Unified subnav in all its glory (clone this branch/pr, paste in some CSS, start the server). Benny asked us to do it this way if we had to change any component code (which I did!).

If any of that sounds like too much, PM me and I can send you a temporary URL to test with. FYI I am out of the office Monday 6/22.

Let me know if you have any questions. Hope you had a great vacation!

marionnegp (commented):
- [x] Font size for toggle label should be 14px
- [x] Padding-block should 8px. I think this should bring the height of the toggle to 37px.
- [x] Border colors for interactive states should be using `--unified-color-border-interactive-subtle-*`
- [x] I have a question for the PF team about the token they're using for their menu toggle's disabled border color. There's also a discrepancy between the disabled menu toggle in the PF library and their docs site. Can we set the disabled state's border width to 0, so it at least matches what they have [here](https://www.patternfly.org/components/menus/menu-toggle#disabled-toggle)? (It would also match what we already have in regular RHDS.)
- [x] In dark scheme, the background color for all states (except for the disabled state) should be `color/gray/70`. I didn't see a unified token for this yet, but in PF's library, it's denoted as `global/background/color/control/default`.
- [x] The compact variant should use 37px for its height and width.
- [x] Border radius for compact should use `--rh-border-radius-pill`
- [x] A disabled compact toggle should have a transparent bg color. Icon color should be `--unified-color-icon-status-disabled`
- [x] The dark scheme dropdown will need a visible shadow. I'm good with using the box shadow values they have (`box-shadow: 0px 4px 10px 0px rgba(0,0,0,0.5)`) for now. We'll want to figure out how to add these as tokens with `light-dark()` eventually.

Other notes:
- I don't think menu dropdown has any status-related variants currently, so I'm ignoring those for this round.

marionnegp (changes requested):
- [x] I'm not seeing the box shadow for the dark dropdown. The box shadow for the light dropdown got lost in the last round of edits too.
Image

- [x] The default state for the compact button should have a transparent bg.

marionnegp (approved):

PR #3003, opened 2026-04-30
docs: Unified Theme rh-tag

markcaron:
## Summary
Unified Theme work for `rh-tag`.

Sub-item of #2827.


adamjohnson:
This PR shows `0 Files Changed`.

...?

markcaron:
@adamjohnson If I'm remembering correctly, it's because Tag was the first element I styled in the original PR.

adamjohnson:
@marionnegp I just pushed up some rh-tag unified theme changes in an attempt to match [the Figma](https://www.figma.com/design/ZSpJ3UKaAYTqMkCqSVl6Zi/Unified-Theme--RHDS---Design-Tokens--Styles---Specs?node-id=1154-12237&m=dev). You can see these changes [in the DP](https://deploy-preview-3003--red-hat-design-system.netlify.app/theming/themes/unified/#unified-tag).

Question: the figma lists a hover state; however, these tags are not interactive (they don't have links). For now, I have omitted the hover state. Is this correct? Or should non-interactive tags include a hover state? TIA.

coreyvickery:
@adamjohnson Looks good so far.

- [x] To the demo, please add `Default size with no icon` and `Compact size with icon` states for each color, both Filled and Outlined styles.
- [x] Were Tags with status coded? I see them in the Figma file, but not in the demo.

Regarding the hover state question, I think we should ask Marionne. @marionnegp

cc @markcaron

bennypowers:
tag has color and icon, but not `status`, and we're not going to add it, at least not any time soon. nonetheless i added some examples to the unified-tag demo of how to work something like this together using unified status tokens.

I've pushed what I have so far. is there `light-dark()` support in the figma?

https://deploy-preview-3003--red-hat-design-system.netlify.app/theming/themes/unified/#unified-tag

coreyvickery:
> tag has color and icon, but not `status`, and we're not going to add it, at least not any time soon. nonetheless i added some examples to the unified-tag demo of how to work something like this together using unified status tokens.

I see the new examples, thank you. All surfaces in light scheme look great. However, dark scheme is a bit buggy I think (hopefully our tokens aren't). It seems weird when flipping to dark scheme, borders are introduced to `Status - filled`.

If we all agreed not to include status, I don't think we need to hack anything together.

Screenshot 2026-06-29 at 11 39 54 PM

cc @marionnegp

markcaron:
@coreyvickery We discussed the Tag hover state last week. We don't need it since Tag isn't interactive (i.e., "Chip") in the same way as PF's Label. We do have a [linkable](https://ux.redhat.com/elements/tag/demos/#demo-links) Tag, so we should include those.

I don't think we need to hack together `status` attributes, much like we chose not to add `ghost` to CTA.

coreyvickery (approved):

PR #2959, opened 2026-04-17
feat(scheme-dropdown): add `<rh-scheme-dropdown>`

adamjohnson:
## What I did

1. Created `<rh-scheme-dropdown>`
1. Closes #2585. 

## Testing Instructions

1. Visit the [demo on the DP](https://deploy-preview-2959--red-hat-design-system.netlify.app/elements/scheme-dropdown/demo/).
2. Interact with the switch.
3. Ensure it changes the color scheme appropriately.
4. Ensure the correct item is being set in localstorage (Devtools > Application > Storage > Localstorage)
5. Clear localstorage.
6. Set your system preference to System / Light / Dark.
7. Ensure the control acts appropriately.
8. Now switch browsers and repeat steps 1-7.
9. Test the [dropdown placement](https://deploy-preview-2959--red-hat-design-system.netlify.app/elements/scheme-dropdown/demo/dropdown-placement/) demo.
10. Test for accessibility.
    * Ensure the control has an accessible name
    * Ensure each option is appropriately labeled
    * Ensure the icons have an appropriate label or are hidden where appropriate
11. Review [the docs](https://deploy-preview-2959--red-hat-design-system.netlify.app/elements/scheme-dropdown/).

### Progressive enhancement

This component uses the new [Custom Select API](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Customizable_select) ([browser support](https://caniuse.com/customizable-select)). Reviewers should test in Chrome, Firefox, Safari and others. Note that we're following the path of _progressive enhancement_. Non-Blink based browsers will not show icons in the select until browser support improves (soon!). This vastly simplifies the code, ensures cross-platform and cross-browser compatibility—and greatly aids in accessibility compliance. 

## To do's

- [x] Docs
- [x] Changeset

PR #3005, opened 2026-05-01
POC docs: `<uxdot-pattern>` support media query patterns with an iframe viewport

zeroedin:
# THIS IS A Proof of Concept

This is meant for initial discussion, likely needing design input and future direction

## What I did

1. Added `viewport` attribute to `uxdot-pattern` which encapsulates the pattern in an iframe
   - WHY?: Because in order to do patterns like stacked-navigation and in the future rh-drawer flow, we need @media query support where running inside a small contained div won't cut it.
2. Added `uxdot-context-select`, heavily borrowing from @adamjohnson work on the `rh-scheme-dropdown`.  
   -  WHY? This replaces the clunky `rh-context-picker` component making the UI more mobile friendly.  I expect if we like this we can release it as `rh-context-select` but those are exported in the package so kept this named uxdot-* for the POC.
3. Added stacked navigation pattern.  
   -  WHY? This is a pattern we've been meaning to share for awhile, used actively on RHDC blogs.  I improved it a bit by adding some progressive enhancement through container scroll-state() but falls back to the intersection observer if not supported.   I also needed a pattern to build to prove the concept.  rh-drawer not being available yet will also need this style of pattern viewport in order to properly render.

## Testing Instructions

1.  [View the Stacked Navigation Pattern](https://deploy-preview-3005--red-hat-design-system.netlify.app/patterns/stacked-navigation/examples/)

## Notes to Reviewers

This exposes `rh-navigation-secondary` inability to respond to system `light | dark` mode.  Known issue we should likely fix.


docs/patterns/index.md
adamjohnson, line 1: This will need an accompanying image in `docs/assets/patterns/` for the Stacked Navigation thumbnail on the [All Patterns page](https://deploy-preview-3005--red-hat-design-system.netlify.app/patterns/).

uxdot/uxdot-context-select.css
```diff
@@ -0,0 +1,144 @@
+:host {
+  display: inline-block;
+}
+
+select {
+  appearance: none;
```
adamjohnson: Screenshot of select in safari missing a caret icon

This is what this looks like in Safari (browser without `base-select` support yet. We definitely wanna add a caret icon for the non-custom select supporting browsers. See #2959 for and example of how to do that.

uxdot/uxdot-pattern.ts
```diff
@@ -80,26 +94,55 @@ export class UxdotPattern extends LitElement {
...
-        </form>
-
         <div id="description" class="${classMap({ empty: this.#slots.isEmpty(null) })}">
           <slot></slot>
         </div>
 
-        <rh-surface id="content">${allContent}</rh-surface>
+        <div id="controls-row">
+          ${this.viewport ? html`
+          <rh-button-group id="viewport-controls"
```
adamjohnson, line 109: I feel like these buttons could use some further design refinement. Perhaps some styling to button group to help unify them, which may include equalizing their height to the neighboring select.

Might add a title attr/tooltip too.

adamjohnson (changes requested):
This is a pretty neat idea. There's still work to do around design refinement and fallback browser support, but cool to see.

uxdot/uxdot-pattern.ts
zeroedin: 💯

uxdot/uxdot-context-select.ts
bennypowers, line 1: just fix rh-context-select. it's an internal lib element, not public, never properly designed, etc.

uxdot/uxdot-pattern.ts
```diff
@@ -49,6 +52,9 @@ export class UxdotPattern extends LitElement {
   /** Path to additional JS file to include */
   @property({ reflect: true, attribute: 'js-src' }) jsSrc?: string;
 
+  /** Render pattern in an iframe for proper @media query support */
+  @property({ type: Boolean, reflect: true }) viewport = false;
```
bennypowers: every boolean really wants to be an enum

uxdot/uxdot-pattern.ts
bennypowers, line 109: really like this idea

bennypowers (changes requested):

uxdot/uxdot-context-select.ts
```diff
@@ -0,0 +1,151 @@
...
+   * elements. After hydration completes, `firstUpdated` triggers a
+   * second render where `hasUpdated` is true and the full base-select
+   * markup (button, selectedcontent, swatches) is emitted declaratively.
+   */
+  render() {
+    const { allow, value } = this;
+    const label = this.accessibleLabel
+                || this.#internals.computedLabelText
+                || 'Color palette';
+    return this.hasUpdated ? html`
```
adamjohnson: @bennypowers What do you think about this pattern? This is essentially a workaround to SSR a custom select.

Currently, Lit SSR and `<selectedcontent>` don't play nice together. 

If you deem this acceptable, I can apply this pattern to `<rh-scheme-dropdown>`.

bennypowers: i need to understand more what the issues are with litssr and selectedcontent. we'll need to have a show and tell i think

generally though, if there's state affecting render which changes on the clientside after SSR, that render has to happen only after `hasUpdated`

adamjohnson: Here's a little backstory:

https://github.com/RedHat-UX/red-hat-design-system/pull/2959/changes/1bee358875cccbcaf5443d4f5e5b18972a0a9747

zeroedin:
@coreyvickery and @marionnegp I'd love your feedback on this proof of concept.  Not necessarily for a critique of what I've done but to get your wheels spinning for how you can take this idea and improve upon it.

uxdot/uxdot-context-select.ts
zeroedin: Just to clarify, do you mean replace `rh-context-picker` with the code in `uxdot-context-select` or create a net new `rh-context-select`?

bennypowers: - i mean don't create any new elements, just fix the ones we already have

uxdot/uxdot-context-select.ts
bennypowers: thanks for bringing this up

https://github.com/lit/lit/issues/5349

go with the multiple render paths solution for now, gated on hasUpdated

zeroedin: with a refactor in [42158ab](https://github.com/RedHat-UX/red-hat-design-system/pull/3005/commits/42158ab028e9d68e11bf9a5751e869ed91c0e3b2) this is now in rh-context-picker.ts

PR #3006, opened 2026-05-05
feat(textarea): add `<rh-textarea>`

### LINKED ISSUE #2323: [feat] Create `<rh-textarea>` component/pattern
### Description

We need a `<textarea>` component that allows us to have a multiline user input area in a form. It needs to have the proper form labelling and possibly a way for updating the aria-labelling by the implementor to fit the context of their form.

We have been encouraged to work on AI apps and in doing so we will need form elements that face the end user (in this case - RH workers I assume) that need to be properly accessible and branded.

### Acceptance Criteria

- [ ] Design done
- [ ] Development done

### Images

I have no design, I just know we need this and other form elements in the RHDS. 

### Design file

_No response_

### Other resources

https://www.w3.org/WAI/tutorials/forms/

Thanks for looking into this!

adamjohnson:
## What I did

1. Added `<rh-textarea>` element.
2. Closes #2323.

## Testing Instructions

1. View the [overview page](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/) on the DP.
2. View the following demos:
    * [Default](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/)
    * [Events](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/events/)
    * [Required](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/required/)
    * [States](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/states/)
    * [Disabled](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/disabled/)
    * [Readonly](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/readonly/)
    * [Horizontal](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/horizontal/)
    * [Field sizing](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/field-sizing/)
    * [Character count](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/character-count/)
    * [Color context](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/demo/color-context/)
3. Ensure the textarea visually matches the [Form pattern specs](https://ux.redhat.com/patterns/form/) for both default and horizontal label orientations.
4. Ensure the correct design tokens are used in the CSS.
5. Submit the required demo with an empty textarea — verify the browser validation message appears.
6. Test disabled textarea is focusable via keyboard (uses `aria-disabled`, not native `disabled`).
7. Test in Chrome, Firefox, and Safari.
8. Switch between light and dark color schemes and verify colors adapt.
9. Test the `field-sizing="content"` demo — textarea should auto-grow in Chrome/Safari but fall back to fixed height in Firefox.
10. Review [Code](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/code/) and [Accessibility](https://deploy-preview-3006--red-hat-design-system.netlify.app/elements/textarea/accessibility/) docs for accuracy.

## Notes to Reviewers

~~Demos on the DP are currently broken due to a docs server rendering issue. You will have to clone this locally to properly view/test it.~~ DP Demos now work!

Does this element need a dedicated design spec? Or are the [Form pattern specs](https://ux.redhat.com/patterns/form/) sufficient?

The element uses `aria-disabled` instead of native `disabled` so the textarea remains focusable for assistive technology users. This matches the pattern used by `rh-button`, `rh-switch`, and other RHDS form controls.

`field-sizing="content"` is a progressive enhancement — it uses the CSS `field-sizing` property which is [not supported in Firefox](https://caniuse.com/mdn-css_properties_field-sizing_content). Firefox users get a standard fixed-height textarea.

## To do's

- [x] Component
- [x] Demos
- [x] Tests
- [ ] Docs
- [ ] Changeset
- [ ] screenshot.png
- [x] Remove temporary Netlify redirect from `netlify.toml`
- [ ] SSR support (`eleventy.config.ts` — requires fixing docs server rendering first)


elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,484 @@
...
+      textarea.ariaLabelledByElements = [];
+      return;
+    }
+
+    const labels = InternalsController.getLabels(this);
+    if (labels.length > 0) {
+      textarea.ariaLabelledByElements = labels;
+      textarea.removeAttribute('aria-label');
+      return;
+    }
```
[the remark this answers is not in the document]
adamjohnson: [30e9100](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/30e910048b5829ee0f38804fa1d8c449c60b5391) [6d1a127](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/6d1a127eab10313268f0b5db62160ff7acc67b0c) [6e14c9b](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/6e14c9bc8604e9e0d9598d0af264723ff11db9ac)

marionnegp (changes requested):
- [x] Can the placeholder text use `--rh-color-text-secondary`? This will match what we have in unified theme already.

elements/rh-textarea/docs/data.yaml
```diff
@@ -0,0 +1,14 @@
+$schema: ../../../docs/schemas/element-data.schema.json
+tagName: rh-textarea
+name: Textarea
+type: element
+overallStatus: new
+libraries:
+  figma: planned
+  rhds: ready
+  shared: planned
+  docs: planned
```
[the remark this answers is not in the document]
bennypowers: #3007 should cause this warning to actually fail PRs

elements/rh-textarea/demo/horizontal.html
```diff
@@ -0,0 +1,63 @@
+---
+description: "Horizontal label layout with labels positioned to the left of the textarea using CSS Grid."
+---
+<div class="form-horizontal">
```
bennypowers: should be a form element?

elements/rh-textarea/docs/40-accessibility.md
```diff
@@ -0,0 +1,48 @@
+## Labeling
+
+Every `<rh-textarea>` must have an accessible name. Use one of the following approaches:
+
+1. External `<label>` element (recommended)
+    * associate a `<label>` with the textarea using `for` and `id`. The component wires the label to the inner textarea via `ariaLabelledByElements`. See the [default demo](/elements/textarea/demo/) for an example.
+
+2. `accessible-label` attribute: use when no visible label is present.
```
bennypowers, line 8: do we also want to add a `label` attribute for visible labels?

PRO: makes simple cases easier
CON: makes form patterns devolve to two paths - attributes for simplest layouts, label elements for complex layouts

cc @marionnegp @coreyvickery

elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  override render() {
+    const hideHelpText = this.#slots.isEmpty('help-text') && !this.helpText;
+    const showCharCount = this.showCharacterCount && this.maxlength != null;
+
+    return html`
+      <textarea id="input"
+                part="textarea"
+                .value="${live(this.value)}"
```
bennypowers, line 210: from the docs (emphasis mine):

> This is useful for cases where the DOM value may change from outside of lit-html, such as with a binding to an <input> element's value property, a content editable elements text, or to a custom element that changes it's own properties or attributes. In these cases if the DOM value changes, but the value set through lit-html bindings hasn't, lit-html won't know to update the DOM value and will leave it alone. **If this is not what you want--if you want to overwrite the DOM value with the bound value no matter what--use the live() directive**

In other words, what you're saying here is that you *always* want the rh-textarea's `value` property to overwrite the internal textarea's `value` property - yes?

elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,517 @@
...
+        <rh-icon ?hidden="${this.state !== 'warning'}"
+                 class="icon-warning"
+                 set="ui"
+                 icon="warning-fill">
+        </rh-icon>
+        <rh-icon ?hidden="${this.state !== 'danger'}"
+                 class="icon-danger"
+                 set="ui"
+                 icon="ban-fill">
+        </rh-icon>
```
bennypowers:
```suggestion
        <rh-icon ?hidden="${!helpIcon}"
                 class="icon ${classMap({ success, warning, danger })}"
                 set="ui"
                 icon="${ifDefined(helpIcon)}">
        </rh-icon>
```

don't even need the `hidden` if you 
```css
rh-icon:not([icon]) {
  display: none;
}
```

elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,517 @@
...
+  }
+
+  /* --- Form lifecycle callbacks --- */
+
+  /**
+   * Called when a parent form/fieldset is disabled or re-enabled.
+   * @param disabled whether the control should be disabled
+   */
+  protected formDisabledCallback(disabled: boolean) {
+    this.disabled = disabled;
```
bennypowers: see how we do this in pfe, noting that `InternalsController` already handles much of this. see also https://github.com/whatwg/html/issues/8365

there's an argument to be made for splitting `FACEController` out from `InternalsController`, but that's out of scope

elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,517 @@
...
+    } else {
+      this.#internals.setFormValue(this.value ?? '');
+    }
+    this.#updateValidity();
+  }
+
+  /** Sync aria-required and re-check validity when required changes. */
+  @observes('required')
+  private requiredChanged() {
+    this.#internals.ariaRequired = this.required ? 'true' : 'false';
```
bennypowers:
```suggestion
    this.#internals.ariaRequired = String(!!this.required);
```

elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,517 @@
...
+
+  /* --- Private methods --- */
+
+  /**
+   * Syncs constraint validity with current state. Checks `required`,
+   * `minlength`, and `maxlength` rules against the current value and
+   * reports the result via `ElementInternals.setValidity()`. Disabled
+   * controls are always valid per the HTML spec and skip all checks.
+   */
+  #updateValidity() {
```
bennypowers, line 426: l10n?

bennypowers (changes requested):

elements/rh-textarea/docs/40-accessibility.md
adamjohnson: I vote no. `<rh-select>` already uses a separate `<label>` element and follows the existing pattern here.

It's pretty trivial to type:

```html
<label for="my-textarea">
  First name
</label>
<rh-textarea id="my-textarea"></rh-textarea>
```

I think the cons you mentioned are pertinent. Best to avoid diverging paths for one element.

KISS - 😉

elements/rh-textarea/rh-textarea.ts
adamjohnson: Yes, we're going to have `live()` handle `value` changes `<rh-textarea>` ➡️ `<textarea>`. Then, `#onInput()` handles syncing the other way.

If this may cause adverse effects I'm not aware of, please speak up. Resolving for now.

elements/rh-textarea/demo/horizontal.html
adamjohnson: [e325d1e](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/e325d1eaa88b5e39abd8e5b4ba74a0fabe02da78)

elements/rh-textarea/rh-textarea.ts
adamjohnson: [ba26bdc](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/ba26bdcbaa28d5a93213a7e7d1cbeaef0cdda765). Great suggestion.

elements/rh-textarea/rh-textarea.ts
adamjohnson: [35d00a6](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/35d00a6c3915cacc58ac9b1db3011b56bbd36db0)

elements/rh-textarea/rh-textarea.ts
adamjohnson: [bb7afa4](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/bb7afa40e209ba5f11724bb2b757b91b66072a3a) and [bfca6a2](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/bfca6a2a9dd4b5911af1b70c5cce07145ed03833)

elements/rh-textarea/rh-textarea.ts
adamjohnson: [c285ae6](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/c285ae6952b5faecdc5374d11327c7c868e0587a) and [9adc2bd](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/9adc2bd51e548d43c44ffa5621aba52ed8059f38)

elements/rh-textarea/rh-textarea.ts
bennypowers: the way i'd usually conceive of this is change or keyup events from textarea setting the host state, and property setter on the host assigning down to the textarea. events up, properties down. that's how [swc](https://github.com/adobe/spectrum-web-components/blob/a6c29ead2237bd02005d19a217e1dfb877a69e45/1st-gen/packages/textfield/src/Textfield.ts#L327-L354) and [carbon](https://github.com/carbon-design-system/carbon-web-components/blob/421a85189575592e10dfbfb1253e26db2f0a0e2f/src/components/textarea/textarea.ts#L198-L213) do it. However [shoelace](https://github.com/shoelace-style/webawesome/blob/d6292b532c68dcc27d60213983abfc58a0dd9809/packages/webawesome/src/components/textarea/textarea.ts#L439-L463) and [mwc](https://github.com/material-components/material-web/blob/3758f46ba64e8f3e9d5dd49b76c03db941975e90/textfield/internal/text-field.ts#L623-L644) take your approach. Lion and Vaadin use slotted textarea elements.

elements/rh-textarea/rh-textarea.ts
```diff
@@ -0,0 +1,563 @@
...
+      </div>
+    `;
+  }
+
+  /**
+   * Sync the Lit property with the native textarea value.
+   * Handles both input (every keystroke) and change (blur) events.
+   */
+  #onInput() {
+    this.value = this._textarea.value;
```
zeroedin, line 243: I wonder if we should dispatch a change event here. Say if a consumer wanted to trigger an action off of the change event, currently it won't bubble outside the element.

adamjohnson: [bfce8ee](https://github.com/RedHat-UX/red-hat-design-system/pull/3006/commits/bfce8eef0d3ab19ad5e85ad4f329c6c107ea6c01) ➕

marionnegp (changes requested):
- [x] In light scheme, the read-only version should use `surface-lighter` as background and `gray-20` (no semantic token) for the border. In dark scheme, the read-only's background should be `surface-darker` and border should be `gray-70`.
- [ ] Can the success, warning, and danger icons also be positioned inside the text area? 
- [x] Can the `margin-block-start` for the character count text be 8px? We don't have that specced out in Figma, but then it'll match the spacing for helper text. (The [spacing image](https://deploy-preview-3006--red-hat-design-system.netlify.app/patterns/form/#spacing) on the Form pattern page will have to be updated.)

adamjohnson:
> Can the success, warning, and danger icons also be positioned inside the text area?

@marionnegp Yes, they can. That said, these icons may appear / disappear on individual keystrokes depending on the form validation rules people implement. This might cause text reflow / layout jank if we don't keep a permanent spot for them (aka increase padding of the textarea).

I want to avoid text reflow. The question then becomes, should we make a permanent spot for them? If so, what would that look like?

Thanks for the review!

bennypowers:
we talked about some trade offs on call today:
- lift-and-chop aria-attributes: it's not ideal from an api perspective, and may lead to unintentional xroot aria issues, but might serve as a useful backstop if downstream authors do the wrong thing. shipping this backstop however does mean adding js burden to clients, only to service misbehaving downstream devs.
- `accessible-describedby`: same xroot aria issue as before, and it would be an innovation. `accessible-label` is always a scalar, passing a reference is the issue
- requiring the use of a subset of description elements (e.g. tooltip, popover, helptext) - allows us to control the ax tree, but requires downstream to actually abide by the guidelines

i'm leaning towards the third option, possibly with the backstop as well.

PR #3007, opened 2026-05-07
docs: `since` and `new` status

bennypowers:
## What I did

- Added `since` field to elements/*/docs/data.yaml
- Added "Since vX.Y.Z" tags to element page headers, linked to release
- Added "New" tag to sidebar for elements from the current release
- Automate addition of `since` to element data in release workflow
- Fail PRs which neglect to promote planned elements upon implementation

## Testing Instructions

1. Click around the DP's /elements/ section - observe "New" tag in sidebar, "Since" tags in header

## Notes to Reviewers

@coreyvickery I made the header tags `vertical-align: middle`, which should align them with baseline ("a", not "p" or "d"). Also PTAL at the "New" tag in sidebar.

After https://github.com/webcomponents/custom-elements-manifest/issues/146, we can move this to jsdoc, and derive it from `cem` (would require changes to the version script, or manual tagging)

### DIFF
--- a/docs/styles/pages/backpage.css
+++ b/docs/styles/pages/backpage.css
@@ -112,6 +112,10 @@
 
   uxdot-header {
     grid-area: header;
+
+    & > h1 > rh-tag {
+      vertical-align: middle;
+    }
   }
 
   :where(article .container rh-code-block) {
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,7 @@ jobs:
         id: changesets
         uses: changesets/action@v1
         with:
+          version: scripts/version.sh
           publish: npx changeset publish
           commit: "chore: prepare release"
           title: "chore: prepare release"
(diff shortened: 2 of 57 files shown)


coreyvickery (dismissed):
@bennypowers Interesting, what was the inspiration for adding these?

- [ ] Please add some space (`--rh-space-md`) in between the element name in the menu and the tag, looks like they're touching right now

bennypowers:
> @bennypowers Interesting, what was the inspiration for adding these?

Slack thread

>     * [ ]  Please add some space (`--rh-space-md`) in between the element name in the menu and the tag, looks like they're touching right now

I noticed that too, but didn't change it since that was the pre-existing state of affairs with "Planned" tag. Will update shortly.

docs/styles/pages/backpage.css
```diff
@@ -112,6 +112,10 @@
 
   uxdot-header {
     grid-area: header;
+
+    & > h1 > rh-tag {
+      vertical-align: middle;
```
bennypowers, line 117:
```suggestion
      vertical-align: middle;
      margin-inline-start: var(--rh-space-md);
```

hellogreg (approved):
latest, greatest tag maker

PR #3008, opened 2026-05-07
chore: update cem

bennypowers:
## What I did

1. remove unnecessary `@alias` and `@demo` jsdoc tags
2. update cem dependency

## Testing Instructions

- Check cem serve, eleventy build


adamjohnson (approved):
Less goofiness, tidy manifest

PR #3009, opened 2026-05-07
fix(dev-server): strip YAML frontmatter from demo pages

bennypowers:
## Summary
- Add WDS plugin to strip YAML frontmatter from demo HTML before serving. pfe-tools' template middleware reads demo files raw, so frontmatter from #2951 renders as visible text. The plugin uses parse5 to find the `<div data-demo>` element and strip frontmatter from its first text node. Can be removed once we migrate to cem serve.
- Add patch-package patch for `@patternfly/pfe-tools@6.0.1` backporting two fixes from `staging/pfv6`:
  - `Manifest.getTagNames()` deduplication via `Set`
  - `PfeDemoPage.navigate()` slug regex for versioned `pf-v5-*` tag prefixes

## Test plan
- [ ] Start dev server, navigate to any element demo page (e.g. `/elements/rh-button/`)
- [ ] Verify no `---` frontmatter text visible in the rendered demo
- [ ] Verify demo content renders correctly
- [ ] Run `npm install` and verify patch applies cleanly


zeroedin (commented):
Lines Gracefully Trimmed, Merging

zeroedin (approved):

PR #3023, opened 2026-05-23
chore: bust wireit cache on css changes

bennypowers:
## What I did

1. add element css to wireit cache keys


markcaron (approved):
Looked-over Git Thing, Maybe?

PR #3001, opened 2026-04-30
docs: Unified Theme rh-switch

markcaron:
## Summary
Unified Theme work for `rh-switch`.

Sub-item of #2827.

Made with [Cursor](https://cursor.com)

PR #3040, opened 2026-06-15
feat(switch): unified API updates

markcaron:
## What I did

1. Changed `show-check-icon` boolean API to now utilize `--rh-switch-check-display` CSS custom prop rather than `hidden` attr in template to give Unified/Felt theme a "hook".

## Testing Instructions

1. Verify `show-check-icon` boolean still works as intended
2. Verify that changing `--rh-switch-check-display` to `none` hides the checkmark
3. Verify that changing `--rh-switch-check-display` to `inline-flex` shows the checkmark

## Notes to Reviewers

Unified theme requires checkmark to always be present

Related: https://github.com/RedHat-UX/red-hat-design-system/pull/3001


bennypowers:
/agentic_review can you suggest alternative approaches here? see #3001 for reference.

bennypowers:
`@property` won't work here, at least not from within the shadow root. We could make a document-level stylesheet, containing all our `@property` declarations, a requirement for using RHDS, but that would be a major change and is anyways out of scope for this work.

So the simplest alternative is to make the theming stylesheet use `inline-flex`. The downside to that approach is it requires the outside caller (i.e. the theme stylesheet) to know about icon implementation details (i.e. `inline-flex`).

Workarounds for that problem:

- **Space toggle hack** — theme sets --rh-switch-show-check: ; (space = on) or omits it (off), and CSS uses var(--rh-switch-show-check, fallback) to branch. Pros: works today. Cons: Clever but obscure, fragile, hard to document.
- **Container style queries** — `@container style(--rh-switch-show-check: true)` to conditionally apply display. Pros: future proof and correct. Cons: browser support is still partial — Firefox shipped it late, Safari too.

Safari supports container style queries (for custom properties only, which is our case) from 18, so we're good there.

Firefox support story is a drop more complicated:

FF ESR 153 releases July 21, 2026 (about a month from now) with container style query support. 

Current ESR is 140 (no style queries). ESR 140 EOL ~September 2026. After the overlap period, ESR 153 becomes the only supported ESR.

So: container style queries initially supported by ESR this quarter, completely safe starting ~October 2026 (when ESR 140 fully sunsets).

I think we can justify going with style queries for this feature, and acknowledge that users of supported ESR won't see the checkmark until the fall.

bennypowers (changes requested):
let's use container style queries

adamjohnson:
Saw this in a comment that is relevant if/when using container style queries:

> One thing to note that I discovered though, Safari is buggy with boolean states. You MUST declare the value in the @container rule or it will fail horribly.

aka `@container style(--rh-switch-show-check: true)`. I have not vetted this but thought it was worth sharing.

markcaron:
@bennypowers implemented the container style query approach. You can check it against the patch in the Unified theme docs: https://github.com/RedHat-UX/red-hat-design-system/pull/3001

.changeset/six-crabs-itch.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": minor
+---
+
+`<rh-switch>`: added `--rh-switch-check-display` CSS custom property to control checkmark icon visibility from themes without the `show-check-icon` attribute
```
bennypowers: update css prop name here

elements/rh-switch/rh-switch.css
```diff
@@ -122,12 +128,23 @@
...
 rh-icon {
+  display: none;
   margin-inline: var(--_margin-inline);
   color: var(--_switch-handle-color);
   position: absolute;
   z-index: 1;
 }
 
+:host([show-check-icon]) {
+  --rh-switch-show-check: true;
```
bennypowers, line 139: v. nice

bennypowers (changes requested):

zeroedin:
> let's use container style queries

"safe" period should start around fall this year anyway, with Firefox ESR rolling out at the end of July.

bennypowers:
> > let's use container style queries
> 
> "safe" period should start around fall this year anyway, with Firefox ESR rolling out at the end of July.

in this specific instance, I'm confident we can merge now, since there is a strong progressive enhancement story to be told for this specific feature.

.changeset/six-crabs-itch.md
bennypowers:
~~~suggestion
`<rh-switch>`: added `--rh-switch-show-check` CSS custom property to control checkmark icon visibility, without requiring the `show-check-icon` attribute on each selected element

```css
.themed-container {
  --rh-switch-check-display: true;
}
```
~~~

bennypowers (approved):

PR #3053, opened 2026-06-24
poc(mcp): add Netlify Function MCP server at /mcp

markcaron:
# POC

_Random idea!_ This PR exposes RHDS component data via the Model Context Protocol using a stateless Streamable HTTP Netlify Function.

- Resources include `cem://elements` (all elements) and `cem://element/{tagName}` (full API).
- Rate-limited to 30 req/min per IP.

## Test setup

```
{
  "mcpServers": {
    "rhds-preview": {
      "url": "https://deploy-preview-{PR#}--red-hat-design-system.netlify.app/mcp"
    }
  }
}
```

## Future setup

```
{
  "mcpServers": {
    "rhds": {
      "url": "https://ux.redhat.com/mcp"
    }
  }
}
```

## Notes to Reviewers

This is just a POC to see if we can create an RHDS endpoint for `cem`.


bennypowers:
I like the idea of offering a single "set it and forget it" url

But there are aspects of the approach that give me pause:
- cem only handles the element side, we still need asimonim for the design tokens side
- a cem endpoint can't provide pattern-level skills, Figma skills, or other red hat-level inside-baseball stuff

For these reasons, it would be best to concentrate on the skills approach in redhat-ux/red-hat-design-skills

PR #2993, opened 2026-04-30
docs: Unified Theme rh-cta

markcaron:
## Summary
Unified Theme work for `rh-cta`.

Sub-item of #2827.

Made with [Cursor](https://cursor.com)

Figma links: [Light scheme](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.3?m=dev&node-id=4107-5506) | [Dark scheme](https://www.figma.com/design/j7Rjo2W6PLloxOSXQOpll0/branch/bdwYbLWL19VAvJ1tid4ngr/RHDS-component-library---v4.1.3?m=dev&node-id=10572-8775)

Note: Project Felt-themed elements are highlighted with purple boxes in the screenshot.
Screenshot 2026-06-12 at 4 24 24 PM

PR #3055, opened 2026-06-25
feat(cta): Added hover background CSS prop and changed the internal focus ring approach

markcaron:
## What I did

1. Outer focus ring moved from `#container` to `:host`, drawn with `outline`
2. Inner focus/active ring replaced from `::after` border to `outline` with negative offset on `#container`
3. Removed `#container::after` pseudo-element entirely
4. Added `--rh-cta-hover-background-color` as a public override in the hover state rule, consistent with how focus and active already expose their background-color APIs
5. No visual changes; both rings are now `outline`-based for forced-colors compatibility

## Testing Instructions

1. Fully test demos, especially the Variants one

## Notes to Reviewers

This improves the API experience overall, but is coupled with https://github.com/RedHat-UX/red-hat-design-system/pull/2993

### DIFF
--- a/.changeset/tiny-waves-dream.md
+++ b/.changeset/tiny-waves-dream.md
@@ -2,11 +2,4 @@
 "@rhds/elements": minor
 ---
 
-`<rh-cta>`: refactor focus ring to use `outline` instead of a `::after` pseudo-element
-
-- Outer focus ring moved from `#container` to `:host`, drawn with `outline`
-- Inner focus/active ring replaced from `::after` border to `outline` with negative offset on `#container`
-- Removed `#container::after` pseudo-element entirely
-- Added `--rh-cta-hover-background-color` as a public override in the hover state rule, consistent with how focus and active already expose their background-color APIs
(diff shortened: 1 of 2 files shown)


.changeset/tiny-waves-dream.md
adamjohnson, line 1: This is a fairly verbose changeset.

Usually, we focus most on what our end users might find important. This means likely highlighting the new public rh-cta vars and probably dropping the other not user-relevant stuff.

adamjohnson (changes requested):
Update changeset, otherwise LGTM.

.changeset/tiny-waves-dream.md
zeroedin: Agree with @adamjohnson here update the changeset and this look GTG.

zeroedin (approved):
Lightweight Graduation To Modern-outlines

.changeset/tiny-waves-dream.md
bennypowers: 5b365676

PR #3060, opened 2026-06-26
feat(drawer): adds `<rh-drawer>` 

zeroedin:
## What I did

1. Created `<rh-drawer>`, a side panel component for supplementary content or navigation, aligned with the rh-drawer Figma spec.
2. Three modes driven by container queries:
    - Overlay (default) — dialog panel that slides over content with a close button, Escape, and click-outside dismiss
    - Inline — complementary landmark, always visible when a parent has container-type: inline-size, and the container is wide enough
    - Collapsible — collapsible attribute enables a push/collapse toggle at wide widths that switches to overlap at narrow widths
3. Automatically inerts surrounding content and locks scroll when in overlay or collapsible overlap mode, like a dialog that it is.
4. trigger-id property for external trigger elements
5. accessible-label for the panel.
6. Slot-based label overrides for close button (close-label), collapse toggle open (collapse-label-open), and collapse toggle closed (collapse-label-closed) with attribute counterparts.
7. 12 demos covering responsive, collapsible, overlay, contained, fixed position, RTL, session storage, and pattern usage with  `rh-navigation-primary` (for discussion) and `rh-navigation-vertical`.
8. Added side navigation pattern docs linking to pattern demos.
9. Fixed upstream PFE dev server text-align: left to start for RTL support.

**Supersedes:** https://github.com/RedHat-UX/red-hat-design-system/pull/2850

The component was simplified significantly from the previous attempt:
  - rh-drawer.ts: 802 → 400 lines 
  - rh-drawer.css: 858 → 399 lines

##  Testing Instructions

  1. Visit the demos on the deploy preview.
  2. Test overlay, responsive, collapsible, and RTL demos.
  3. Run npx wtr elements/rh-drawer/test/rh-drawer.spec.ts.

## Notes to Reviewers

The drawer responds to its nearest ancestor with container-type: inline-size. Without one, it always behaves as an overlay. This was one of the refactorings from the previous drawer. I'm still not convinced this implementation gets it right, as it almost always forces full pattern usage, external container + drawer. In my opinion, a drawer is both the panel (the face and knob) and the box for the contents (holding the clothes).  Doing so would allow for easier implimenation I think.


elements/rh-drawer/README.md
bennypowers, line 1: `variant` attr reference in out of date.

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+  @property({ attribute: 'close-label' }) closeLabel?: string;
+
+  /** Label for the collapse toggle when open. Overridden by the `collapse-label-open` slot. */
+  @property({ attribute: 'collapse-label-open' }) collapseLabelOpen?: string;
+
+  /** Label for the collapse toggle when closed. Overridden by the `collapse-label-closed` slot. */
+  @property({ attribute: 'collapse-label-closed' }) collapseLabelClosed?: string;
+
+  @state() private _suppressTransition = false;
+  @state() private _narrowContainer = false;
```
bennypowers:
```suggestion
  @state() protected _narrowContainer = false;
```
deal with those ts errors

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+    if (!this.dispatchEvent(event)) {
+      this.#reverting = true;
+      this._suppressTransition = true;
+      this.open = oldValue;
+      await this.updateComplete;
+      this.#reverting = false;
+      this.#suppressTransitionBriefly();
+      return;
+    }
+    this.#triggerElement?.setAttribute('aria-expanded', String(!!newValue));
```
bennypowers, line 325: do this on init as well

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+  @observes('_narrowContainer')
+  protected _narrowContainerChanged(old?: boolean, value?: boolean) {
+    if (old == null || value == null || old === value) {
+      return;
+    }
+    this.#restoreInertedContent();
+    document.body.style.overflow = '';
+    document.removeEventListener('pointerdown', this.#onDocumentPointerDown);
+    this.#suppressTransitionBriefly();
+    this.open = !value;
```
bennypowers, line 365: make sure to reset _narrowContainer as well if the event is cancelled: make sure all state is synchronized

elements/rh-drawer/docs/40-accessibility.md
bennypowers, line 1: needs prose

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+ * enable a collapse toggle (push at wide, overlap at narrow).
+ *
+ * @summary Slides a panel in from the side for supplementary content or navigation
+ *
+ * @alias drawer
+ *
+ * @slot - Expects block elements for panel content such as `div`, `nav`, headings, or `rh-navigation-vertical`.
+ * @slot close-label - Expects inline text for the close button. Defaults to "Close drawer". Should be localized for screen readers.
+ * @slot collapse-label-open - Expects inline text for the collapse toggle when open. Defaults to "Collapse panel". Should be localized for screen readers.
+ * @slot collapse-label-closed - Expects inline text for the collapse toggle when closed. Defaults to "Expand panel". Should be localized for screen readers.
```
bennypowers: labels shouldn't change with state. remember switch? cc @adamjohnson

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+   * closed. The referenced element receives click handling automatically
+   * and receives focus when the panel closes.
+   */
+  @property({ attribute: 'trigger-id' }) triggerId?: string;
+
+  /**
+   * Accessible label for the drawer panel. Used as `aria-label` on the
+   * panel element. Defaults to 'Panel'.
+   */
+  @property({ attribute: 'accessible-label' }) accessibleLabel = 'Panel';
```
bennypowers: what about labelledby a slotted header? there are some xroot aria issues here if the host doesn't have the role

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+  #reverting = false;
+  #hasContainerContext = false;
+  #resizeObserver?: ResizeObserver;
+  #inertedElements = new Set<Element>();
+
+  /** Which side the panel appears on. */
+  @property({ reflect: true }) inline: 'start' | 'end' = 'start';
+
+  /** CSS positioning mode for the overlay panel. */
+  @property({ reflect: false }) position: 'absolute' | 'fixed' = 'absolute';
```
bennypowers: does this have to be an attr? can it be automatic via container query?
or maybe can we use a custom css property and a style query?

elements/rh-drawer/rh-drawer.ts
```diff
@@ -0,0 +1,400 @@
...
+      document.addEventListener('keydown', this.#onDocumentKeyDown);
+      let ancestor: Element | null = this.parentElement;
+      while (ancestor) {
+        const ct = getComputedStyle(ancestor).containerType;
+        if (ct === 'inline-size' || ct === 'size') {
+          this.#hasContainerContext = true;
+          this.#resizeObserver = new ResizeObserver(entries => {
+            const [entry] = entries;
+            if (entry) {
+              const narrow = entry.contentBoxSize[0].inlineSize < 992;
```
bennypowers, line 193: does this need to be configurable per spec?

bennypowers (changes requested):

elements/rh-drawer/rh-drawer.ts
zeroedin: That is definitely a question for @coreyvickery  prior iteration allowed it to be via tshirt sizes, curernt figma suggests only 992.

elements/rh-drawer/rh-drawer.ts
zeroedin: IIRC, a fixed doesn't technically need a parent container, so I believe that's the reason I moved it to an attribute. 

Brings us back to the outstanding question of: does the drawer "always" need a parent container? If so, is that better served as a pattern-level requirement, or should we revisit adding that back to the element itself?

elements/rh-drawer/rh-drawer.ts
zeroedin: oh good call

elements/rh-drawer/rh-drawer.ts
zeroedin: The mode change from _narrowContainerChanged is driven by container resize, not something a consumer can meaningfully cancel. However, the drawer is still opening or closing, so consumers should be notified. Updated the DrawerOpenEvent and DrawerCloseEvent constructors to accept an optional { cancelable } option (defaults to true). When _narrowContainerChanged triggers, it sets a #modeChange flag, so _openChanged dispatches a non-cancelable event; consumers still receive the event, but preventDefault() has no effect. User-initiated open/close remains cancelable.

coreyvickery (changes requested):
@zeroedin Let's remove any box shadows on the actual tray, but not the button. Ready to ship after that.

elements/rh-drawer/rh-drawer.ts
zeroedin: After further review, this has more to do with inert scoping for focus trap than anything else. I"ll need to do more research here on how best to handle.

> absolute is scoped to the containing block — so only the parent's siblings need to be inert because content outside that container is already outside the overlay's visual scope. fixed covers the viewport, so everything else in the document needs to be inert.

we are emulating a dialog here as the drawer functions as a modal
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/

PR #3066, opened 2026-06-29
fix(jump-links): replaced after psuedo in favor of outline

zeroedin:
## What I did

1. Removed `::after` psuedo in favor of `outline` reduces need for the focus-offset css var in unified support
2. Updated changeset to reflect removal

## Testing Instructions

1.

## Notes to Reviewers

May close this PR; outline is broken in horizontal variant; this approach may not work as hoped.

### DIFF
--- a/.changeset/fiery-coins-march.md
+++ b/.changeset/fiery-coins-march.md
@@ -2,4 +2,4 @@
 "@rhds/elements": minor
 ---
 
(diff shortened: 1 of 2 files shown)


zeroedin:
Closing this. I'm pretty sure we need 2 levels of containers to make the unified theme work here.    The `:before` needs moved to an outer container to seperate it from the link itself to really mimic the focus outline in unified.

PR #3000, opened 2026-04-30
docs: Unified Theme rh-subnav

markcaron:
## Summary
Unified Theme work for `rh-subnav`.

Sub-item of #2827.

See https://github.com/RedHat-UX/red-hat-design-system/pull/3044 for testing instructions. This requires cloning this branch/PR to your computer and taking the CSS from 3044.

PR #3068, opened 2026-06-30
feat(navigation-link): add indicator CSS custom properties

bennypowers:
## Summary
- Add public CSS custom properties to `rh-navigation-link`'s `::before` indicator pseudo-element for theming the indicator as a pill-shaped background bar
- New properties: `--rh-navigation-link-indicator-border-radius`, `--rh-navigation-link-indicator-background-color`, `--rh-navigation-link-indicator-inset-block`, `--rh-navigation-link-indicator-inset-inline-end`, `--rh-navigation-link-indicator-inline-size`
- Each public var falls back to a private channel var (for parent component use), then to existing defaults
- Also adds private var fallback to link element `border-radius` so `--_navigation-link-border-radius` set by `rh-subnav` is consumed

## Context
Needed by #3000 (unified subnav) to implement pill-shaped indicator on `rh-navigation-vertical` per unified theme Figma spec (Corey's review feedback).

## Test plan
- Existing unit tests pass (24/24, 0 failures)
- Stylelint passes
- No visual regression: all new properties default to existing behavior

### DIFF
--- a/.changeset/brave-dots-glow.md
+++ b/.changeset/brave-dots-glow.md
@@ -1,0 +1,5 @@
+---
+"@rhds/elements": minor
+---
+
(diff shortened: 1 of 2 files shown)


zeroedin:
Closing this in favor of the changes that were made with rh-navigation-vertical in https://github.com/RedHat-UX/red-hat-design-system/pull/3073

PR #3071, opened 2026-06-30
feat(cta): add  CSS prop

markcaron:
## What I did

1. Added `--rh-cta-focus-text-decoration` CSS prop

Related to https://github.com/RedHat-UX/red-hat-design-system/pull/2993


.changeset/lemon-sheep-dress.md
zeroedin, line 1: Edit the previous changeset instead of creating a new one.

zeroedin (approved):

PR #3002, opened 2026-04-30
docs: Unified Theme rh-tabs

markcaron:
## Attn: Contains changes to `<rh-tabs>`. Do NOT merge yet

See "Component changes" below for more info.

## Summary
Unified Theme work for `rh-tabs`.

## How to test:

1. Visit the [Felt Theme](https://deploy-preview-3002--red-hat-design-system.netlify.app/theming/themes/unified/#unified-tabs) page on the DP.
2. Scroll down to the "Tabs" section.
3. Right click > Inspect element.
4. Locate the `<uxdot-pattern>` element that contains tabs.
5. Remove the `max-width: 56rem;` on the `:host` selector. 
    * This enables you to see the vertical tabs on the DP.
    * Otherwise, the media query makes vertical tabs > horizontral tabs due to space constraints.
6. Verify the margin and paddings are correct.
7. Verify the borders and "pills" look correct.
8. Verify in light and dark color schemes.
9. Verify I used the appropriate unified theme tokens for various colors and pixel values
10. Verify the left and right buttons look correct for narrow viewports across default and box inset variants.
11. Compare to the [Figma](https://www.figma.com/design/qRn3S225WwGHIdgGZX0LTC/Unified-Theme--Design-Tokens--Styles---Specs?m=dev&node-id=12810-2355).

### Limitations

The following require API/component changes. They are not currently achievable via CSS-only theming:

  * Rounded inner pill on hover/focus for tab and overflow arrows

### Component changes

In order to make rh-tabs look like the "Unified theme" tabs, we have to make API changes to rh-tabs. I have pushed the changes to rh-tabs up to this PR. This makes testing and review easier; however, we will have to revert these changes and move them to their own PR (`staging/unified-tabs`). Then, using the Unified theme's tabs will require an RHDS version of >4.1.3.

Before moving rh-tabs changes into their own PR, I would like to get the unified theme tabs reviewed for accuracy to make sure we get everything 100% right and don't have to make multiple releases of RHDS.

__________________

Sub-item of #2827.

Made with [Cursor](https://cursor.com)

PR #3075, opened 2026-06-30
feat(tabs): add CSS custom properties and overflow-button parts for theme support

markcaron:
## What I did

`<rh-tab>`: add `wrapper` element plus CSS custom props — the inner pill-shaped element wrapping the icon and label, enabling themed hover/focus backgrounds.

`<rh-tabs>`: add `overflow-button`, `overflow-button previous`, and `overflow-button next` CSS parts on the overflow scroll buttons, enabling themed padding, sizing, and focus/hover styling.

## Testing Instructions

1. Double-check the changes match the changes in [Adam's original Unified Tabs PR](https://github.com/RedHat-UX/red-hat-design-system/pull/3002/changes#diff-36cb788fdca5f81e140abbf7eaf87c1dc03fc5ccc1aa10885c03acf9cac65c7e), which have been abstracted here. 

## Notes to Reviewers

Related to https://github.com/RedHat-UX/red-hat-design-system/pull/3002

### DIFF
--- a/elements/rh-tabs/rh-tab.css
+++ b/elements/rh-tabs/rh-tab.css
@@ -46,9 +46,6 @@ rh-icon,
   position: relative;
   display: flex;
   flex: 1;
-
-  /** Gap between icon and label using rh-space-md */
-  gap: var(--rh-space-md, 8px);
   text-decoration: none;
   cursor: pointer;
   align-items: center;
@@ -156,7 +153,7 @@ rh-icon,
       background-color: transparent;
 
       &:before {
-        border-inline-color: var(--rh-color-border-subtle);
+        border-inline-color: var(--_tab-box-side-border-color, var(--rh-color-border-subtle));
         border-inline-start-width: var(--rh-border-width-sm, 1px);
       }
 
@@ -167,14 +164,18 @@ rh-icon,
 
       &.first {
         &:before {
-          border-inline-start-color: var(--rh-color-border-subtle);
+          border-inline-start-color:
+            var(--_tab-box-side-border-color,
+              var(--rh-color-border-subtle));
           border-inline-start-width: var(--rh-border-width-sm, 1px);
         }
       }
 
       &.last {
         &:before {
-          border-inline-end-color: var(--rh-color-border-subtle);
+          border-inline-end-color:
+            var(--_tab-box-side-border-color,
+              var(--rh-color-border-subtle));
           border-inline-end-width: var(--rh-border-width-sm, 1px);
         }
       }
@@ -224,7 +225,9 @@ rh-icon,
         &.first {
           @container (min-width: 768px) {
             &:before {
-              border-inline-start-color: var(--rh-color-border-subtle);
+              border-inline-start-color:
+                var(--_tab-box-side-border-color,
+                  var(--rh-color-border-subtle));
               border-inline-start-width: var(--rh-border-width-sm, 1px);
             }
           }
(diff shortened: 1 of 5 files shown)


elements/rh-tabs/rh-tab.css
```diff
@@ -156,7 +153,7 @@ rh-icon,
       background-color: transparent;
 
       &:before {
-        border-inline-color: var(--rh-color-border-subtle);
+        border-inline-color: var(--_tab-box-side-border-color, var(--rh-color-border-subtle));
```
bennypowers: `--_` "private" css props *must not* be used for theming purposes

elements/rh-tabs/rh-tab.ts
```diff
@@ -138,19 +138,14 @@ export class RhTab extends LitElement {
...
-        <slot name="icon"
-              part="icon">
-          <rh-icon ?hidden="${!this.icon}" icon="${ifDefined(this.icon)}" set="${ifDefined(this.iconSet)}"></rh-icon>
-        </slot>
-        <!-- summary: Tab label
-             description: |
-               Tab label text. Authors should keep labels short
-               and descriptive. -->
-        <slot part="text"></slot>
+        <div id="wrapper" part="wrapper">
```
bennypowers: It makes sense to expose the overflow buttons as parts, since those are actually "part" of the tabs interface. A layperson might say "the tabs bar has three parts: previous button, next button, and tab list". It's immediately obvious from the name what exactly the next and previous button parts control.

This new wrapper part and element though, is different. It's not a priori obvious why a "tab" has or needs a "wrapper", or why one would want to style it. This, especially, since we *already* expose a button part. 

I don't think we should add this wrapper part. Can we achieve the same end result by adding new component props, styling the host, or adding a part to an existing element?

bennypowers (changes requested):

elements/rh-tabs/rh-tab.css
markcaron: Unbelievable, Claude. My fault for not catching it.

elements/rh-tabs/rh-tab.ts
bennypowers: Drop part="wrapper", keep #wrapper as internal structure, replace --_ private props with public --rh-tabs-* props.

Proposed mapping:

|          Private (remove)           |               Public (add)                |
|-------------------------------------|-------------------------------------------|
| --_tab-wrapper-radius               | --rh-tabs-link-inner-radius               |
| --_tab-wrapper-padding              | --rh-tabs-link-inner-padding              |
| --_tab-wrapper-justify              | --rh-tabs-link-inner-justify              |
| --_tab-hover-background             | --rh-tabs-link-hover-background           |
| --_tab-focus-background             | --rh-tabs-link-focus-background           |
| --_tab-focus-button-outline         | --rh-tabs-link-focus-outline              |
| --_tab-focus-button-outline-offset  | --rh-tabs-link-focus-outline-offset       |
| --_tab-focus-wrapper-outline        | --rh-tabs-link-focus-inner-outline        |
| --_tab-focus-wrapper-outline-offset | --rh-tabs-link-focus-inner-outline-offset |
| --_tab-box-side-border-color        | --rh-tabs-box-border-color                |

elements/rh-tabs/rh-tab.css
markcaron: Fixed.

elements/rh-tabs/rh-tab.ts
markcaron: CSS part `wrapper` dropped. Only using public CSS props within `#wrapper` now.

bennypowers (approved):
luxury grouped tab management

PR #3093, opened 2026-07-13
docs(jump-links): remove toggle event jsdoc

zeroedin:
## What I did

1. Removed JSdoc for an event jump-links has never fired.

## Testing Instructions

1.


adamjohnson (approved):
Legacy Garbage Tossed Mercilessly

PR #3105, opened 2026-07-16
fix(navigation-primary): accessibility improvement ensure hidden 

zeroedin:
## What I did

1. Added `!important` to the `.hidden` class so it doesn't accidentally get overridden in container query breakpoints.

## Testing Instructions

1. View DP navigation-primary/demo/no-secondary-items demo, ensure #events remains `display: none` so it isn't reported as a role="list" in the AT


hellogreg (approved):
Lumar's good. Manual, too.

PR #3158, opened 2026-08-05
fix: patch @rhds/tokens, no-unknown, light-dark() support

### LINKED ISSUE #3121: docs: create patch file (temporary) for @rhds/tokens
Couple changes needed upstream in `@rhds/tokens` to support requiring the fallbacks in #3120 

1.  no-unknown-token-name.js:
    - Makes the rule function async
    - Adds null-safe access for root.source?.input?.file so it doesn't crash when the file path is missing
    - Dynamically imports package.json and custom-elements.json to auto-populate the allowed set with --rh-prefixed CSS custom properties from the custom elements manifest (only when running inside
  @rhds/elements)

2.  token-values.js:
    - Adds extractLightDarkParts and lightDarkMatches helper functions that parse a light-dark() value into
  its two sub-tokens (name + hex fallback each), then compare those parts structurally between the
  expected token value and what the author wrote
    - For light-dark() tokens, the rule compares sub-token names and hex values individually instead of
  doing a raw string comparison against the full expression. This means light-dark(var(--on-light,
  #151515), var(--on-dark, #fff)) written across multiple lines matches the single-line canonical value —
  formatting differences don't cause false positives
    - Wrong sub-token names or hex values are caught and reported
    - The autofix inserts the canonical light-dark() expression as the fallback

zeroedin:
## What I did

Updated the @rhds/tokens patch to fix two stylelint plugin issues and configured the no-unknown-token-name rule to use the custom elements manifest.

1. token-values rule — added light-dark() support: The upstream rule doesn't know how to validate light-dark() fallback values. Added extractLightDarkParts() and lightDarkMatches() functions to structurally compare light-dark(var(...), var(...)) fallbacks against the canonical token values, with an auto-fixer that replaces incorrect fallbacks.
2. no-unknown-token-name rule — added opt-in cem option: The rule flags any --rh-* custom property not in the system token map, but component CSS custom properties (like --rh-tabs-link-color) are intentional API. Added a cem config option that reads custom-elements.json from the project root to discover and allow these properties. Also fixed null safety for root.source.input.file.
3. Configured cem location in .stylelintrc.yml:
```yaml
rhds/no-unknown-token-name:
  - true
  - cem: ./custom-elements.json
```

Closes #3121

## Testing Instructions

1.

## Notes to Reviewers

We expect additional lint failures. This PR stack contains many changes across the stack, and the lint failures should decrease with each level of fixes.

### DIFF
--- a/.stylelintrc.yml
+++ b/.stylelintrc.yml
@@ -83,9 +83,14 @@ rules:
 
   rhds/deprecated: true
 
-  rhds/no-unknown-token-name: true
+  rhds/no-unknown-token-name:
+    - true
+    - cem: ./custom-elements.json
 
   rhds/require-token-fallback: true
+
+  rhds/token-values: true
+
 overrides:
   - files:
       - docs/**/*.scss
--- a/patches/@rhds+tokens+3.1.0.patch
+++ b/patches/@rhds+tokens+3.1.0.patch
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@rhds/tokens/plugins/stylelint/rules/no-unknown-token-name.js b/node_modules/@rhds/tokens/plugins/stylelint/rules/no-unknown-token-name.js
-index 5c51f5f..86fc42b 100644
+index 5c51f5f..0a61471 100644
 --- a/node_modules/@rhds/tokens/plugins/stylelint/rules/no-unknown-token-name.js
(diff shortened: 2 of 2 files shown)


adamjohnson (approved):
Linting Gets Tamed, Mostly

You should open an issue on rhds/tokens to track what is needed for this patch (if there's not one already) and link it in this issue/PR.

PR #3161, opened 2026-08-05
docs: remove inline style tags were appropriate, fix missing fallbacks

### LINKED ISSUE #3152: [docs]: inline styles should include fallbacks
Tracked by #3119

`<style>` tags that include usage of red hat design system token primitives or theme tokens should provide fallbacks.  global.css or default-theme.css are not guaranteed to be loaded on a page.  default-theme.css is only ever injected when an element using `@themeable` is present, but that is not a guarantee.

zeroedin:
## What I did

Extracted inline `<style>` and `<style data-helmet>` blocks from docs HTML and Markdown files into colocated CSS files, making them lintable by stylelint.

1. Extracted `<style>` blocks across docs into colocated .css files (e.g., `docs/about/index.md` → `docs/about/index.css`), replacing them with `<link rel="stylesheet" href="..." data-helmet>` tags
2. Kept inline styles for uxdot-pattern sources used by `<uxdot-pattern src="...">` fallbacks corrections were applied, since the component requires inline `<style>` blocks the files were abstracted to standalone css then set back to inline `<style>` until we can decide on a path forward for inline style stylelinting.
3. Applied token fallbacks and line-length fixes to all extracted CSS via `stylelint --fix`, then manually reformatted `light-dark()` values to stay under the 100-character line limit
4. Fixed pre-existing bugs surfaced by linting the previously invisible inline CSS:
`--rh-color-interactive-primary-visited` → `--rh-color-interactive-primary-visited-default` (wrong token name in 2 files) `--rh-space--2xl` → `--rh-space-2xl` (double-dash typo) and `--rh-font-weight-heading-normal` → `--rh-font-weight-heading-regular` (non-existent token) `.page-overvie` typo (fixed in docs/tokens/index.css)
5. Removed duplicate CSS blocks from `project-felt/index.css` (3 `<style>` blocks had been merged with duplicated selectors)
6. Improved CSS quality by nesting flat selectors where appropriate and adding a stylelint-disable comment for no-descending-specificity in files where CSS nesting creates false positives
7. Added class="theming-example" to 2 inline SVGs in project-felt/index.md to scope their fill/stroke class selectors under `svg.theming-example` removing those styles from the `<svg>` tags themselves.

Closes #3152

## Testing Instructions

1.

## Notes to Reviewers

Lint errors are still expected; they should be related to `uxdot-*` elements, which will be a PR stack on top of this branch.  The remaining issues should be within `elements/**/*.css` and will be handled by separate PRs outside the stack cut from the base feature branch.


markcaron (approved):
Linting Guts, Trims Methodology

docs/theming/themes/project-felt/index.css
```diff
@@ -0,0 +1,88 @@
+#theming-screenshots {
+  margin-block: var(--rh-space-2xl, 32px) var(--rh-space-4xl, 64px);
+
+  &::part(container) {
+    padding: 0;
+  }
+}
+
+svg.theming-example {
```
adamjohnson, line 9: Looking at these SVG files [in the DP](https://deploy-preview-3161--red-hat-design-system.netlify.app/theming/themes/project-felt/#project-felt-theme), they are blank.

They should not be blank (see [Project Felt theme](https://ux.redhat.com/theming/themes/project-felt/#project-felt-theme) on uxdot).

By extracting these styles to an external stylesheet, the shadowroot blocks applying these styles to these SVGs. Thus, they look broken.

Tl;dr these styles need to stay inline so that they can be read by `<uxdot-example>`.

adamjohnson (changes requested):
Fix the styles for the Project Felt SVGs then we're set to jet for this one.

docs/theming/themes/project-felt/index.css
zeroedin: Screenshot 2026-08-20 at 11 35 51 AM

Interestingly enough they are not in the shadowroot, these are slotted, but are empty in contents so empty for a separate reason.

adamjohnson (approved):
Loose Globals Turn Modular

PR #3162, opened 2026-08-06
docs: update focus guidance with border-radius soft default

### LINKED ISSUE #3153: [docs] add soft default for focus indicator border-radius
The example CSS for our current [focus indicator guidance](https://ux.redhat.com/foundations/interactions/focus-indicators/) doesn't include a default `border-radius`. As a result, any focus indicators on elements without a specified `border-radius` will appear sqaured-off, rather than having our rounded style.

We should update the example CSS with a selector and property that:

* Applies a 3px radius (same as the line thickness and offset) to indicators on elements that don't have a radius specified.
* Does _not_ apply a radius to elements with a specified border radius (e.g., pill-shaped buttons).

The simplest way to do this is likely by adding a `:where` pseudo-class selector (which has zero specificity) to the example CSS:

```
:is(*, :hover):focus-visible {
  outline-color: light-dark(--rh-color-blue-50, --rh-color-blue-30);
  outline-offset: 3px;
  outline-style: solid;
  outline-width: 3px;
  transition: none; 
}

:where(:focus-visible) {
  border-radius: 3px;
}
```

hellogreg:
## What I did

1. Added text on supplying a default border radius.
2. Updated the CSS samples with `:where` selectors containing default `border-radius` property value of 3px.

## Testing Instructions

Go to the [deploy preview page](#) and...

1. Check that the following verbiage is accurate and has been added to the list in the "How we style focus" section:

> has a default border radius of 3px (which is overridden if the element already has a specified radius).

2. Verify that the "Basic CSS" and "Full example CSS" code blocks include the following CSS (updated to include token):

```
:where(:focus-visible) {
  border-radius: var(--rh-border-width-lg, 3px);
}
```

(Note: we use `:where` in order to remove specificity. That way, if an element already has a border-radius, the 3px radius should not be applied.)

### DIFF
--- a/docs/foundations/interactions/focus-indicators/index.md
+++ b/docs/foundations/interactions/focus-indicators/index.md
@@ -55,7 +55,7 @@ In CSS terms, here is some minimal code that meets these requirements:
 
 ```css rh-code-block
 :is(*, :hover):focus-visible {
-  outline-color: light-dark(--rh-color-blue-50, --rh-color-blue-30);
+  outline-color: light-dark(var(--rh-color-blue-50), var(--rh-color-blue-30));
   outline-offset: 3px;
   outline-style: solid;
   outline-width: 3px;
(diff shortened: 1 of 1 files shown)


coreyvickery (approved):
@hellogreg Large Giants Take Meth

PR #3183, opened 2026-08-10
fix(scheme-dropdown): correct option labels, delegatesFocus, and SSR

adamjohnson:
## What I did

1. Fix customizable `<select>` option labels rendering as `"false"` / mangled text under SSR + hydration.
    * Lit child bindings inside `<option>` leave `<!--lit-part-->` markers that `<selectedcontent>` clones. This is a problem. See [lit#5349](https://github.com/lit/lit/issues/5349).
    * Option labels now use escaped `unsafeStatic` text instead.
1. Move `?selected` derivation into `render()` so SSR marks the correct option (no more client-only `#isLight` / `#isDark` / `#isSystem` flags).
1. Add `delegatesFocus` so host `.focus()` lands on the internal `<select>`.
1. Enable `<rh-scheme-dropdown>` in the Eleventy Lit SSR `componentModules` list.
1. After hydration, sync `select.value` and each option’s `selected` content attribute to scheme in `updated()` (Lit’s `?selected` can stay stale after SSR / user interaction).
1. Fall back nullish `accessible-label-*` values to English defaults (`System` / `Light` / `Dark`) so `render()` does not throw.
1. Map unknown or malformed `scheme` values to System (`light dark`) via `#resolvedScheme`.
1. Add tests for selected-attr sync, nullish accessible labels, and malformed scheme fallback.
1. Add a localization demo.

## Testing Instructions

1. First, view the [scheme dropdown demo on uxdot](https://ux.redhat.com/elements/scheme-dropdown/demo/) in Chrome/Edge.
1. Note the broken option labels (`false` / mangled text).
1. Now view the [scheme dropdown demo on this DP](https://deploy-preview-3183--red-hat-design-system.netlify.app/elements/scheme-dropdown/demo/).
    * Labels should read **System**, **Light**, and **Dark**
    * System should be selected by default (unless `localStorage.rhdsColorScheme` is set)
1. View page source on the DP demo and confirm:
    * Declarative shadow DOM is present on `<rh-scheme-dropdown>`
    * `<option value="light dark" selected>` is in the SSR HTML
    * No `<!--lit-part-->` / `<!--?lit-->` comments inside `<option>`
1. Switch Light → Dark → System and confirm selection + `document.body` color-scheme update.
1. Tab to the control and confirm the focus ring appears (host focus is delegated to the `<select>`).
1. Check the [localization demo on this DP](https://deploy-preview-3183--red-hat-design-system.netlify.app/elements/scheme-dropdown/demo/localization/) — labels should be **Sistema**, **Claro**, and **Oscuro**.
1. Hard-refresh demos and confirm there are no Lit hydration errors in the console related to option parts.
1. After switching options, confirm in DevTools that `selected` moves with the active `<option>` (not stuck on System).
1. In DevTools, set a nonsense `scheme` value (or clear an `accessible-label-*` prop) and confirm System / English defaults still show.

## Notes to Reviewers

Workaround for [lit#5349](https://github.com/lit/lit/issues/5349): child bindings inside `<option>` break under custom select / hydration. `?selected` is fine for SSR first paint, but can go stale after hydration — `updated()` realigns `select.value` and the `selected` attributes.

`rh-scheme-dropdown` was previously commented out of `componentModules` as “still not working nicely with ssr” — this PR re-enables it now that the custom-select label issue is addressed.

### DIFF
--- a/elements/rh-scheme-dropdown/demo/localization.html
+++ b/elements/rh-scheme-dropdown/demo/localization.html
@@ -1,0 +1,26 @@
+---
+description: "Scheme dropdown with Spanish accessible labels for light, dark, and system options."
+---
+<rh-scheme-dropdown accessible-label="Tema"
+                    accessible-label-light="Claro"
+                    accessible-label-dark="Oscuro"
+                    accessible-label-system="Sistema"></rh-scheme-dropdown>
+
+<script type="module">
+  import '@rhds/elements/rh-scheme-dropdown/rh-scheme-dropdown.js';
+</script>
+
+<style>
+  body {
+    color-scheme: light dark;
+    background-color: var(--rh-color-surface);
+    color: var(--rh-color-text-primary);
+  }
+
+  rh-scheme-dropdown {
+    /* Wider than the English defaults so Spanish option text fits. */
+    --rh-scheme-dropdown-select-inline-size: 120px;
+    --rh-scheme-dropdown-picker-inline-size: 160px;
+    padding: var(--rh-space-md, 8px);
+  }
+</style>
--- a/elements/rh-scheme-dropdown/rh-scheme-dropdown.ts
+++ b/elements/rh-scheme-dropdown/rh-scheme-dropdown.ts
@@ -1,4 +1,5 @@
-import { html, isServer, LitElement } from 'lit';
+import { isServer, LitElement } from 'lit';
+import { html, unsafeStatic } from 'lit/static-html.js';
 import { customElement } from 'lit/decorators/custom-element.js';
 import { property } from 'lit/decorators/property.js';
 
@@ -98,6 +99,16 @@ export class RhSchemeDropdown extends LitElement {
   }
 
   render() {
+    // IMPORTANT: do not put Lit child bindings (${...}) inside <option>.
+    // In Custom Select browsers, <selectedcontent> cloneNode()'s the selected
+    // option — including Lit's <!--?lit--> markers — which desyncs template
+    // parts and renders boolean values as labels ("false"). See lit#5349.
+    // unsafeStatic inlines escaped label text with no ChildPart markers.
+    // Author-facing strings must be HTML-escaped before unsafeStatic.
+    const labelSystem = unsafeStatic(this.#escapeHtml(this.accessibleLabelSystem));
+    const labelLight = unsafeStatic(this.#escapeHtml(this.accessibleLabelLight));
+    const labelDark = unsafeStatic(this.#escapeHtml(this.accessibleLabelDark));
+
     return html`
       <label for="scheme-dropdown" class="visually-hidden">${this.accessibleLabel}:</label>
       <select id="scheme-dropdown" @change="${this.#onChange}">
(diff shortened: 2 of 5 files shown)


elements/rh-scheme-dropdown/rh-scheme-dropdown.ts
```diff
@@ -83,47 +80,87 @@ export class RhSchemeDropdown extends LitElement {
...
   render() {
+    // IMPORTANT: do not put Lit child bindings (${...}) inside <option>.
+    // In Custom Select browsers, <selectedcontent> cloneNode()'s the selected
+    // option — including Lit's <!--?lit--> markers — which desyncs template
+    // parts and renders boolean values as labels ("false"). See lit#5349.
+    // unsafeStatic inlines escaped label text with no ChildPart markers.
+    // Author-facing strings must be HTML-escaped before unsafeStatic.
+    const labelSystem = unsafeStatic(this.#escapeHtml(this.accessibleLabelSystem));
+    const labelLight = unsafeStatic(this.#escapeHtml(this.accessibleLabelLight));
+    const labelDark = unsafeStatic(this.#escapeHtml(this.accessibleLabelDark));
```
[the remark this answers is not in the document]
adamjohnson: Using the `.textContent` binding flattens text inside the `<option>`. This is a [custom select](https://developer.mozilla.org/en-US/docs/Learn_web_development/Extensions/Forms/Customizable_select) and has extra HTML within each `<option>` tag—so using the `.textContent` binding is not feasible.

I updated the comment above these lines in [96ff980](https://github.com/RedHat-UX/red-hat-design-system/pull/3183/commits/96ff980a5be41daad6262d998cf7df0602dd3320) to reflect this. 

This also says that re-renders can be an issue. From my research, in this specific use case, it seems like a non-issue. That said, I suggest we keep this in mind during code review. cc @zeroedin.

elements/rh-scheme-dropdown/demo/localization.html
```diff
@@ -0,0 +1,26 @@
...
+  body {
+    color-scheme: light dark;
+    background-color: var(--rh-color-surface);
+    color: var(--rh-color-text-primary);
+  }
+
+  rh-scheme-dropdown {
+    /* Wider than the English defaults so Spanish option text fits. */
+    --rh-scheme-dropdown-select-inline-size: 120px;
+    --rh-scheme-dropdown-picker-inline-size: 160px;
```
zeroedin, line 23: If we are truely adding space for an icon maybe:

```Suggestion
    --rh-scheme-dropdown-select-inline-size: min-content;
    --rh-scheme-dropdown-picker-inline-size: calc(max-content + var(--rh-size-icon-04, 40px));
```

could maybe use space-2xl + space-md? for 40px if we think that is more appropriate

zeroedin (approved):
Labels Guarded, Templates Mended

elements/rh-scheme-dropdown/demo/localization.html
adamjohnson: Unfortunately, `calc()` doesn't support intrinsic size values. We'd need to use the `calc-size()` function, which is not yet baseline. 

So, for now, we can leave these values hardcoded. KISS until we can use `calc-size()`.

PR #3188, opened 2026-08-17
feat(footer): add `logoHref` prop, update RHDC default URL

### LINKED ISSUE #3186: [bug] `<rh-footer-universal>` Needs to allow different urls passed in to support multiple languages
### Describe the bug

Currently the logo link is hard coded:
https://github.com/RedHat-UX/red-hat-design-system/blob/0eb4dd9a07148484aaa0bdcfe047bdba9e41ead2/elements/rh-footer/rh-footer-universal.ts#L142

This needs to:
1. Change to `https://www.redhat.com/en` as the default.
2. Allow sites to override the url without wiping out the entire slot so different languages can be added.

### Screenshots

_No response_

### RHDS version number

4.2.1

### Steps to reproduce

1. look at https://github.com/RedHat-UX/red-hat-design-system/blob/0eb4dd9a07148484aaa0bdcfe047bdba9e41ead2/elements/rh-footer/rh-footer-universal.ts#L142

### Expected behaviour

See description.

### Element HTML

```HTML

```

### Operating System (OS)

macOS

### Browser

_No response_

_No response_

adamjohnson:
## What I did

1. Added a `logo-href` attribute on `<rh-footer-universal>` so authors can point the default Red Hat logo at a locale homepage. 
    * Defaults to `https://www.redhat.com/en`.
3. Updated the default `<rh-footer>` logo link from `https://redhat.com` to `https://www.redhat.com/en`.
4. Simplified the universal footer logo `aria-label` from "Visit Red Hat" to "Red Hat".
5. Closes #3186. 

## Testing Instructions

1. Open the [universal footer demo](https://deploy-preview-3188--red-hat-design-system.netlify.app/elements/footer/demo/footer-universal/).
2. Inspect the default logo `<a>`. Confirm `href` is `https://www.redhat.com/en` and `aria-label` is `Red Hat`.
3. In the console, run:
   ```js
   document.querySelector('rh-footer-universal').setAttribute('logo-href', 'https://www.redhat.com/ja');
   ```
1. Confirm the logo `href` updates.
6. Slot a custom logo (an `<a>` in the `logo` slot). Confirm `logo-href` no longer applies.
7. Open the [footer demo](https://deploy-preview-3188--red-hat-design-system.netlify.app/elements/footer/demo/). Inspect the site logo link and confirm it is `https://www.redhat.com/en`.
8. Check the [Code page](https://deploy-preview-3188--red-hat-design-system.netlify.app/elements/footer/code/#rh-footer-universal-attributes) for `logo-href` docs.
9. Ensure the two changesets are appropriate (`minor` for `logo-href`, `patch` for the `<rh-footer>` URL).
10. Moreover, ensure adding this prop is appropriate considering users can already overwrite the slot (but then have to maintain the SVG logo too).


elements/rh-footer/rh-footer.ts
```diff
@@ -145,7 +145,7 @@ export class RhFooter extends LitElement {
                            Red Hat corporate logo. Screen readers rely on the img \`alt\` attribute
                            or link text for identification. -->
                     <slot name="logo">
-                      <a href="https://redhat.com" aria-label="Red Hat">
+                      <a href="https://www.redhat.com/en" aria-label="Red Hat">
```
zeroedin, line 148: what if `rh-footer` and `rh-footer-universal` were configurable via a property `logo-href` ? But if you place the attribute on rh-footer, rh-footer-universal inherits it using context if available, local property overrides context, and fallbacks to default value?

```html
<rh-footer logo-href="https://www.redhat.com/es">
  <rh-footer-universal> <!-- gains /es as logo-href from context -->
  ...
  </rh-footer-universal>
</rh-footer>
```

if universal is used standalone:
```html
<rh-footer-universal logo-href="https://www.redhat.com/es"> <!-- uses local property setting /es -->
  ...
</rh-footer-universal>

<!-- or if no attr set uses default -->

<rh-footer-universal > <!-- uses default setting /en -->
  ...
</rh-footer-universal>

 ```
no attr set
```html
<rh-footer> <!-- gains /en default -->
  <rh-footer-universal> <!-- gains /en as logo-href from context -->
  ...
  </rh-footer-universal>
</rh-footer>
```

adamjohnson: I asked Claude whether this passes the sniff test. It makes sense at first glance, but a few concerns stood out:

<details>
<summary>Pros and cons of using Lit Context</summary>

**Pros of Lit Context**
- One attribute on `<rh-footer>` can set the locale URL for a nested `<rh-footer-universal>` without the author repeating it.
- Correct tool for slotted (light DOM) children — `rh-footer` cannot pass `logoHref` through its template.
- Matches existing RHDS parent/child patterns (`@provide` / `@consume`, `createContextWithRoot`).
- Subscribed context updates if `logo-href` changes on the parent at runtime.
- Standalone `<rh-footer-universal>` still works: no provider → local attr or default.

**Cons of Lit Context**
- Overkill for one string and one consumer. Authors already write `<rh-footer-universal>`; they can set `logo-href` there.
- “Local overrides context” is not how `@consume()` works. `@consume` writes the provided value onto the decorated property. You need a private consumed field plus a resolver (`local || context || default`), and you cannot default `logoHref` to `/en` or you cannot tell “unset” from “explicitly `/en`”.
- Couples two different logos. `rh-footer`’s logo is the site/product wordmark (usually slotted with its own href). `rh-footer-universal`’s logo is the corporate hat. One inherited URL can send the product logo to a locale homepage by accident.
- Timing: if universal connects before `rh-footer` upgrades, you need `ContextRoot` (this repo already has `createContextWithRoot`).
- Extra API surface, docs, and tests (provide / consume / override / standalone / no provider).
- Does not solve #3186 by itself — that issue is “override the universal hat URL without replacing the slot.” The current prop already does that.

**Current implementation (no context)**
- **Pros:** Small, obvious, matches the issue. The element that renders the link owns the attr. No merge/timing logic.
- **Cons:** Nested usage requires `logo-href` on `<rh-footer-universal>`, not on `<rh-footer>`. `rh-footer`’s default wordmark stays hardcoded unless slotted.
</details>

I had Claude examine different Red Hat properties and how they use rh-footer and rh-footer-universal href's. There's a roughly 50/50 split: about half keep the two logo hrefs in sync, and about half set them independently. Inheriting from `<rh-footer>` would get the independent cases wrong unless authors remember to override (if they eventually did opt into the `logo-href` attribute).

The timing concern also gives me pause, and the local-vs-context resolver adds a fair amount of code for a single string.

So I think we *could* do this, but it may be too much magic for something sites already configure as two separate links. Right now `logo-href` does exactly what it says on the tin. So, maybe KISS is the better solution here?

zeroedin: I forgot about having two logos.  Maybe that causes some tension.  I will say that claude response is overstated.  The timing issue if it was a concern we'd have severe problems in other areas that use context and we don't.

That said I agree about having the tension of two different logos / hrefs, but also when you slot a new logo you would essentially lose out on the automation anyways as you would overwrite the full contents of the slot including the set url via the logo-href attribute.  The value here is really only when the logo isn't overridden via the slot

```
<slot name="logo">
  <a href="${this.logoHref}" aria-label="${this.logoLabel}">
    ...
  </a>
</slot>
```

if you override that at the slot you lose that interpolation along with it, it would never receive either the context value or the set 'logo-href' value.  Maybe the 2 logo problem isn't the problem as outlined by claude.

I wonder if the universal footer should have a `<slot>` around the logo in the first place.  I wonder if there is ever an instance where this logo would be replaced.

This is a non blocker just food for thought.

adamjohnson: > I wonder if there is ever an instance where this logo would be replaced.

In my research, everyone uses the default slotted logo for the universal footer. Thus...

> I wonder if the universal footer should have a <slot> around the logo in the first place. 

You're right, probably not necessary.

zeroedin (approved):
Link Gains Tunability, Marvelous

PR #3237, opened 2026-09-09
fix(navigation-vertical): corrected aria-labelledby target missing id

### LINKED ISSUE #2915: [bug] `<rh-navigation-vertical>` @query('#title') references non-existent id, breaking aria-labelledby
## Describe the bug

In `rh-navigation-vertical.ts`, `@query('#title')` attempts to find an element with `id="title"`, but the rendered `<h2>` only has `class="visually-hidden"` — it has no `id` attribute. As a result, `this._title` is always `null` in `firstUpdated()`, and `this.#internals.ariaLabelledByElements` is never set.

This means the `<nav>` landmark has no programmatic accessible name via `aria-labelledby`, even though a visually hidden heading is rendered.

```ts
// line 34-35
@query('#title')
private _title!: HTMLHeadingElement;
```

```ts
// line 56 — missing id="title"
<h2 class="visually-hidden">${this.accessibleLabel}</h2>
```

```ts
// line 45-51 — never executes because this._title is null
protected firstUpdated(): void {
  if (!isServer) {
    if (this._title) {
      this.#internals.ariaLabelledByElements = [this._title];
    }
  }
}
```

## Debugging Info

**RHDS version number:** current `main`

## Steps to reproduce

1. Render any page with `<rh-navigation-vertical>`
2. Inspect the element internals or use a screen reader
3. The `<nav>` landmark does not have an accessible name linked via `aria-labelledby`

## Expected behaviour

The visually hidden `<h2>` should have `id="title"` so that `@query('#title')` resolves correctly and `ariaLabelledByElements` is wired up, giving the navigation landmark a programmatic accessible name.

## Suggested fix

Add `id="title"` to the heading in `render()`:

```html
<h2 id="title" class="visually-hidden">${this.accessibleLabel}</h2>
```

## Browser and System Info

All browsers — this is a DOM/ARIA wiring issue, not browser-specific.

zeroedin:
## What I did

1. Improved accessibility by adding missing title id correcting internal `aria-labeledby`

Closes #2915 

## Testing Instructions

1.

### DIFF
--- a/.changeset/whole-trees-agree.md
+++ b/.changeset/whole-trees-agree.md
@@ -2,5 +2,4 @@
 "@rhds/elements": patch
 ---
 
-`<rh-navigation-vertical>`: improved accessibility
-  
+`<rh-navigation-vertical>`: improved accessibility of the navigation label for screen readers
--- a/.changeset/whole-trees-agree.md
+++ b/.changeset/whole-trees-agree.md
@@ -1,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-navigation-vertical>`: improved accessibility
+  
--- a/elements/rh-navigation-vertical/rh-navigation-vertical.ts
+++ b/elements/rh-navigation-vertical/rh-navigation-vertical.ts
@@ -53,7 +53,7 @@ export class RhNavigationVertical extends LitElement {
 
   render(): TemplateResult<1> {
     return html`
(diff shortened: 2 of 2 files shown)


adamjohnson (approved):
Labels gain targeted markers

...if you wanted, I think the changeset could maybe be just ever-so-slightly more descriptive. YMMV.

hellogreg (approved):
Listerine gargle tastes medicinal.

Issue #3240, opened 2026-09-11
[feat]: align footer and navigation-primary logo-href api


zeroedin:
Currently navigation-primary and footer use the `<slot name="logo">` to control both the the link and the image slotted into each element:

simplified example
```
<slot name="logo">
  <a href="${this.logoHref}">
    <img> | <svg>
  </a>
</slot>
```
This tightly couples the link and the image.  However given footer is now becoming `light-dark()` native when a user wants to override the link they also have to override the image as well, resulting in having to again slot the image themselves and ensure that it can handle a light dark scheme toggle change if the page is setup to adapt to user preference.

To fix this we decouple the `<a>` and the `<img>` in the slot allowing the `<slot name="logo">` to do precisely what is intended, allow for a seperate image to be slotted and controlled by the implementation.  If they want to change the href only they just use the `logo-href` attribute.  Now when a page becomes scheme aware a change the link does not cause the implementation to have to slot a new inline svg to ensure they remain adaptable to the scheme change the default red hat logo remains which is already scheme aware built in.

Only when the user directly changes the logo will they have to ensure it handles scheme toggle changes.

suggested change
```
<a href="${this.logoHref}">
  <slot name="logo">
    <img> | <svg>
  </slot>
</a>
```

PR #3242, opened 2026-09-11
fix(navigation-primary)!: decouple logo and link in logo slot

zeroedin:
## What I did

1. Decoupled the link and the logo from the logo slot. Allows implementers to change the logo without updating the URL as a combined block when a custom image is required.
2. Aligns API with changes coming to rh-footer, and rh-footer-universal

References #3240 

## Testing Instructions

1.

## Notes to Reviewers

This is a breaking change for when the slotted content has been overridden.  Otherwise no changes should have to be made by implementers for those not overriding the logo slot.  See changes to slotted-logo demo.


elements/rh-navigation-primary/rh-navigation-primary.css
```diff
@@ -107,11 +107,25 @@
...
   border-block-end: none;
+  block-size: 100%;
+  place-content: center;
 
   @container navigation-primary (min-width: 576px) {
     border-block-end-color: transparent;
   }
 
+  & ::slotted(svg, img),
+  & svg {
```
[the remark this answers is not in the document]
adamjohnson: `::slotted(svg, img)` can turn into:

```css
::slotted(:is(svg, img)) {}
```

Be sure to test appropriately.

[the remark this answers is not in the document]
adamjohnson: Also, do people ever slot a `<picture>`?

elements/rh-navigation-primary/rh-navigation-primary.css
```diff
@@ -107,11 +107,25 @@
   display: flex;
   flex-direction: column;
   border-block-end: none;
+  block-size: 100%;
```
adamjohnson, line 110: FYI this shrinks the target click area for the logo. See explanation below:

-------------

## Logo link hit area collapsed to the graphic

The `block-size: 100%` / `fill-available` rules moved off `#logo a` onto `svg`/`img`. Default `#svg` then overrides to `24px`, so the wrapper `<a>` shrinks to the mark.

Deploy preview: bar cell is 82px tall; the logo `<a>` is 24×102 (default) and 24×32 (sub-domain). Previously the slotted/default `<a>` filled the lockup cell. 24px meets WCAG 2.5.8 on the nose; 32×24 on sub-domain is a smaller target than the rest of the bar.

elements/rh-navigation-primary/rh-navigation-primary.css L107–127, L158–160

Recommended fix: Put flex + `block-size: 100%` back on #logo a. Keep graphic sizing on `svg / ::slotted(...)`. If the 4px focus ring around the graphic is intentional, say so in the PR — the hit-area shrink is not mentioned.

elements/rh-navigation-primary/demo/slotted-logo.html
```diff
@@ -2,16 +2,14 @@
...
-      </svg>
-    </a>
+  <svg preserveAspectRatio="xMinYMid slice" viewBox="0 0 613 145" slot="logo">
```
adamjohnson, line 5: On focus, the fedora does not animate / "tip". Compare: [slotted demo on uxdot](https://ux.redhat.com/elements/navigation-primary/demo/slotted-logo/) vs. [DP here](https://deploy-preview-3242--red-hat-design-system.netlify.app/elements/navigation-primary/demo/).

elements/rh-navigation-primary/demo/slotted-logo.html
```diff
@@ -2,16 +2,14 @@
 description: Custom SVG logo slotted into the logo slot with animated fedora hat tip.
 ---
 <rh-navigation-primary role="navigation">
```
adamjohnson: Since this is showing a "custom" slotted logo, it is probably prudent to add `logo-href="/"` to the host to complete the chain of custom logo stuffs.

elements/rh-navigation-primary/rh-navigation-primary.ts
adamjohnson: The trailing `/` should not be on this URL:

```
  @property({ attribute: 'logo-href' }) logoHref = 'https://www.redhat.com/en';
```

adamjohnson: This JSDoc needs updated to reflect the proper default link.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -247,14 +247,13 @@ export class RhNavigationPrimary extends LitElement {
...
-                contain a \`<title>\` or \`alt\` text so screen readers can
-                identify the brand link.
-              -->
-              <slot name="logo">
-                <a href="${this.logoHref}">
+              <a href="${this.logoHref}">
+                <!--
+                  Accepts an inline SVG or \`<img>\` to override the default Red Hat logo.
+                  Slotted images should be swapped on scheme change or support light-dark()
+                  such as in an inline SVG.
```
adamjohnson: Consider adding something about accessible labels here (`alt` or `<title>`).

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -247,14 +247,13 @@ export class RhNavigationPrimary extends LitElement {
...
+                -->
+                <slot name="logo">
                   <svg preserveAspectRatio="xMinYMid slice" viewBox="0 0 613 145" role="img" aria-labelledby="svgtitle">
                     <title id="svgtitle">Red Hat</title>
```
adamjohnson, line 269: The footer PR adds other properties like:

- `logo-label` for the inner `<a>` accessible name
- `logo-analytics-category` and `logo-analytics-text` for analytics stuff

It looks like nav analytics live on the host (AFAICT) but consider the `logo-label` attribute and whether or not we want to include it here and/or in rh-footer.

adamjohnson (changes requested):
Mic check 1-2, 1-2.

elements/rh-navigation-primary/rh-navigation-primary.css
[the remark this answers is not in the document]
zeroedin: `<picture>` could definitely be an alternative to ``, i don't see why we couldn't shouldn't support it.

elements/rh-navigation-primary/rh-navigation-primary.css
zeroedin: This was intentional to match interactions in the design specs in Figma.

elements/rh-navigation-primary/demo/slotted-logo.html
zeroedin: Ok this is an issue that is not fixable.

Hover doesn't care about the shadowroot, if you hover over the element you can react to it in the lightdom.

Focus however is on the `<a>` which is outside the slot, and not targetable from the lightdom css.  Because the slotted image is now fully in lightdom but the focusable element around it is in the shadowdom so there is no way to trigger the animation when focus is applied from the lightdom side.

The closest we can get is:

```css
&:focus-within svg[slot="logo"] #fedora
```

However this is for when focus is anywhere inside the nav.

elements/rh-navigation-primary/rh-navigation-primary.ts
zeroedin: logo-label was added.

analytics should use composed events.

elements/rh-navigation-primary/demo/slotted-logo.html
markcaron: I'm seeing the animation in Chrome. The only difference is the focus ring size.

Image

Image

markcaron (approved):
Logo-link Gets Template Modification

elements/rh-navigation-primary/demo/slotted-logo.html
markcaron: Could be wrong, but if the logo isn't animating on focus in some cases, I'm not sure this is a blocker at this time.

markcaron: Ah, wrong demo link above. [Correct DP one](https://deploy-preview-3242--red-hat-design-system.netlify.app/elements/navigation-primary/demo/slotted-logo/)

Not sure if it's a blocker still. I believe we might need to rethink how we're composing the logo and the link, if we want to allow for custom logo animations on focus.

zeroedin: We decided that if you use a slotted logo then animations based on focus where not supported.  Removed the animation from the slotted-logo demo.

elements/rh-navigation-primary/demo/slotted-logo.html
```diff
@@ -1,17 +1,15 @@
 ---
 description: Custom SVG logo slotted into the logo slot with animated fedora hat tip.
```
adamjohnson:
```suggestion
description: Custom SVG logo in the logo slot, with logo-href and logo-label.
```

Slotted fedoras are not animated.

elements/rh-navigation-primary/demo/slotted-logo.html
```diff
@@ -362,41 +360,31 @@ <h2>Recommendations for you</h2>
 <style>
   /* animated fedora on hover */
```
adamjohnson:
```suggestion
```

Seems this comment is pretty far from the actual transform props.

elements/rh-navigation-primary/rh-navigation-primary.ts
```diff
@@ -152,12 +152,20 @@ export class RhNavigationPrimary extends LitElement {
...
+   * `<title>`, or `img` `alt`. Leave unset so the slotted mark or the
+   * default SVG title names the link. Defaults to `''`.
+   */
+  @property({ attribute: 'logo-label' }) logoLabel = '';
+
   /**
    * Sets the `href` for the default logo link. Avoid changing this value
    * unless the site requires a non-root landing page for the logo. Defaults
-   * to `'/'`.
+   * to `'https://www.redhat.com/en'`.
```
adamjohnson:
```suggestion
Sets the `href` for the logo link. Applies whether or not the `logo` slot
is overridden. Avoid changing this value except for a locale-specific
redhat.com homepage (e.g. `https://www.redhat.com/ja`). Defaults to
`'https://www.redhat.com/en'`.
```

adamjohnson (approved):
Couple small docs nits then LGTM

PR #3248, opened 2026-09-15
bug: `<rh-tile>` fix transparent background

### LINKED ISSUE #3025: [bug] `<rh-tile>` is showing transparent bg
### Describe the bug

`rh-tile` seems to be transparent now, which is unexpected.

### Screenshots

Image

### RHDS version number

4.1.3

### Steps to reproduce

Check out the [color context tile demo](https://ux.redhat.com/elements/tile/demos/#demo-color-context).

### Expected behaviour

- The default background colors should be `surface-lightest` for light scheme and `surface-darkest` for dark scheme. 
- Users should still be able to override the default surface color with the existing `color-palette` attribute.

### Element HTML

```HTML

```

### Operating System (OS)

macOS

### Browser

_No response_

_No response_

markcaron:
## What I did

1. Fixed `<rh-tile>` backgrounds, which are no longer transparent.

## Testing Instructions

1. Test Tiles color context demo.

Closes https://github.com/RedHat-UX/red-hat-design-system/issues/3025


elements/rh-tile/rh-tile.css
```diff
@@ -420,10 +429,10 @@ svg {
...
-:host([color-palette='lightest']) { background-color: var(--rh-color-surface-lightest, #ffffff); }
+:host([color-palette='lightest']) { --_background-color: var(--rh-color-surface-lightest, #ffffff); }
 
 /** Lighter palette background */
-:host([color-palette='lighter']) { background-color: var(--rh-color-surface-lighter, #f2f2f2); }
+:host([color-palette='lighter']) { --_background-color: var(--rh-color-surface-lighter, #f2f2f2); }
 
 /** Light palette background */
-:host([color-palette='light']) { background-color: var(--rh-color-surface-light, #e0e0e0); }
+:host([color-palette='light']) { --_background-color: var(--rh-color-surface-light, #e0e0e0); }
```
zeroedin: can delete the individual host selectors and use `--rh-color-surface` which gets set by the `color-palette` attribute.  

```
:host {
  background-color:
   /** Default surface, set by color-palette attribute */
    var(--rh-color-surface,
      light-dark(
        /** Default surface for light scheme */
        var(--rh-color-surface-lightest, #ffffff),
        /** Default surface for dark scheme */
        var(--rh-color-surface-darkest, #151515)
      ));
}
```

then you can remove the background on #outer that uses the private css var.

zeroedin (changes requested):

.changeset/cyan-camels-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+"@rhds/elements": patch
+---
+
+`<rh-tile>`: added default background color (`surface-lightest`/`surface-darkest`) so tiles are no longer transparent
```
zeroedin:
```Suggestion
`<rh-tile>`: ensured tiles have a visible background when no `color-palette` is specified.
```

elements/rh-tile/rh-tile.css
markcaron: I was trying to avoid styling `:host`, but this is cleaner.

marionnegp (changes requested):
I'm still seeing a transparent background in [tile's color context demo](https://deploy-preview-3248--red-hat-design-system.netlify.app/elements/tile/demos/#demo-color-context).

Image

zeroedin:
> I'm still seeing a transparent background in [tile's color context demo](https://deploy-preview-3248--red-hat-design-system.netlify.app/elements/tile/demos/#demo-color-context).

@marionnegp these tiles aren't actually transparent but they are setting the same background color by default as the parent container (`--rh-color-surface`) when `color-palette` isn't individually set on the on tile.

```css
:host {
  background-color:
   /** Default surface, set by color-palette attribute of the tile itself or parent cascade */
    var(--rh-color-surface,
      light-dark(
        /** Default surface for light scheme */
        var(--rh-color-surface-lightest, #ffffff),
        /** Default surface for dark scheme */
        var(--rh-color-surface-darkest, #151515)
      ));
}
```
If the parent cascade or host doesn't set `--rh-color-surface` then it falls back to a light-dark() value pair.

So for example if a tile is in a `rh-surface` which sets a `color-palette` attribute and there for sets `--rh-color-surface` in the cascade, like the color-context demo does for tile and other components, then the child tiles will use the same color as set in the parent unless they themselves have an explicit `color-palette` attribute to force a switch.

If you have a standard `div` parent that sets the `background-color` value say to be `purple` and no `color-scheme` in the cascade set.  Then the tile will default to be lightest `#ffffff` due to that being the default for a `light-dark()` with no set `color-scheme`.  If that div also sets `color-scheme: dark` then the tiles will be `#151515`.   

If the following css is used we can override the cascade and customize the tile backgrounds.

```css
div  { 
  color-scheme: light dark; 
  background: purple
} 
div > rh-tile { 
  --rh-color-surface: light-dark(violet, rebeccapurple); 
}
``` 

With this additional CSS the child `rh-tiles` with be reactive for `light-dark()` scheme between violet and rebeccapurple.  You can think of this as a mini theme layer.

This is in general a composition problem/solution and that likely needs to be documented better overall.  That said we likely should do a better job noting that setting a `color-palette` attribute anywhere will remove that component from being flexible for `color-scheme` switching and that effect cascades to the children. This is sort of a unsaid gotcha in my opinion right now.  

That said, equally a tile with no `color-palette` is `light-dark()` reactive but only between `#ffffff` and `#151515` by default and to override which backgrounds they use you need CSS like as shown in the example above with the purples.

zeroedin (approved):
Light-Dark Guarded; Tile Maintenance

marionnegp (approved):

